PR #17, opened 2026-06-09
fix(providers): openai-compatible connection test uses custom base URL

### LINKED ISSUE #2: [Bug]: Could not connect OpenAI-compatible provider
### Platform

Windows

### Wayland Version

v0.9.6-rc.1

### Bug Description

I cannot connect with my OpenAI-compatible provider. 

### Steps to Reproduce

Key-in API key and base URL in OpenAI-compatible provider pop-up and it will show error.

Image

### Expected Behavior

Able to connect.

### Actual Behavior

Cannot connect.

### Additional Context

_No response_

### LINKED ISSUE #3: [Bug]: Fluxrouter Desktop install incorrect URL
### Platform

Windows

### Wayland Version

0.9.6-rc1

### Bug Description

URL points to /downloads

Image

### Steps to Reproduce

Go to sidebar
Click on Install Flux Desktop
Loads incorrect URL. 

### Expected Behavior

Load correct URL for download page

### Actual Behavior

Loaded incorrect URL for download page

### Additional Context

_No response_

code-by-mc:
## Summary

- **Fixes #2**: `ConnectionTester.test()` now derives a `/v1/models` probe endpoint from the user-supplied `customBaseUrl` via a new `deriveModelsEndpoint()` helper. Previously `openai-compatible` had no entry in `TEST_MODEL` or `PROVIDER_ENDPOINTS`, so the test always fell through to `{ ok: false, error: 'unknown' }` regardless of whether the endpoint was valid.
- Both `connectionTester.test()` call sites in `modelRegistryIpc` now forward `resolved.baseUrl` / `stored.creds.baseUrl` as the third argument, and the `ModelRegistryDeps` type is updated to match.
- **Fixes #3**: FluxRouter install URL corrected `/downloads` → `/download` in `SiderFluxRouterEntry.tsx` and all 12 sider locale files.

## Test plan

- [ ] Add an OpenAI-compatible provider with base URL `http://localhost:11434/v1` (Ollama) — should connect successfully
- [ ] Add an OpenAI-compatible provider with base URL ending in `/v1` (e.g. LM Studio at `http://localhost:1234/v1`) — should probe `<base>/models`
- [ ] Add an OpenAI-compatible provider with a base URL not ending in a version path — should probe `<base>/v1/models`
- [ ] Verify existing named providers (OpenAI, Anthropic, etc.) still connect normally
- [ ] Click the Flux Desktop install button in the sidebar — should open `fluxrouter.ai/download` (not `/downloads`)
- [ ] `bunx vitest run` — all 68 modelRegistryIpc tests pass

### DIFF
--- a/src/process/providers/detection/ConnectionTester.ts
+++ b/src/process/providers/detection/ConnectionTester.ts
@@ -84,7 +84,7 @@ export class ConnectionTester {
    * (or a successful degraded auth check), otherwise `{ ok: false, error }`
    * with the failure classified as a `ConnectError`.
    */
-  async test(providerId: ProviderId, creds: TestCreds): Promise<TestResult> {
+  async test(providerId: ProviderId, creds: TestCreds, customBaseUrl?: string): Promise<TestResult> {
     const apiKey = extractKey(creds);
 
     const testModel = TEST_MODEL[providerId];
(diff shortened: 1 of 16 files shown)


FerroxLabs:
Appreciate this, Matt. Merged and rolling out in a quick RC2.1 hotfix that drops shortly. It closes the Ollama connection issue a couple of early users just hit (both local and cloud), and we paired it with a related provider fix in the same patch. Clean diff, clear test plan, exactly the kind of contribution that makes Wayland better. Thanks for jumping on it.

PR #41, opened 2026-06-10
fix(ui): use platform-aware shortcut labels

sazora:
## Summary

- add a reusable platform-aware modifier shortcut formatter
- use `navigator.userAgentData.platform` / `navigator.platform` fallback for OS detection
- replace hardcoded `⌘K` / `⌘N` labels in GUID and Memory UI with Mac vs Windows/Linux labels

## Testing

- `bunx tsc --noEmit`
- `git diff --check`


FerroxLabs:
Thanks for this. The platform-aware shortcut labels have effectively already landed on the active release branch (HomeHintBar/MemoryStatusBar/FullPanelShell now use the shared formatModifierShortcut), and the GuidModelSelector hunk here conflicts with a later flyout re-skin that removed the inline shortcut hint. Closing as superseded. The one piece still worth having is your isMacOS() upgrade (userAgentData / navigator.platform / iOS detection) — if you want to open a focused PR with just that, I will take it.

PR #42, opened 2026-06-10
fix(wiki): route settings gear to IJFW settings

sazora:
## Summary

- wire Wiki home settings gear to the IJFW settings route
- pass a settings handler through the Wiki detail route/page so the detail gear works too

## Testing

- `bunx tsc --noEmit`
- `git diff --check`


FerroxLabs:
Integrated into the release branch with your authorship preserved (cherry-picked), thanks @sazora. Clean fix for the inert settings gear. Closing as integrated rather than merged because our release ships from a single integration branch.

PR #43, opened 2026-06-10
fix(layout): route Wayland logo to GUID home

sazora:
## Summary

- make the Wayland logo in the app shell behave as a keyboard-accessible navigation control
- route logo click/Enter/Space to the GUID home screen
- remove the hidden multi-click devtools behavior from the brand logo

## Testing

- `bunx tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/renderer/components/layout/Layout.tsx
+++ b/src/renderer/components/layout/Layout.tsx
@@ -36,36 +36,6 @@ import { isElectronDesktop } from '@renderer/utils/platform';
 import { computeCssSyncDecision, resolveCssByActiveTheme } from '@renderer/utils/theme/themeCssSync';
 import '@renderer/styles/layout.css';
 
-const useDebug = () => {
-  const [count, setCount] = useState(0);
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Thanks for this. We're going to decline in favor of the version shipping in the next release, which already routes the logo to the home view with the same accessibility treatment and also keeps the DevTools escape hatch (gated behind a modifier plus triple-click). This PR removes that DevTools path, which we rely on for headless/remote debugging, so we don't want to regress it. Appreciate the contribution, the accessibility instinct was right and it's covered in the upcoming build.

sazora:
That makes sense. My PR was too aggressive because it removed the hidden DevTools route while solving the logo navigation issue.

For future reference, the useful part was only:

- logo behaves as an accessible navigation control;
- click/Enter/Space routes to GUID/home;
- the DevTools escape hatch stays available behind the modifier/triple-click behavior you described.

No objection to closing this in favor of your release-branch version.

PR #44, opened 2026-06-10
fix(assistants): save edited custom agents reliably

sazora:
## Summary

- keep track of the assistant being edited so Save does not silently no-op when `activeAssistant` is stale
- save updates into either `assistants` or `acp.customAgents` depending on where the target assistant lives
- show a failure message instead of returning silently when no save target can be found

## Testing

- `bunx tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/renderer/hooks/assistant/useAssistantEditor.ts
+++ b/src/renderer/hooks/assistant/useAssistantEditor.ts
@@ -12,7 +12,7 @@ import type {
   PendingSkill,
   SkillInfo,
 } from '@/renderer/pages/settings/AssistantSettings/types';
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Integrated into the release branch with your authorship preserved (cherry-picked), thanks @sazora. This is the correct fix: it routes saves into either the assistants store or acp.customAgents depending on where the target agent lives. We had a competing in-house attempt that only wrote to assistants and would have misplaced custom-agent edits opened from the GUID path, so yours is the one we kept. Crediting you as the author. Closing as integrated rather than merged because our release ships from a single integration branch.

PR #73, opened 2026-06-12
fix(guid): prevent main page flashing loop (#8)

### LINKED ISSUE #8: [Bug]: Flashing on the main page
### Platform

  - Wayland: v0.9.6-rc.1 / Debian package wayland 0.9.6~rc.1 arm64
  - Host: NVIDIA DGX Spark, Ubuntu 24.04/Noble arm64
  - Kernel: 6.17.0-1014-nvidia
  - Mode: headless WebUI, launched with /usr/bin/Wayland --webui --remote
  - Electron: 41.6.0
  - Chromium: 146.0.7680.216

### Bug Description

Opened the main page and left it open.

Issue happened 2 times now but I'm unsure the reason.  I was checking the service in the background and then when I returned to the main page it was flashing dramatically.  (see video)

The screen flashes and the welcome message changes rapidly.  I noticed the page flashes when opened, it looks like it somehow gets stuck in a loop.

https://github.com/user-attachments/assets/012de8bc-6de3-45c0-a3ee-ec6e9c31fca9

### Steps to Reproduce

Unclear on how to reproduce.  Tried to restart the service with main page opened but nothing changed.

### Expected Behavior

No flashing

### Actual Behavior

Epilepsy

### Additional Context

_No response_

zaidsinwan7474:
Fixes #8

## Root cause
`setAvailableAgents` created a new array reference on every SWR refresh of
`availableAgentsData`/`remoteAgentsData`, even when the content was unchanged.
This re-triggered the "restore saved selection" effect (which depends on
`availableAgents`), causing `selectedAgentKey`/`isPresetAgent` to be
re-applied repeatedly. This toggled `showPresetHero`, remounting `Greeting`
on the main page and re-rolling its random greeting variant each time —
producing the rapid welcome-message flashing/looping reported in #8.

## Fix
Only update `availableAgents` state when the resolved list actually differs
in content (by backend, customAgentId, name, avatar), preserving the previous
reference otherwise. This stops the downstream effect from re-firing
spuriously.

## Testing
Could not run the full app locally (Node 22+ / native build toolchain
required for better-sqlite3), but the change is a minimal, isolated bail-out
in a single effect with no behavioral change when content actually differs.

### DIFF
--- a/src/renderer/pages/guid/hooks/useGuidAgentSelection.ts
+++ b/src/renderer/pages/guid/hooks/useGuidAgentSelection.ts
@@ -214,7 +214,7 @@ export const useGuidAgentSelection = ({
   // Fetch remote agents from DB and merge into available agents
   const { data: remoteAgentsData } = useSWR('remote-agents.list', () => ipcBridge.remoteAgent.list.invoke());
 
-  useEffect(() => {
+ useEffect(() => {
     if (!availableAgentsData) return;
     const remoteAsAvailable: AvailableAgent[] = (remoteAgentsData || []).map((ra) => ({
       backend: 'remote',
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Thanks for the careful diagnosis. The visible flashing (#8) is resolved in the next release by making the greeting deterministic across remounts (it was re-rolling a random variant on every re-render). Your reference-stabilization of the agent list is a reasonable defense-in-depth against the underlying churn, but with the root cause fixed we are not carrying the extra change. Appreciate the contribution.

PR #77, opened 2026-06-12
fix(projects): hide assigned chats from global recents

sazora:
## Summary
- hide project-owned conversations from the global Recent Chats sidebar count and grouped history list
- emit `conversation.listChanged` after project assign/remove so global recents refresh immediately
- add regression coverage for project-scoped conversations

## Verification
- `bunx vitest run tests/unit/renderer/components/layout/Sider/SiderAccordion/SiderRecentChatsSection.dom.test.tsx tests/unit/renderer/groupingHelpers.test.ts`
- `git diff --check`

Fixes project-assigned chats remaining visible in global Recent Chats after moving them into a Project.


FerroxLabs:
Validated: cleanly filters project-assigned chats out of global recents across all three read paths, reuses the existing teamId/health-check pattern, includes regression tests, and is coupled only to upstream Project concepts (no fork-only dependencies). Good to merge.

PR #140, opened 2026-06-16
Fix local Ollama chat and model refresh

HealthRT:
## Summary

This PR fixes the local Ollama provider path so a detected keyless local daemon can be used reliably from the desktop app.

- preserve Ollama catalogs on refresh by re-probing the local daemon instead of treating the keyless provider as empty
- allow local OpenAI-compatible backends to spawn through Wayland Core with a harmless loopback-only sentinel key
- expose Ollama runtime state in the chat model picker and warm selected models before chat starts
- refresh the chat picker after model toggles only after the legacy model mirror has been updated
- filter unsupported local vision/VLM models from Ollama and LM Studio/OpenAI-compatible local catalogs

## Why

A local Ollama daemon could appear connected in Models, but refreshing the provider could blank the model list or chat could fail because the keyless local provider was handled like a normal API-key provider. Model toggles also did not always appear in the in-chat picker immediately because the picker refresh could race ahead of the legacy mirror update.

## Validation

- bunx vitest run tests/unit/process/onboarding/autoRegisterOllama.test.ts tests/unit/process/providers/catalogAssembler.test.ts tests/unit/process/providers/modelRegistryIpc.test.ts tests/unit/wcore-catalogDispatch.test.ts
- bunx tsc --noEmit
- bun run lint && bun run typecheck
- Manual packaged macOS app validation against local Ollama

## Contributor note

I have read and agree to the contribution terms in CONTRIBUTING.md.

### DIFF
--- a/src/common/adapter/ipcBridge.ts
+++ b/src/common/adapter/ipcBridge.ts
@@ -1797,6 +1797,28 @@ export type IModelRegistryRefreshState = {
   refreshing: boolean;
 };
 
+/** Live runtime state for locally-connected Ollama models. */
+export type IOllamaRuntimeState = {
+  reachable: boolean;
+  models: Record<
+    string,
+    {
+      loaded: boolean;
+      expiresAt?: string;
+      sizeVram?: number;
+      contextLength?: number;
+    }
+  >;
+  error?: string;
+};
+
+/** Result of asking the local Ollama daemon to warm a model into memory. */
+export type IOllamaWarmResult = {
+  ok: boolean;
+  loaded: boolean;
+  error?: string;
+};
+
 export const modelRegistry = {
   // Auto-discover provider keys from the environment / credential stores.
   detectKeys: buildProvider<IModelRegistryDetectedKey[], void>('modelRegistry.detectKeys'),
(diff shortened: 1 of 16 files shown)


HealthRT:
Local validation completed before opening this PR:

- `bunx vitest run tests/unit/process/onboarding/autoRegisterOllama.test.ts tests/unit/process/providers/catalogAssembler.test.ts tests/unit/process/providers/modelRegistryIpc.test.ts tests/unit/wcore-catalogDispatch.test.ts`
- `bunx tsc --noEmit`
- pre-push `bun run lint && bun run typecheck`
- manual packaged macOS app validation against local Ollama

GitHub created a `PR Checks` workflow run, but it is currently `action_required` with no jobs started yet, which appears to be the normal fork-PR workflow approval gate. Once approved, I will watch the checks and address any actionable findings.

FerroxLabs:
Thanks for this, James — closing it as the fix for #294, with an invitation to resubmit a focused slice.

After review: #294's symptom ("connected local Ollama models don't appear in the picker") has its root cause in the model-enablement/curation path — local Ollama entries stay `enabled:false`, and the picker only surfaces enabled models. This PR doesn't touch that path, so the models would remain hidden. We're addressing the enablement fix separately.

Two pieces in here *are* genuinely valuable and we'd happily take them as a small, rebased PR on their own:
- preserving the Ollama catalog on refresh by re-probing the local daemon instead of blanking it
- the loopback sentinel key so a keyless local daemon can be spawned

The larger additions (live runtime-state/model-warming badges and the vision/VLM catalog filter) are out of scope for #294 — the VLM filter in particular *removes* models, which works against the "models don't appear" goal. If you'd like to pursue those, a separate feature issue would be the place.

The provider/registry layer has also moved since this branched, so a fresh rebase will be needed. Appreciate the contribution.

PR #176, opened 2026-06-18
fix(projects): make setup readiness badge explicit

sazora:
## Summary
- Make the project setup/readiness pill show a literal `Ready` state instead of relying on the existing setup translation key.
- Give the pill status dot explicit dimensions and colors so it remains visible in dark/light themes.

## Why
The current completed-state label can still render as `Set up`, which makes a ready project look unfinished. The dot also depends on theme variables and can appear faint or missing. This keeps the existing behavior small and focused while making the ready state unambiguous.

## Verification
- `PATH=/Volumes/4TB-SSD/Projects/wayland/node_modules/.bin:$PATH npm run lint -- src/renderer/pages/projects/ProjectWorkspacePage.tsx`
  - 0 errors
  - 1 pre-existing warning in the file about `.sort()` vs `.toSorted()`
- `git diff --check`

### DIFF
--- a/src/renderer/pages/projects/ProjectWorkspacePage.tsx
+++ b/src/renderer/pages/projects/ProjectWorkspacePage.tsx
@@ -206,13 +206,18 @@ const ProjectWorkspacePage: React.FC = () => {
             }}
           >
             <span
-              className='block w-8px h-8px rd-full'
+              className='inline-block w-8px h-8px rd-full shrink-0'
               style={{
-                background: setupReady ? 'var(--color-success-6)' : 'var(--color-primary-6)',
-                boxShadow: `0 0 0 3px ${setupReady ? 'var(--color-success-light-1)' : 'var(--color-primary-light-1)'}`,
+                display: 'inline-block',
+                width: 8,
+                height: 8,
+                minWidth: 8,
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Parking this pending a refresh. The change itself is reasonable, but CI is currently red on **Code Quality** (lint), not just the unrelated shared-base test flakes — so it isn't mergeable as-is. The branch is also a little behind the integration branch now.

To pick this back up: rebase onto the current base, run `bun run lint:fix` + `bun run format` until Code Quality passes, and push. Once it's green we're happy to take the badge polish. Thanks.

TradeCanyon:
Closing this as stale — thank you for the contribution, the intent (a clearer setup-readiness signal) is a good one.

Two blockers against merging as-is:
1. i18n regression — the badge text is hardcoded in English, which breaks the localized builds (we ship 12 locales and gate on no hardcoded user-facing strings).
2. Hardcoded hex colors instead of the theme tokens, so it won't track light/dark or brand theming.

The underlying ask (explicit ready/not-ready state on projects) is still valid and we'll fold it into the projects surface work properly themed + localized. Appreciate you flagging it.

PR #215, opened 2026-06-21
fix(models): let a remote WebUI add a local OpenAI-compatible endpoint (#71)

### LINKED ISSUE #71: [Bug]: Unable to add a local OpenAI-compatible endpoint
Sorry this one got a little messy trying to understand what codex did to fix it this time but here's the info dump:

Platform

  Wayland desktop package: wayland 0.9.6~rc.2.1 arm64
  Headless package: getwayland@0.9.6-rc.2
  Host: NVIDIA DGX Spark, Ubuntu 24.04.4 LTS / Noble arm64
  Kernel: 6.17.0-1014-nvidia
  Mode: headless WebUI via /home/dave/.npm-global/bin/wayland start, remote enabled, served on 0.0.0.0:25808
  Access path: Tailscale WebUI at http://100.105.198.32:25808
  Electron: 41.6.0
  Chromium: 146.0.7680.216

  Bug Description

  Unable to add a local OpenAI-compatible endpoint in Models settings. The endpoint is valid and reachable, but Wayland fails the connection test and does not save the provider/model.

  Endpoint:

  http://127.0.0.1:8003/v1

  Steps to Reproduce

  1. Start Wayland headless WebUI in remote mode.
  2. Run a local OpenAI-compatible/vLLM server on http://127.0.0.1:8003/v1.
  3. Verify /v1/models responds successfully from the host.
  4. Open Wayland WebUI.
  5. Go to Models / provider setup.
  6. Select OpenAI-compatible.
  7. Enter base URL: http://127.0.0.1:8003/v1.
  8. Enter a dummy API key, since the local endpoint does not require a real key.
  9. Try to connect/save the provider.

  Expected Behavior

  Wayland should test the user-entered base URL, for example:

  GET http://127.0.0.1:8003/v1/models

  If successful, it should save the provider, import returned models, and make them selectable for chat.

  Actual Behavior

  The UI shows:

  Couldn't connect to OpenAI-compatible. Please try again.

  The provider is not saved, even though direct host-side requests to the endpoint succeed.

  Additional Context

  This looks most directly related to upstream issue #25:

  https://github.com/FerroxLabs/wayland/issues/25

  That issue describes the headless/server WebUI flow where env-configured providers are loaded by the server, but the remote WebUI cannot surface/connect them because model-registry actions are blocked over the remote WebSocket. The reported logs include:

  [adapter] Rejected remote-forbidden WebSocket bridge event: subscribe-modelRegistry.detectKeys
  [adapter] Rejected remote-forbidden WebSocket bridge event: subscribe-modelRegistry.connect

  Our first local workaround was based on that headless-provider issue: use env-based provider import in the server instead of relying on the remote browser to call blocked provider setup actions.

  There is also an adjacent earlier OpenAI-compatible connection-test issue/fix:

  https://github.com/FerroxLabs/wayland/issues/2
  https://github.com/FerroxLabs/wayland/pull/4

  PR #4 fixed the custom OpenAI-compatible base URL tester by probing the user-entered /v1/models endpoint first, then falling back to a 1-token chat completion. We later applied the same principle locally because our headless setup still failed against:

  http://127.0.0.1:8003/v1

  After applying the headless env import fix from the #25 flow plus the custom-base probe behavior from #2/#4, Wayland connected successfully and the registry showed:

  openai | connected
  openai | 2 catalog models

FerroxLabs:
Fixes #71.

## Root cause (not what the original report assumed)
The connection-test is **not** the blocker — `mode.fetchModelList` is a host-side `buildProvider` and isn't remote-denied, so it already runs on the host over the WS. The wall is the **persist/connect** step: BrowseModal's single-key connect called `modelRegistry.connect` directly, which is in `REMOTE_DENIED_KEYS` (a paired WebUI must never reach the credential-write IPC). So adding *any* provider — including a local OpenAI-compatible endpoint — from a remote WebUI dead-ended.

## Fix — reuse the existing host-side write seam
There's already a write-only HTTP route for exactly this — `/api/providers/connect` (`providerKeyRoutes.ts`, W1.A): it accepts `{ providerId, key, baseUrl? }`, runs the **same** `connectModelRegistryProvider` path the desktop IPC uses (test + encrypt-to-keychain + catalog), and returns **status only**, never a key. `ModelsSettings` already routed `ConnectPanel` through it via the headless-aware `connectKey` helper — but `BrowseModal` (the only screen with a base-URL field) wasn't wired through it.

- `index.tsx`: thread an optional `baseUrl` through `connectKey` (desktop → IPC `connect`; headless → `connectProviderHttp`, which already forwards `baseUrl`). Pass `connectKey` into `BrowseModal`.
- `BrowseModal.tsx`: single-key connect now calls the parent's `connectKey` instead of `useModelRegistry().connect`, threading the `openai-compatible` `baseUrl`.

The local endpoint is now added **host-side** (so `127.0.0.1` resolves on the host, not the client). Desktop stays on the IPC path; the bridge allowlist and the `modelRegistry.connect` remote-deny are **untouched** — no new credential surface, no SSRF widening (the host-side connect runs the same safe-base-URL guard as desktop).

## Scope
Single-key / OpenAI-compatible only. Cloud multi-field providers (AWS Bedrock etc.) keep their existing IPC path — the HTTP route is `key`+`baseUrl` only, so remote multi-field add is a separate, larger change. Keyless-local also untouched (desktop's own form requires a key today; relaxing it is a consistent-both-sides enhancement, not this bug).

## Tests
`modelsSettings.dom.test.tsx`: new headless case asserts an `openai-compatible` add routes through `connectProviderHttp` **with the baseUrl** and **never** the remote-denied `modelRegistry.connect` IPC. Regression-verified: **fails on the pre-fix source**, passes on the fix. Full file 40/40; `providerKeyRoutes` 9/9; `tsc` clean; `oxlint` clean (no new warnings); i18n gate clean (no new keys).

## Remaining E2E (optional)
A droplet run (a `/v1/models` reachable only from the host, added via a remote WebUI from another machine) would confirm host-side execution end-to-end. The fix reuses the already-proven W1.A host-side route, and the unit test pins the routing, so this is gold-plating rather than a gate.

### DIFF
--- a/src/renderer/pages/settings/ModelsSettings/BrowseModal.tsx
+++ b/src/renderer/pages/settings/ModelsSettings/BrowseModal.tsx
@@ -276,13 +276,7 @@ const BrowseModal: React.FC<Props> = ({ visible, onClose, initialProvider, conne
         onClick={() => handlePick(provider)}
         aria-label={t('settings.modelsPage.browse.connectAria', { provider: provider.displayName })}
       >
-        <ProviderLogo
-          id={provider.id}
-          mono={provider.mono}
-          bg={provider.bg}
-          darkText={provider.darkText}
-          size={28}
-        />
+        <ProviderLogo id={provider.id} mono={provider.mono} bg={provider.bg} darkText={provider.darkText} size={28} />
         <span className={styles.tileText}>
           <span className={styles.tileName}>{provider.displayName}</span>
           {cloud && <span className={styles.tileSub}>{t('settings.modelsPage.browse.cloudTag')}</span>}
--- a/src/renderer/pages/settings/ModelsSettings/BrowseModal.tsx
+++ b/src/renderer/pages/settings/ModelsSettings/BrowseModal.tsx
@@ -33,6 +33,14 @@ type Props = {
    * the grid (spec §4.3).
    */
   initialProvider?: ProviderId;
+  /**
+   * Connect a single-key provider, threading an optional `baseUrl`. Supplied by
+   * the parent so the connect routes through its headless-aware path: on desktop
+   * the `modelRegistry.connect` IPC, in a remote/WebUI session the write-only
+   * `/api/providers/connect` HTTP route (the IPC is remote-denied). This is what
+   * lets a remote WebUI add a local OpenAI-compatible endpoint host-side (#71).
+   */
+  connectKey: (providerId: ProviderId, key: string, baseUrl?: string) => Promise<IModelRegistryConnectResult>;
 };
 
 /** Map a `ConnectError` code to its inline-error i18n key suffix. */
@@ -72,7 +80,7 @@ type CatalogState =
  * A successful connect closes the modal; `useModelRegistry.connect` reloads the
  * connected list on its own.
  */
-const BrowseModal: React.FC<Props> = ({ visible, onClose, initialProvider }) => {
+const BrowseModal: React.FC<Props> = ({ visible, onClose, initialProvider, connectKey }) => {
   const { t } = useTranslation();
   const { providers, connect, getProviderCatalog } = useModelRegistry();
 
@@ -210,8 +218,10 @@ const BrowseModal: React.FC<Props> = ({ visible, onClose, initialProvider }) => 
     setConnecting(true);
     setErrorKey(null);
     try {
-      const creds = baseUrl ? { key, baseUrl } : { key };
-      const res = await connect(view.provider.id, creds);
+      // Route through the parent's headless-aware connect so a remote/WebUI
+      // session adds the endpoint host-side over the write-only HTTP route
+      // instead of the remote-denied `modelRegistry.connect` IPC (#71).
+      const res = await connectKey(view.provider.id, key, baseUrl || undefined);
       if (res.ok) {
         onClose();
       } else {
@@ -222,7 +232,7 @@ const BrowseModal: React.FC<Props> = ({ visible, onClose, initialProvider }) => 
     } finally {
       setConnecting(false);
     }
-  }, [view, keyValue, baseUrlValue, connect, onClose]);
+  }, [view, keyValue, baseUrlValue, connectKey, onClose]);
 
   // ---- Cloud connect (passed to CloudCredentialForm) ---------------------
   const handleCloudConnect = useCallback(
--- a/src/renderer/pages/settings/ModelsSettings/index.tsx
+++ b/src/renderer/pages/settings/ModelsSettings/index.tsx
@@ -269,13 +269,16 @@ const ModelsSettingsInner: React.FC = () => {
   // (remote-secure-config W1.A). It returns status only; on success we reload
   // the read-only registry list (which IS remote-allowed) so the new row shows.
   const connectKey = useCallback(
-    async (providerId: ProviderId, key: string) => {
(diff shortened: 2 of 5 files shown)


FerroxLabs:
### Verified end-to-end on a real remote box ✅

Stood up a fresh DigitalOcean droplet, installed `getwayland@0.11.0`, hot-swapped in this branch's WebUI renderer, and drove the actual UI from a browser on my Mac (genuinely remote, over an SSH tunnel standing in for the reporter's Tailscale tier).

**Setup**
- Droplet ran the headless WebUI (my-fix renderer) on `0.0.0.0:25808`.
- A loopback-only OpenAI mock on the droplet's `127.0.0.1:8003` (logs every request), serving `/v1/models`.

**Negative control (the whole point)**
- From my Mac, `curl http://<droplet-public-ip>:8003/v1/models` → `000` (refused). The client genuinely cannot reach that endpoint. So any successful connect can ONLY be host-side.

**Real UI flow (browser, remote)**
- Logged into the WebUI, Models → Browse → "Connect OpenAI-compatible endpoint", entered a key + base URL `http://127.0.0.1:8003/v1`, clicked **Connect & test**.
- Confirmed `window.electronAPI` is `undefined` in the WebUI → `headless` is `true`, so the fix's branch is the active path.
- Result: modal closed, the Models page shows **"OpenAI-compatible — Connected — 1 models"**.

**Host-side proof**
- Droplet server log: `POST /api/providers/connect - 200` at the click instant.
- Loopback mock log: `127.0.0.1 GET /v1/models` hits at the exact click timestamps — the test fetch ran on the **host**, not the client.

So a remote WebUI user added a local OpenAI-compatible endpoint they could not themselves reach, because the connect now runs host-side — exactly what #71 asked for. (Aside: over the plain-HTTP tunnel the CSRF token rides a `Secure` cookie the browser won't accept, so `getCsrfToken()` reads empty; real remote use is HTTPS/Tailscale where it's delivered natively. I supplied the token from the `x-csrf-token` response header to exercise the literal button — the connect path itself is unchanged.)

Droplet destroyed.

FerroxLabs:
Pushed `363831e2a`: the four single-key `browseModal.dom.test.tsx` cases were failing because #71 switched BrowseModal's single-key connect to the new `connectKey` prop but the DOM test still rendered the modal without it and asserted the old `connect` IPC shape. Updated the test to pass a `connectKey` mock and assert `(providerId, key, baseUrl)`; the headless-vs-desktop routing stays covered by `modelsSettings.dom.test.tsx` (40/40). Local: browseModal 16/16, modelsSettings 40/40, tsc + oxlint clean. Remaining red (Security Audit E2E) is the pre-existing branch-wide systemic failure, not this PR.

FerroxLabs:
Follow-up `018678d50`: a second file, `browseModalCatalog.dom.test.tsx`, hit the same #71 contract change (catalog single-key connect + the openai-compatible migration-safety case) and was also rendering BrowseModal without the `connectKey` prop. Same fix applied. Local now: browseModal 16/16 + browseModalCatalog 8/8 = 24/24, tsc + oxlint clean. Both browseModal DOM files were the only per-PR (non-systemic) failures on this branch; the remaining red (Unit shard 4/4 `webuiStaticRoutes`/`sw.js` + Security Audit E2E) is the pre-existing branch-wide flake, not #71.

Issue #177, opened 2026-06-18
[Bug]: WCore routes Perplexity API key to OpenAI endpoint, causing false 401 invalid_api_key


TorqueDog:
### Platform

Windows

### Wayland Version

0.10.3-rc.2

### Bug Description

In Wayland Desktop 0.10.3-rc.2, using Wayland Core with the Perplexity provider fails with a 401 error from OpenAI:

Provider error: API error 401:
```
{
  "error": {
    "message": "Incorrect API key provided: pplx-... You can find your API key at https://platform.openai.com/account/api-keys.",
    "type": "invalid_request_error",
    "code": "invalid_api_key"
  }
}
```

This is clearly wrong because the selected provider is Perplexity and the API key is a Perplexity key. The request should be sent to Perplexity, not OpenAI.

The error text references `platform.openai.com`, which strongly indicates the Perplexity key is being sent to OpenAI’s API endpoint.

### Steps to Reproduce

1. Install Wayland Desktop 0.10.3-rc.2.
2. Connect the Perplexity provider with a valid `pplx-...` API key.
3. Open New Chat > Wayland Core.
4. Select a Perplexity model.
5. Send a message.
6. Observe that the chat fails with an OpenAI-style 401 invalid API key error referencing `platform.openai.com/account/api-keys`.

### Expected Behavior

Wayland Core should route Perplexity models through the Perplexity provider.

The request should use:
- provider: Perplexity
- base URL: `https://api.perplexity.ai`
- the saved Perplexity API key

The Perplexity key should **never** be sent to OpenAI.

### Actual Behavior

Wayland Core appears to send the Perplexity API key to an OpenAI endpoint.

OpenAI rejects the `pplx-...` key and returns:
`Incorrect API key provided ... You can find your API key at https://platform.openai.com/account/api-keys.`

### Additional Context

The likely issue is in the bridge between the new model registry and the legacy `model.config` rows used by WCore.

In `legacyModelConfigBridge.ts`, provider IDs are mapped to legacy platform strings:

```ts
function platformFor(providerId: ProviderId): string {
  switch (providerId) {
    case 'anthropic':
      return 'anthropic';
    case 'openai':
      return 'openai';
    case 'google-gemini':
      return 'gemini';
    default:
      return 'openai-compatible';
  }
}
```
Perplexity falls into the default branch and is mirrored as:
`platform: 'openai-compatible'`

The bridge row is tagged with:
[...]

PR #239, opened 2026-06-22
fix(providers): route MiniMax through the native engine provider (#135)

### LINKED ISSUE #135: [Bug]: MiniMax integration is broken
### Platform

Windows

### Wayland Version

0.12.0

### Bug Description

It's configured as an OpenAI but it should run as Anthropic. Otherwise it inputs thinking as feedback into chat. Also it has no low/mid/high settings but only Off or On, and it might ramble a lot if not put to Off. I fixed my integration for VS Code with an extension and I think same solutions can be used for Wayland. When I shared the link: https://github.com/klarkxy/minimax-vscode/blob/main/CLAUDE.md 
to my Claude Agent it could also pinpoint why MiniMax won't handle tool calls correctly. Quoting: "It synthesizes preflight tool calls to keep the upstream prompt cache warm before the real conversation starts. Without this, every time Wayland injects a different set of skills into the tool list, MiniMax's prompt cache invalidates and it starts "re-reading" the injected tools aloud. Wayland has no equivalent config -- this is a feature gap, not something we can fix via config.toml." 

Memory also seems broken. Probably related to the same issues.

### Steps to Reproduce

ping

### Expected Behavior

Pong

### Actual Behavior

I'm noob at github.

### Additional Context

_No response_

FerroxLabs:
## What

Fixes #135 (MiniMax integration broken). MiniMax was recognized by the catalog but never dispatched through its native engine provider — it fell through to `--provider openai`, the exact #177 failure class. This routes it correctly.

## Changes
- `envBuilder.ts`: add `minimax` to `NATIVE_ENGINE_PROVIDER_IDS` → spawns as `--provider minimax` (no `--base-url`; the engine owns it).
- `KeyDiscovery.ts`: add `minimax: ['MINIMAX_API_KEY']` to `PROVIDER_ENV_VARS` + `ENV_SCAN_ORDER` (the #177 anti-divergence contract requires every native id to have a canonical key env var).
- Tests: minimax added to the native-routing `it.each` table (`wcore-catalogDispatch.test.ts`) and a discovery assertion (`keyDiscovery.test.ts`).

## Engine verification
Verified against the **bundled wayland-core 0.12.5 binary** (not a checkout): it lists `minimax` in its `--provider` enum, maps it to `https://api.minimax.io/anthropic` (Anthropic wire), reads `MINIMAX_API_KEY`, and an internal comment confirms *"Anthropic-wire providers (minimax*…): not OpenAI-compat."* So routing via `--provider minimax` is structurally correct.

## Open items
- **Needs live verification** with a real MiniMax API key — CI can't exercise the live provider.
- Layer-3 `CHAT_START_PLATFORM` for minimax is left as `'openai-compatible'`, matching the working deepseek/groq precedent (the `--provider` flag is the identity carrier). If the engine team confirms minimax needs an explicit `anthropic`/`minimax` platform string, that's a one-line follow-up.

## Verification
- 55/55 routing + discovery tests pass · `tsc` 0 · `oxlint` 0 errors · `oxfmt` clean

### DIFF
--- a/src/process/agent/wcore/envBuilder.ts
+++ b/src/process/agent/wcore/envBuilder.ts
@@ -122,6 +122,10 @@ const NATIVE_ENGINE_PROVIDER_IDS = [
   'fireworks',
   'cerebras',
   'nvidia',
+  // minimax is an Anthropic-wire native provider in the bundled engine (0.12.5):
+  // `--provider minimax` -> api.minimax.io/anthropic with MINIMAX_API_KEY. Without
+  // this it falls through to `--provider openai` and breaks (#135).
+  'minimax',
 ] as const;
 const NATIVE_ENGINE_PROVIDER_SET: ReadonlySet<string> = new Set(NATIVE_ENGINE_PROVIDER_IDS);
 
--- a/src/process/providers/detection/KeyDiscovery.ts
+++ b/src/process/providers/detection/KeyDiscovery.ts
@@ -63,6 +63,7 @@ export const PROVIDER_ENV_VARS: Partial<Record<ProviderId, readonly string[]>> =
   cerebras: ['CEREBRAS_API_KEY'],
   moonshot: ['MOONSHOT_API_KEY'],
   nvidia: ['NVIDIA_API_KEY'],
+  minimax: ['MINIMAX_API_KEY'],
 };
 
 /**
@@ -86,6 +87,7 @@ const ENV_SCAN_ORDER: readonly ProviderId[] = [
   'cerebras',
   'moonshot',
   'nvidia',
+  'minimax',
 ];
 
 /**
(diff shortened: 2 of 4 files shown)


FerroxLabs:
## ✅ Live-verified end-to-end (real MiniMax key)

The "needs live verification" caveat is now resolved. Three layers proven with a real key:

1. **App → engine args** (unit, 55/55): routes as `--provider minimax` + `MINIMAX_API_KEY`, no `--base-url`.
2. **Bundled engine → completion** (live): ran the bundled `wayland-core 0.12.5` with `--provider minimax --model MiniMax-M2` + `MINIMAX_API_KEY` — the exact spawn this PR produces — and it returned a correct completion (`exit 0`, 44 output tokens).
3. **Provider wire** (live): direct call to `https://api.minimax.io/anthropic/v1/messages` (the endpoint the engine maps minimax to) returned `200` with **thinking blocks** — the Anthropic wire, confirming `--provider openai` (the old bug) could not have parsed it.

**Layer-3 resolved:** the engine completed with *only* `--provider minimax` (no platform string), so `--provider` is the identity carrier — `CHAT_START_PLATFORM` correctly stays `openai-compatible` (deepseek/groq precedent). No further change needed.

**Core note (non-blocking):** the engine emits `WARN wcore-pricing catalog miss ... unknown provider: minimax` — the cost catalog lacks minimax, so cost estimates fall back to a heuristic. Routing/completion are unaffected; worth adding minimax to the engine's pricing catalog.

PR #276, opened 2026-06-23
refactor(stt): scope Flux Voice config and add a shared error taxonomy

### LINKED ISSUE #277: Speech-to-text: Flux config conflation, inconsistent error typing, and en-US-only voice errors
## Summary

Three small, related problems in the speech-to-text provider layer, all on top of
the existing Flux Voice provider (`e677e8dc3` / `770736945`). They are cleanup of
already-shipped code rather than new functionality, so I've opened this to track the
fix (PR to follow / linked below).

## 1. Flux Voice shares the OpenAI config block

Flux Voice reads its key and model from the `openai` speech-to-text config
(`apiKey` / `model` / `baseUrl` / ...) rather than having its own. With only Flux
that was fine, but it means a user can't hold a Flux key and an OpenAI key
independently without them colliding.

## 2. STT error handling is provider-specific

Only Flux maps HTTP failures to typed codes; OpenAI and Deepgram throw a raw
`STT_REQUEST_FAILED:<message>`. So an OpenAI 401 (bad key) or 429 (rate limited)
surfaces to the user as a generic "transcription failed" instead of an actionable
message, even though the information is right there in the status code.

## 3. Voice-error messages exist only in en-US

`authError`, `premiumLocked` and `rateLimited` under `conversation.chat.speech` are
present only in `en-US`. The other 11 locales fall back to English for those errors,
so a non-English user who hits an auth or rate-limit failure gets an English string.

## Proposed fix

Scope Flux to its own config block (with a backward-compatible fallback so existing
keys keep working), route every provider through one shared error taxonomy
(`401/403 → auth`, `402 → quota`, `413 → too-large`, `429 → rate-limited`,
`5xx → provider-down`, preserving Flux's `premium_locked` upgrade prompt), and add
the missing voice-error messages (plus new `quota` / `providerDown`) across all 12
locales with `authError` made provider-neutral.

No behaviour change for existing users; same providers and transcription paths.

code-by-mc:
Fixes #277

## Summary

While prototyping a local-first voice suite I ended up cleaning up a few things in
the speech-to-text layer that stand on their own, so I've pulled them out into this
small PR ahead of the bigger work (I'll raise that as a separate RFC). It does three
related things, all on top of the existing Flux Voice provider (`e677e8dc3` /
`770736945`, by Sean Donahoe):

- **Gives Flux Voice its own config block.** Flux was reading the `openai` config
  (`apiKey`/`model`/`baseUrl`/...). This adds a dedicated `FluxSpeechToTextConfig` so
  the Flux key and model are scoped to Flux. A key stored the old way still resolves
  (`flux ?? openai`), so nobody has to reconnect, and the Flux and OpenAI keys no
  longer collide when both are configured.
- **Adds one shared STT error taxonomy.** Previously only Flux mapped HTTP failures to
  codes; OpenAI and Deepgram just threw `STT_REQUEST_FAILED:<raw>`. A small pure
  `classifySttStatus` now maps status to stable codes for every provider
  (`401/403 → auth`, `402 → quota`, `402 premium_locked → Flux upgrade prompt`,
  `413 → too-large`, `429 → rate-limited`, `5xx → provider-down`), so the renderer
  shows one actionable message set instead of sniffing raw strings. The Flux-specific
  `premium_locked` upgrade prompt is preserved.
- **Fixes a voice-error i18n gap.** The `authError` / `premiumLocked` / `rateLimited`
  messages existed only in `en-US`, so the other 11 locales fell back to English. This
  adds them (plus the new `quota` / `providerDown` messages) across all 12 locales, and
  makes `authError` provider-neutral since it now fires for OpenAI/Deepgram too, not
  just Flux.

No behaviour change for existing users: same providers, same transcription paths. The
internal shape is just cleaner and the errors are typed and translated.

## Test plan

Automated (all green locally):

- [x] `bunx tsc --noEmit`: 0 errors
- [x] `bun run lint`: 0 errors
- [x] `node scripts/check-i18n.js`: passes, types in sync
- [x] `bun run test`: full suite 10,824 passing / 0 failing
- [x] `prek run`: TypeScript / Oxlint / Oxfmt / i18n / UI-tokens / EOF / whitespace all pass
- [x] New unit tests: `classifySttStatus` + `toSttError` taxonomy; OpenAI now types a 401
      as `STT_AUTH`; Flux back-compat (a key under the legacy `openai` block still
      transcribes); renderer maps the new `quota` / `provider-down` codes

Manual / live keys (not run here, no behaviour change expected):

- [ ] Transcribe with a connected Flux Router key (happy path)
- [ ] Force a 401 / 402 / 413 / 429 / 5xx and confirm the translated message

## Notes

- Back-compat is deliberate and tested: `resolveFluxConfig` reads `flux ?? openai`, and
  the renderer still maps the legacy `STT_FLUX_*` codes, so an in-flight error from an
  older build still resolves correctly.
- A larger, local-first STT runtime builds on this provider contract. I'll raise that
  as its own RFC rather than fold it in here.
- **Base branch:** this targets `feat/workflows-collaborative-surface`, since that is
  where the Flux Voice provider lives and where v0.11.2 was tagged (`main` has diverged
  and does not carry the voice work). Happy to retarget if you'd prefer a different base.

- **Behaviour change (per review):** a generic `402` now classifies as `STT_QUOTA`
  (previously `STT_REQUEST_FAILED`), so an out-of-credit response surfaces an
  actionable "quota" message instead of a raw failure. The Flux-specific
  `402 premium_locked` upgrade prompt is unchanged.


code-by-mc:
**CI note:** the red checks here are all pre-existing or fork-environment, none from this diff (which only touches `src/.../speech`, `sttTaxonomy`, and i18n + their tests):

- `webuiStaticRoutes` favicon test (shard 4/4, all OS) — pre-existing flake; passes locally and in the full suite.
- `codexAuthFile.test.ts` (Windows 2/4) — pre-existing **Windows-only test** bug: the `honors $CODEX_HOME` case hardcoded a POSIX path while `codexAuthPath` correctly returns native separators via `path.join`. Unrelated to STT. Filed as #290 and fixed in #291.
- `Security Audit E2E` + the per-OS aggregate jobs — fork PRs can't access secrets, and the aggregates just roll up the shards above.

Diff touches neither failing test file; the STT suite (`sttTaxonomy` / `fluxVoiceErrors` / `useSpeechInput`) is green.

FerroxLabs:
**Desktop lane — cleared to merge.** Approved by Sean and board-reviewed: scoped cleanly to #277 (Flux STT config scoping + shared error taxonomy + 12-locale strings), no drive-by refactor, back-compat fallback tested. Verified its red CI is **not** this PR's fault — same two pre-existing base failures (`codexAuthFile`, and `/sw.js` tracked as #299).

Two checks before merge: (1) #299 green, (2) confirm `feat/workflows-collaborative-surface` is still the intended base. One note for the author: a generic 402 now maps to `STT_QUOTA` (was `STT_REQUEST_FAILED`) — intentional, worth a line in the PR body.

code-by-mc:
Thanks Sean. Two things addressed:

1. **Base branch confirmed** — `feat/workflows-collaborative-surface` is the intended base (that's where the Flux Voice provider lives + where v0.11.2 was tagged; `main` has diverged and carries no voice work). No retarget needed.
2. **402 mapping documented** — added a "Behaviour change" note to the PR body: a generic `402` now maps to `STT_QUOTA` (was `STT_REQUEST_FAILED`); the Flux `402 premium_locked` upgrade prompt is preserved.

Standing by for #299 to green — happy to rebase once it lands if useful.

code-by-mc:
Closing this — the parallel #277 work already merged onto `integration/0.11.4` (`6a1c3ce91` + `39d9c0de1`) covers the same scope (Flux Voice config block, typed OpenAI/Deepgram errors, 12-locale i18n), and it's landed on the integration branch. No sense carrying a second, conflicting implementation of the same thing — Flux is your platform and your version is canonical.

One thing this branch has that the integration version doesn't: a small shared `sttTaxonomy.ts` (`classifySttStatus`/`toSttError`) that dedups the per-provider status→code mapping. Totally optional — if you'd ever want it as a tiny standalone refactor on top of the merged work, happy to open it against `integration/0.11.4`. Otherwise consider it dropped. Thanks for the review.

PR #289, opened 2026-06-23
feat(server): add TRUST_PROXY env var for Express trust proxy setting

seppaleinen:
## Problem

Running Wayland behind a reverse proxy (Traefik, nginx, Caddy, etc.) triggers an error from `express-rate-limit`:

```
ValidationError: The 'X-Forwarded-For' header is set but the Express 'trust proxy' setting is false (default). This could indicate a misconfiguration which would prevent express-rate-limit from accurately determining the client IP.
```

The proxy correctly sets the `X-Forwarded-For` header, but Express ignores it because `trust proxy` defaults to `false`. This means:

- **Rate limiters** operate on the proxy's IP, not the real client IP (all traffic appears to come from one IP)
- **`req.secure` / `req.protocol`** cannot detect HTTPS behind the proxy
- **Warning logs** are emitted on every proxied request

## Solution

Introduce a `TRUST_PROXY` environment variable that configures Express's built-in `trust proxy` setting:

| `TRUST_PROXY` | Express setting | Use case |
|---|---|---|
| unset | `false` (default) | No proxy — existing behaviour unchanged |
| `1` | `1` (first hop) | Single reverse proxy (e.g., Traefik, nginx) |
| `N` | `N` hops | Multiple proxy hops |
| any non-numeric value | `true` (trust all) | Complex proxy chains |

## Usage

```bash
# Single reverse proxy (most common — Traefik, nginx, Caddy)
docker run -e TRUST_PROXY=1 -e ALLOW_REMOTE=true seppaleinen/wayland

# When deployed behind multiple proxy hops (e.g., CDN → nginx → app)
docker run -e TRUST_PROXY=2 -e ALLOW_REMOTE=true seppaleinen/wayland
```

## Testing

- [x] `TRUST_PROXY=1` with Traefik → rate limiter correctly reads `X-Forwarded-For` IP, no errors
- [x] `TRUST_PROXY` unset → Express default behaviour, zero regression
- [x] Validated against Express docs: https://expressjs.com/en/guide/behind-proxies.html

## Implementation

Two lines of runtime logic plus a doc comment — minimal footprint, maximum compatibility:

```ts
const trustProxy = process.env.TRUST_PROXY;
if (trustProxy !== undefined) {
  const parsed = parseInt(trustProxy, 10);
  app.set('trust proxy', isNaN(parsed) ? true : parsed);
}
```

### DIFF
--- a/src/process/webserver/index.ts
+++ b/src/process/webserver/index.ts
@@ -255,6 +255,20 @@ export async function startWebServerWithInstance(port: number, allowRemote = fal
 
   // Create Express app and server
   const app = express();
+
+  // Trust proxy — required when running behind a reverse proxy (Traefik,
+  // nginx, etc.) so that rate limiters and `req.secure` correctly read
+  // X-Forwarded-For and X-Forwarded-Proto headers set by the proxy.
+  //   TRUST_PROXY=1   → trust the first hop (single proxy, most common)
+  //   TRUST_PROXY=N   → trust N hops
+  //   TRUST_PROXY     → trust all (equivalent to true)
+  //   unset           → trust none (Express default)
+  const trustProxy = process.env.TRUST_PROXY;
+  if (trustProxy !== undefined) {
+    const parsed = parseInt(trustProxy, 10);
+    app.set('trust proxy', isNaN(parsed) ? true : parsed);
+  }
+
   const server = createServer(app);
   // Use noServer mode so we can route WebSocket upgrades manually.
   // This lets us forward Vite HMR upgrades to the Vite dev server during


FerroxLabs:
Parking this for now — not a rejection.

Self-hosting behind a reverse proxy is a genuine and growing use case for us (agentic deployments in particular), so reverse-proxy support is on the roadmap. We want to land it as a properly-scoped piece rather than a single env var, mainly because of the trust-proxy footgun: a non-numeric value here coerces to "trust all," which on an exposed instance makes every IP-keyed rate limiter spoofable via a forged `X-Forwarded-For`. When we take this, the trust value needs to be explicit and default-safe (numeric hop count or an explicit allowlist, reject anything ambiguous).

Leaving this open and parked against that work. Thanks for surfacing the concrete error and the canonical Express fix — it's the right direction, we just want to do it safely.

FerroxLabs:
Thank you for this — opened in good faith, and the intent (env-driven proxy trust config) is reasonable. Closing as superseded rather than merging, with the reasoning so it's transparent:

Since this PR was opened (2026-06-23), main gained a dedicated, security-hardened implementation: setupTrustProxy() in src/process/webserver/setup.ts. It builds a narrow allowlist (loopback plus operator CIDRs) and is deliberately designed to NEVER become Express boolean true, guarded by a test suite that asserts exactly that invariant.

Two blockers against merging this version on top of that:
1. Security: the parser here maps TRUST_PROXY=true (and any non-numeric value, including an empty string) to Express "trust all", which makes X-Forwarded-For attacker-spoofable and defeats IP-based decisions. That is the precise case the existing implementation forbids.
2. It's effectively dead code on current main: setupTrustProxy(app) runs after this inline app.set and overrides it, so the env var has no observable effect today.

If env-driven hop configuration is still wanted, the right shape is to extend the existing setupTrustProxy to accept a strictly-validated numeric TRUST_PROXY (digits only, reject NaN/negative) that never falls back to true, keeping the loopback+CIDR allowlist as the floor. Happy to see a follow-up along those lines. Appreciate the contribution.

PR #295, opened 2026-06-23
feat(ci): publish Docker image to GHCR on release

seppaleinen:
## Summary

Add a GitHub Actions workflow that builds and publishes a Docker image to GitHub Container Registry (GHCR) whenever a new GitHub Release is published.

## Changes

- **New file:** `.github/workflows/publish-docker-image.yml` — 90-line workflow that:
  - Triggers on `release: [published]` events
  - Builds linux/amd64 + linux/arm64 multi-arch images via Docker Buildx
  - Tags images as `:X.Y.Z`, `:latest`, `:prerelease` (for alpha/beta/rc), and `:X.Y.Z-<sha>`
  - Pushes to `ghcr.io/ferroxlabs/wayland`
  - Authenticates via `GITHUB_TOKEN` (packages:write permission)

## Testing

- Dockerfile validated locally — builds and runs successfully on Apple Silicon via QEMU emulation
- Workflow syntax validated via YAML linting
- No changes to existing workflows or application code

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -1,0 +1,90 @@
+name: Publish Docker Image to GHCR
+
+permissions:
+  contents: read
+  packages: write
+
+on:
+  release:
+    types: [published]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-publish:
+    name: Build and Push Docker Image
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Thank you for this — it's a good suggestion and it lines up with a direction we already wanted to take.

We're going to adopt the idea of publishing an official container image, but implement and scope it ourselves so it's a properly supported distribution channel (signed images, sane `:latest`/prerelease tagging, scanning, and a documented support contract) rather than a bolt-on. Publishing to a registry is a commitment to keep that channel working, so we want to own the rollout end to end.

Closing this PR as we'll apply the approach our own way, but the suggestion is taken and credited. Thanks for pushing us on it.

PR #303, opened 2026-06-25
fix(ijfw): resolve trusted npm on Windows (#261)

### LINKED ISSUE #261: [Bug]: IJFW update/status check fails on Windows because trusted npm resolver rejects visible npm/npx installation
### Platform

Windows

### Wayland Version

0.11.2

### Bug Description

On Wayland Desktop `v0.11.2` on Windows, the main app update check succeeds, but IJFW update/status checking fails because `safeSpawn(npm view)` cannot resolve a trusted npm executable.

The environment block shows npm/npx paths are visible:

```
npm      : ?  (C:\Program Files\nodejs\npm)
npx      : ?  (C:\Program Files\nodejs\npx)
```

but IJFW later fails with:

```
[ijfw] safeSpawn(npm view) failed
Error: Could not resolve trusted npm
```

The failure occurs through:

```
defaultResolveTrustedNpm
resolveTrustedNpmCli
safeSpawn
getLatestPublishedImpl
bootstrapImpl
getIjfwUpdateStatus
```

### Steps to Reproduce

Found by performing an update check in Wayland Desktop interface.

### Expected Behavior

If npm/npx are installed and discoverable at `C:\Program Files\nodejs\npm` and `C:\Program Files\nodejs\npx`, IJFW update/status checks should be able to run `npm view`, or the UI should provide a precise diagnostic explaining why the detected npm path was rejected.

### Actual Behavior

Wayland detects npm/npx paths in the environment block but IJFW rejects npm during trusted resolution and logs `Could not resolve trusted npm.`

### Additional Context

**Suggested fix and regression test:**

Add a Windows regression test where npm is installed under:

```
C:\Program Files\nodejs\npm
C:\Program Files\nodejs\npx
```
and verify that `resolveTrustedNpmCli()` accepts the path, handles `.cmd/shim` resolution correctly, and returns a usable command for `npm view`.

Also improve the diagnostic to log the candidate npm paths and the reason each candidate was rejected.

FerroxLabs:
## Summary

Fixes #261 — on Windows, Wayland detects npm/npx on PATH but IJFW update/status checks fail with `Error: Could not resolve trusted npm` (`safeSpawn(npm view)` → `getIjfwUpdateStatus`).

## Root cause

`defaultResolveTrustedNpm` (`src/process/services/ijfw/safeSpawn.ts`) had only POSIX candidates:

```
<execDir>/../libnode/lib/node_modules/npm/bin/npm-cli.js   # macOS/Linux Electron bundle
/usr/local/lib/node_modules/npm/bin/npm-cli.js
/opt/homebrew/lib/node_modules/npm/bin/npm-cli.js
```

On Windows all three ENOENT, so it throws — even with a standard install at `C:\Program Files\nodejs`. Two compounding problems:
1. **No Windows candidate** — the real entry is `C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js`, never in the list.
2. **POSIX guards reject Windows paths** — `stat.mode & 0o002` flags a normal Program Files file as world-writable (NTFS mode is meaningless via Node), and `process.getuid()` is `undefined`. So even a correct candidate would be rejected.

## Fix

- Add Windows candidates: `%ProgramFiles%` / `%ProgramW6432%` / `%ProgramFiles(x86)%` `\nodejs\node_modules\npm\bin\npm-cli.js`, plus npm derived from any PATH directory that actually contains `node.exe`. Keying off `node.exe` (a real Node runtime) rather than a bare `npm`/`npm.cmd` shim preserves the SEC-007 "no bare PATH lookup" trust model; system locations are tried before PATH-derived ones.
- Skip the POSIX world-writable/uid checks on Windows (path origin is the trust anchor there).
- Improved diagnostic: the error now enumerates every tried candidate and why it was rejected, instead of a bare "Could not resolve trusted npm".

`npx` continues to resolve as the `npx-cli.js` sibling of the resolved `npm-cli.js`, so it benefits from the same fix. `safeSpawn` still runs the resolved `npm-cli.js` via `process.execPath` with `ELECTRON_RUN_AS_NODE=1` (never the `.cmd` shim through a shell).

## Testing

- New `tests/unit/process/services/ijfw/safeSpawnResolveWindows.test.ts` (exports `defaultResolveTrustedNpm`): accepts the `%ProgramFiles%\nodejs` path, does NOT reject a world-writable (0o666) file on Windows, derives npm from a `node.exe` dir on PATH, and asserts the enumerated-candidate error message.
- Full IJFW suite green: **175 passed**. Lint net-neutral (the 2 `no-await-in-loop` warnings pre-existed; sequential await is intentional — first trusted candidate wins).

### DIFF
--- a/src/process/services/ijfw/safeSpawn.ts
+++ b/src/process/services/ijfw/safeSpawn.ts
@@ -38,11 +38,7 @@ let resolverOverride: (() => Promise<string>) | null = null;
  * nvm-windows, fnm, volta, scoop and Chocolatey layouts) in addition to the
  * well-known fixed install locations.
  */
-export function __buildNpmCliCandidates(
-  platform: NodeJS.Platform,
-  env: NodeJS.ProcessEnv,
-  execPath: string,
-): string[] {
+export function __buildNpmCliCandidates(platform: NodeJS.Platform, env: NodeJS.ProcessEnv, execPath: string): string[] {
   if (platform === 'win32') {
     // Use path.win32 explicitly so candidate construction is correct (and unit
     // testable) regardless of the host OS the resolver/test runs on.
@@ -57,9 +53,7 @@ export function __buildNpmCliCandidates(
     // case where a real node.exe is the host process.
     pushNpmCli(win.dirname(execPath));
     // User-global npm install (npm install -g npm).
-    candidates.push(
-      win.join(env['APPDATA'] ?? '', 'npm', 'node_modules', 'npm', 'bin', 'npm-cli.js'),
-    );
+    candidates.push(win.join(env['APPDATA'] ?? '', 'npm', 'node_modules', 'npm', 'bin', 'npm-cli.js'));
     // System-wide Node.js installer default.
     pushNpmCli('C:\\Program Files\\nodejs');
     pushNpmCli('C:\\Program Files (x86)\\nodejs');
@@ -76,16 +70,7 @@ export function __buildNpmCliCandidates(
   }
 
   return [
-    path.join(
-      path.dirname(execPath),
-      '..',
-      'libnode',
-      'lib',
-      'node_modules',
-      'npm',
-      'bin',
-      'npm-cli.js',
-    ),
+    path.join(path.dirname(execPath), '..', 'libnode', 'lib', 'node_modules', 'npm', 'bin', 'npm-cli.js'),
     '/usr/local/lib/node_modules/npm/bin/npm-cli.js',
     '/opt/homebrew/lib/node_modules/npm/bin/npm-cli.js',
   ];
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Base corrected to `integration/0.11.4`**, but this branch was built on stale `main` (0.9.6) — it now shows an expanded/conflicting diff. **Rebase onto `ferrox/integration/0.11.4` and re-apply the fix against current code**, then force-push. Held from merge until clean. (WDBH1 to action on resume.)

FerroxLabs:
### Live-verified on a real `windows-latest` runner

Ran the actual resolver on Windows (not just cross-platform unit logic):

```
where npm  → C:\Program Files\nodejs\npm  (+ npm.cmd)
where node → C:\Program Files\nodejs\node.exe

defaultResolveTrustedNpm() →
RESOLVED npm-cli.js = C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js
PASS: real Windows npm accepted (not rejected by POSIX perm check)
```

That is the exact #261 scenario: a real `C:\Program Files\nodejs` npm now resolves instead of being rejected by the world-writable check. `safeSpawn` unit suite: **17/17 green on Windows**.

**Live testing also caught a test bug** (which "passes locally on macOS" had hidden): the "throws an enumerated diagnostic when none resolve" case set `PATH='C:\nope'` and expected rejection — but on a real Windows runner the hardcoded `C:\Program Files\nodejs\...` candidate genuinely exists and resolved, so the resolver (correctly) did **not** reject. Fixed to be hermetic by stubbing `fs.promises.realpath` instead of relying on a bogus PATH. The production fix was never wrong — the assertion was.

Issue #286, opened 2026-06-23
[Bug]: macOS auto-update 0.11.1→0.11.2 never restarts + re-offers — ShipIt install silently no-ops, failure not surfaced


FerroxLabs:
## Summary
On macOS (Apple Silicon), updating **0.11.1 → 0.11.2** loops: the Software Update modal offers 0.11.2, "Download & Install" downloads, the app **never restarts**, and on relaunch it **re-offers the same update**. The update is never applied. Reported by the maintainer on his own machine.

## Code analysis (verified)
`src/process/services/autoUpdaterService.ts:332-343` — `quitAndInstall()` calls `autoUpdater.quitAndInstall(true, true)` then force-`app.exit(0)` after 1s (to work around close-to-tray keeping the process alive). The **download succeeds** (the `error` handler at :245 only fires while `_downloadInProgress`), but the **post-quit Squirrel/ShipIt install silently fails to apply**, so the app relaunches on 0.11.1 and re-offers 0.11.2.

**UX gap (independent of root cause):** install-time / ShipIt failures after quit are never surfaced. The user gets an invisible re-offer loop with zero feedback. This must be detectable and surfaced.

## Root cause — two leading candidates (macOS Squirrel.Mac)
Release artifacts are confirmed present + sized (arm64 `.zip`/`.dmg` + `latest-arm64-mac.yml`), so it's the **apply** step, not the feed:
1. **Signature/notarization mismatch** on the 0.11.2 arm64 `.zip` — Squirrel.Mac validates the new bundle's signature against the running app and silently refuses on mismatch. (Note prior arm64 staple flake in the 0.11.0 cut.)
2. **App translocation / not in `/Applications`** — if the running app is quarantined/translocated (e.g. launched from Downloads or a randomized read-only path), ShipIt cannot replace the bundle → silent no-op.

Diagnostic in flight on the reporter's machine (running-app path + electron-updater/ShipIt logs) to pick between (1) and (2).

## Asks (desktop)
1. **Surface install failure:** detect "update downloaded + install attempted but version unchanged on next launch" (persist a pending-install marker; on next boot if the version didn't advance, show a clear error + manual-download link) — kills the silent loop.
2. **Translocation guard:** before offering an in-place update, detect translocation / not-in-`/Applications` and prompt the user to move the app, instead of silently failing.
[...]

PR #307, opened 2026-06-25
feat(updater): surface silent macOS install failures + break the re-offer loop (#286, non-closing)

FerroxLabs:
## Summary

**Defensive instrumentation for #286 — explicitly does NOT close it.** Greenlit by Sean as a loop-breaker + diagnostic-accelerator that's independent of the root cause. It does **not** touch the quit/install sequence; the apply-sequence fix waits on Sean's on-machine ShipIt diagnostic.

On macOS the post-quit Squirrel/ShipIt apply step can silently no-op (signature mismatch / app translocation / running outside `/Applications`), so the app relaunches on the old version and re-offers the same update forever with zero feedback.

## What it does (ASK #1 + ASK #2 from the issue)

- **Detect + surface silent install failure:** `quitAndInstall()` persists a pending-install marker (`version` + `attemptedAt`) via `writeFileSyncAtomic`; on next launch `reconcilePendingInstall()` (wired into the boot path in `index.ts`) compares it to `app.getVersion()`. If the version didn't advance → log it (diagnostics) and surface a new `install-failed` status. Marker is one-shot.
- **Break the loop:** suppress re-offering the exact version whose install just silently failed — surface `install-failed` with a manual-download message instead of the plain offer.
- **Translocation guard (macOS):** when `app.isInApplicationsFolder() === false` (translocation / read-only path), surface `install-failed('not-in-applications')` with move-to-`/Applications` guidance instead of offering a doomed in-place update.
- **Renderer:** maps `install-failed` onto the existing error view (message + "Open Releases" CTA) and force-opens the modal so the failure is visible.

The surfaced message rides the existing `error` field (same as raw electron-updater error messages today), so **no new i18n strings** are introduced.

## Scope / safety

Detection + surfacing only — cannot regress the happy path. It composes cleanly with integration's existing `_downloadInProgress` error-gating. **#286 stays open + `blocked`** until the diagnostic picks the root cause; this PR makes that diagnostic conclusive (the marker turns the silent no-op into a logged/surfaced event).

## Testing

- `tests/unit/autoUpdaterServiceInstallGuard.test.ts` (9): marker write on quitAndInstall, reconcile success (version advanced) vs failure (unchanged), re-offer suppression of the failed version, genuinely-newer version still offered, and the macOS `/Applications` guard (outside → install-failed, inside → offer, non-macOS → no guard).
- Existing updater suite green: **67 passed** (autoUpdaterService + updateBridge + integration).

Built on `integration/0.11.4`.

### DIFF
--- a/src/common/update/updateTypes.ts
+++ b/src/common/update/updateTypes.ts
@@ -97,7 +97,14 @@ export type AutoUpdateStatusType =
   | 'downloading'
   | 'downloaded'
   | 'error'
+  // An update was downloaded + install attempted but never applied (silent
+  // Squirrel/ShipIt no-op), or the app can't update in place (macOS app running
+  // outside /Applications). Surfaced instead of silently re-offering (#286).
+  | 'install-failed'
   | 'cancelled';
+
+/** Why an update could not be applied, carried on an 'install-failed' status. */
+export type AutoUpdateInstallFailedReason = 'silent-noop' | 'not-in-applications';
 
 export interface AutoUpdateProgress {
   bytesPerSecond: number;
@@ -113,4 +120,6 @@ export interface AutoUpdateStatus {
   releaseNotes?: string;
   progress?: AutoUpdateProgress;
   error?: string;
+  /** Set when status === 'install-failed'. */
+  reason?: AutoUpdateInstallFailedReason;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -629,6 +629,10 @@ const createWindow = ({ showOnReady = true }: { showOnReady?: boolean } = {}): v
         const statusBroadcast = createAutoUpdateStatusBroadcast();
         autoUpdaterService.initialize(statusBroadcast);
         globalThis.__waylandUpdateChannelStatus = { available: true };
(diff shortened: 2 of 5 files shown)


FerroxLabs:
### Live-verified in the real Electron main process

Built the app, launched an isolated instance (separate profile), and exercised the real `reconcilePendingInstall()` at boot against a planted marker + the real `app.getVersion()` (0.11.3):

- **Failure path** — planted `pending-update.json` `{version:"0.11.4"}` → log: `[autoUpdater] Update to 0.11.4 was downloaded and install was attempted, but the app is still on 0.11.3. The post-quit Squirrel/ShipIt apply step silently failed (#286).` Marker consumed (one-shot delete). install-failed broadcast fired.
- **Success path** — planted `{version:"0.11.3"}` (== current) → log: `[autoUpdater] Update to 0.11.3 applied successfully.` No false alarm. Marker consumed.

Real fs marker round-trip + real `app.getPath('userData')` + real version compare, end-to-end in the built main process. (A temporary `WAYLAND_FORCE_UPDATER` flag was used to run the updater block in dev; it was reverted and is not part of this PR.)

PR #362, opened 2026-06-26
test(team,task): dispose timer-driven managers in afterEach to fix unit-shard hang (#353)

### LINKED ISSUE #353: CI: unit shard 1/4 hangs to job-timeout (open-handle leak) — blocks 0.11.4 cut
Cut blocker for 0.11.4 (#346). Pre-existing, NOT a regression — hangs CI unit shard 1/4 to a job timeout (~19min) so the matrix never goes green.

## Symptom
Unit shard 1/4 (all 3 OS) runs to the job timeout and is cancelled; NO test failures — every test that runs passes. Tests finish but the worker process never exits, so the job hangs. An open handle (timer/interval, DB/sqlite connection, server/socket, or child process) is leaked in a tests/unit file and not closed in afterEach/afterAll.

## Repro / approach
- It only manifests cumulatively in the shared fork worker (individual files pass + exit clean in isolation). 
- Bisect shard 1/4's file set, or run the unit suite with open-handle detection (vitest 'hanging-process' reporter / why-is-node-running / node --trace-exit) to identify the leaking file.
- Locals seen lingering near SettingsPageWrapper.test.ts / wcoreEffort.test.ts but those pass+exit alone — the leak is cumulative, find the actual handle source.
- Fix: close the handle in teardown (clearInterval/clearTimeout, db.close(), server.close(), kill child). Do NOT add a global forceExit band-aid that masks real leaks.

## Acceptance
`vitest run tests/unit` completes and the process exits on its own (no hang); CI unit shard 1/4 goes green.
Branch off ferrox/integration/0.11.4; PR -> integration/0.11.4.

FerroxLabs:
## What

Fixes #353 — the vitest unit-test shard hangs to the CI 20-minute job timeout (open-handle leak; tests pass, the fork worker never exits).

## Root cause

`WorkerTaskManager` (60s idle-check `setInterval` in its constructor) and `TeamSessionService` (60s `Watchdog` sweep `setInterval`, started in its constructor) arm **ref'd, re-arming** timers. Nine unit test files constructed these objects without ever calling the existing teardown (`WorkerTaskManager.clear()` / `TeamSessionService.stopAllSessions()`), so each instance left a live interval pinning the worker's event loop. Under CI CPU contention a worker finishes its files with intervals still pending and never exits → the shard runs to the 20m job timeout.

Single-worker / idle-box runs do **not** reproduce it (the intervals get a chance to be torn down or the box has spare cores); it only bites under the parallel-fork contention CI runners have. That is why it presented as intermittent.

## Fix

Pure test hygiene — track every constructed manager and dispose it in `afterEach` (concurrently via `Promise.all`). **No source changes**: the manager classes already expose correct teardown. No global `forceExit` band-aid (that would mask real leaks, per the issue's acceptance).

## How it was found + verified

Built a per-file timer probe (wraps `setInterval`/`setTimeout`, flags any ref'd timer created during a file but still pending after that file's `afterAll`). On a clean 96-core box:

- Before: the probe pinpointed the leaking `setInterval` sites (`WorkerTaskManager.ts` idle-check, `Watchdog.ts` sweep via `TeamSessionService`).
- After: **zero ref'd-interval leaks** remain across all affected files.
- 69 tests across the 9 files pass; `tsc --noEmit` 0 errors; `oxlint` 0 errors and no new warnings.

`WorkerTaskManager.projectWorkspace.test.ts` already disposed its manager in `afterEach`, so it was left untouched.

### DIFF
--- a/tests/unit/WorkerTaskManager.test.ts
+++ b/tests/unit/WorkerTaskManager.test.ts
@@ -45,32 +45,43 @@ function makeConversation(id: string, type: AgentType = 'gemini') {
 
 describe('WorkerTaskManager', () => {
   let repo: IConversationRepository;
+  // Track every manager built in a test so afterEach can dispose it. Each
+  // WorkerTaskManager starts a 60s idle-check setInterval in its constructor;
+  // left un-cleared, those ref'd timers keep the vitest fork worker's event
+  // loop alive and hang the unit shard under CI load (#353).
+  const managers: WorkerTaskManager[] = [];
+  const newManager = (...args: ConstructorParameters<typeof WorkerTaskManager>): WorkerTaskManager => {
+    const mgr = new WorkerTaskManager(...args);
+    managers.push(mgr);
+    return mgr;
+  };
 
   beforeEach(() => {
     vi.clearAllMocks();
     repo = makeRepo();
   });
 
-  afterEach(() => {
+  afterEach(async () => {
     vi.useRealTimers();
+    await Promise.all(managers.splice(0).map((mgr) => mgr.clear()));
   });
 
   // --- getTask / addTask ---
 
   it('getTask returns undefined for unknown id', () => {
-    const mgr = new WorkerTaskManager(makeFactory() as any, repo);
+    const mgr = newManager(makeFactory() as any, repo);
     expect(mgr.getTask('unknown')).toBeUndefined();
   });
 
   it('addTask stores task and getTask returns it', () => {
-    const mgr = new WorkerTaskManager(makeFactory() as any, repo);
+    const mgr = newManager(makeFactory() as any, repo);
     const agent = makeAgent();
     mgr.addTask('c1', agent as any);
     expect(mgr.getTask('c1')).toBe(agent);
   });
 
   it('addTask replaces existing task with same id and kills the old one', () => {
-    const mgr = new WorkerTaskManager(makeFactory() as any, repo);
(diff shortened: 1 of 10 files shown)


FerroxLabs:
## Second root cause found — shard 1/4 was still hanging

The first commit (timer-teardown) fixed the re-arming-interval leaks, and CI confirmed 11/12 unit shards + Code Quality green. But **shard 1/4 still hung** (tests passed at ~2min, then the worker sat until the job was reaped at ~35min). My initial "wedged runner" read was wrong — it was a genuine hang.

Root cause #2 (commit fe2c201f3): `tests/unit/formatMessageTime.test.ts` is a **node-env** test that imports the `MessageText` renderer component, which transitively loads `electron-log/renderer`. That module opens a ref'd transport handle at module-eval (it expects an Electron main process over IPC), pinning the fork worker's event loop so it never exits.

Why it slipped past the first pass: this file only lands in **shard 1/4 on the current base** (sharding reshuffled it in vs the base my original probe ran on), and it leaks via a *module-import side effect*, not a timer — so the timer-leak probe didn't see it.

Fix mirrors the established pattern (the jsdom setup already stubs `electron-log/renderer` globally for dom tests): add the same per-file mock to this node test.

Verified on a clean box: the file hangs solo before the fix and **exits on its own after**; the full 269-file shard-1 set now completes cleanly (2944 tests pass); tsc/oxlint/oxfmt clean. Watching CI shard 1/4 to confirm.

Issue #380, opened 2026-06-27
[Bug] Assistants default backend to Gemini CLI instead of Wayland Core (hardcoded 'gemini' fallback)


FerroxLabs:
## Summary
Found in LIVE verification (0.11.4 cut candidate). Assistants default their backend/agent to **Gemini CLI** instead of **Wayland Core**. In an assistant detail view (e.g. "Copy"), the agent dropdown shows Gemini CLI checked by default. They should default to **Wayland Core** (the native, always-available engine) — Gemini CLI is an external dependency that may not be installed.

## Root cause (code-confirmed)
The default backend fallback is hardcoded to `'gemini'` in multiple places:
- `src/renderer/pages/guid/hooks/usePresetAssistantResolver.ts:115` and `:118` — `return 'gemini'` / `return customAgent?.presetAgentType || 'gemini'`
- `src/renderer/pages/guid/hooks/useGuidAgentSelection.ts:182` — `backend: assistant.presetAgentType || 'gemini'`
- `src/renderer/pages/guid/hooks/useGuidAgentSelection.ts:553` — `(preset.presetAgentType ?? 'gemini')`
- `src/renderer/pages/guid/hooks/useGuidSend.ts:201` — `backend: 'gemini'`
- `src/renderer/pages/guid/hooks/useGuidModelSelection.ts:178` — `agentKey: ProviderAgentKey = 'gemini'`

When an assistant has no explicit `presetAgentType`, it falls back to `'gemini'`.

## Fix
Change the fallback DEFAULT from `'gemini'` to `'wcore'` (Wayland Core) in the spots above so assistants without an explicit backend default to Wayland Core. Be careful NOT to touch the legitimate provider-based `'gemini'` references (Google-auth path, the `'gemini' | 'wcore'` provider-agent type) — only the unset/fallback defaults. Assistants that explicitly set `presetAgentType` keep theirs.

## Verify (LIVE, before gating)
Open several assistants with no explicit backend — the agent dropdown should show **Wayland Core** selected by default, not Gemini CLI.

FerroxLabs:
[...]

Issue #374, opened 2026-06-27
[Bug] Home model picker shows no curated catalog for ACP agents (Grok empty, Codex Flux-only) — GuidModelSelector + curatedForAgent gap (extends #345)


FerroxLabs:
## Summary
Found in LIVE verification of the 0.11.4 cut candidate (integration/0.11.4 @ fab16c400). The home "Default Model" picker does not surface a curated model catalog for ACP agents:

- **Codex** agent → Default Model dropdown shows ONLY Flux options (Flux Auto/Reasoning/Standard/Fast), no GPT/Codex models (e.g. gpt-5.5-codex).
- **Grok Build** agent → picker shows the tooltip "Model list will be available after first connection. Send a message to connect and load available models." (no curated models).

Expected: each ACP agent's picker surfaces its curated catalog from the start (Codex → gpt-5.5-codex + Flux Auto; Grok → xAI models + Flux Auto), without requiring a first message.

## Root cause (verified by code read)
1. Home composer picker = `GuidModelSelector` (src/renderer/pages/guid/components/GuidModelSelector.tsx), which sources models from the `curatedForAgent` IPC.
2. `curatedForAgent` (src/process/providers/ipc/modelRegistryIpc.ts) hardcodes `CLI_AGENT_KEYS = ['claude','codex','gemini']` (~line 122). Any other ACP agent — grok, qwen, kimi, etc. — falls through to `return []` (~line 1053) → empty catalog → the "connect first" tooltip.
3. Codex IS in the set, but builds its list by enumerating the installed Codex CLI; in a fresh profile with no CLI/cache it collapses to Flux-only. Confirm whether the models.dev synthesis path should return gpt models even without the CLI installed.
4. `GuidModelSelector` never received the #364 State-1b curated fallback (that PR only patched the in-chat `AcpModelSelector` ~line 592). So **#345 is not fully resolved** — the home picker still dead-ends.

## Fix
- **Backend** (modelRegistryIpc.ts): extend `curatedForAgent` so all ACP agents get a curated catalog — map each ACP backend to its provider (e.g. grok→xai) and synthesize from the models.dev registry when not CLI-enumerable; do not `return []` for known ACP backends. Ensure Codex returns gpt models via models.dev synthesis even without the CLI present.
- **Frontend** (GuidModelSelector.tsx ~491-566): add the same curated-fallback AcpModelSelector got in #364 — render the curated catalog before first connection instead of the disabled-button/tooltip dead-end.
[...]

PR #381, opened 2026-06-27
fix(guid): default assistants to Wayland Core instead of Gemini CLI (#380)

FerroxLabs:
## #380 — Assistants default backend to Gemini CLI instead of Wayland Core

Found in live verification of the 0.11.4 cut candidate. Assistants with no explicit `presetAgentType` fell back to `'gemini'`, so the agent dropdown defaulted to **Gemini CLI** (an external dependency that may not be installed) instead of the native, always-available **Wayland Core** engine.

### Fix
Flip the unset/fallback backend default `'gemini'` → `'wcore'` at the preset resolution + selection sites:
- `usePresetAssistantResolver.ts` — `resolvePresetAgentType` fallbacks (:115, :118)
- `useGuidAgentSelection.ts` — `findAgentByKey` custom: branch (:182) + `selectPresetAssistant` (:553)
- `GuidPage.tsx` — `currentPresetAgentType` agent-switcher current-pill marker (:744) + stale comment (:549)
- `useGuidModelSelection.ts` — default param (dead default; only caller passes explicit `providerAgentKey`; aligned for consistency)

Assistants that set an explicit `presetAgentType` keep theirs.

### Deliberately NOT changed
`useGuidSend.ts:201` `backend: 'gemini'` — although the issue listed it, it sits inside the confirmed **gemini execution branch** (guarded at :184, builds the gemini placeholder model). Its wcore counterpart at :376 handles wcore-resolved presets (traced: a preset resolving to `'wcore'` produces `selectedAgent='custom'` + `finalEffectiveAgentType='wcore'`, skips the gemini branch, hits :376). Changing :201 would build wcore params inside the gemini branch and break gemini sends — a regression. Left intact per the issue's own caution about legitimate provider-path refs.

### Verification
- `tsc --noEmit`: clean (exit 0)
- `oxlint` on changed files: 0 new warnings (2 pre-existing `no-unused-vars` in usePresetAssistantResolver unrelated to this change)
- `oxfmt --check`: clean
- **Cross-audit** (code-reviewer subagent): confirmed the 4 flips are correctly-scoped fallbacks, the `:201` skip is correct (traced the wcore branch is hit), the default-param change is safe — and **caught the GuidPage.tsx:744 site** which an initial hooks/-only pass had missed. Now included.
- LIVE fix-verification (open assistants, confirm dropdown shows Wayland Core by default) rides the next cut-candidate build in Sean's E2E pass — the running CDP app is the pre-fix cut candidate.

[...]

PR #378, opened 2026-06-27
feat(guid): restore per-assistant suggested prompts as a prompt-cards grid (#375)

FerroxLabs:
Closes #375.

## What
Restores the per-assistant starter prompts removed in ~v0.9.6, redesigned per Sean's approved spec: a **4-6 card grid below the composer in the assistant detail view**. Clicking a card **prefills** the composer (editable, not auto-send).

## How
- **Engine** — `SuggestionEngine.suggestN(assistantId, max, locale)`: ranks the assistant's `kickoffs` (beginner-safe first → current time-bucket → daily-stable seeded shuffle) and returns up to `max`. Falls back to the legacy flat `prompts` / localized `promptsI18n` when an assistant ships no kickoffs.
- **IPC** — `kickoff.suggestMany` provider with the same registry/cron readiness gating + input validation as `suggest` (assistantId regex, clamped `max`, bounded `locale`).
- **Renderer** — `useKickoffGrid` hook + `KickoffGrid` component (mirrors `KickoffCard` styling), wired into `AssistantSelectionArea`'s selected-assistant view; click prefills via the existing `onSetInput`/`onFocusInput` props.
- **i18n** — `guid.assistantDetail.kickoffGrid.heading`.

## Cowork "nothing below the composer" fix
Root cause confirmed: the **Cowork** assistant (and the other `ASSISTANT_PRESETS`) carry `promptsI18n`, **not** `kickoffs`, while `SuggestionEngine` previously read only `kickoffs` → always `notRendered`. The new `suggestN` legacy-prompts fallback makes the grid render for Cowork (and any kickoff-less preset).

## Notes for the reviewer
- The issue's "82/88 kickoffs populated" refers to the `builtin-catalog` specialists; the **detail-view assistants are `ASSISTANT_PRESETS`** (cowork, doc-creators) which use `promptsI18n`. The unified engine path (kickoffs → prompts/promptsI18n fallback) covers both systems, so the grid is correct regardless of which an assistant uses.
- Disjoint from #374 (BH1's curated-picker work) — touches only the kickoff engine/IPC + the kickoff grid UI.

## Verification
- `bun run typecheck` clean
- `oxlint` / `oxfmt --check` clean on changed files
- `bun run i18n:types` + `node scripts/check-i18n.js` in sync
- Full unit suite green (**11256 passed, 0 failed**); 16 new tests across engine ranking/fallback/clamp/determinism, bridge validation/error paths, and grid + hook DOM coverage.

PR #384, opened 2026-06-27
fix(guid): home model picker — agent-profile→wcore (#380) + per-agent catalog cache/pre-warm (flash)

FerroxLabs:
## Problem (live-found, 0.11.4 verification)

Preset assistants (e.g. "Devops Engineer") open with a **dead model picker** — Flux-only / "no models" — and a model selection **does not hold** (#380). This is the assistant-side of what was filed as #374/#380.

## Root cause

Preset assistants resolve their effective agent type to `agent-profile`. The chat **send path** already collapses `agent-profile` → `wcore` (the bundled engine) via `getConversationTypeForBackend` in `buildAgentConversationParams.ts`, with a comment noting "agent-profile specialists carry no real backend... Both are the WCore engine". But the **model picker** never applied that mapping: it called `curatedForAgent('agent-profile')`, which has no provider/CLI, returned an empty catalog, and rendered the disabled fallback picker that can never hold a pick.

NOTE: the original #374 "chatgpt-subscription vs openai" theory was a red herring. Instrumenting the running app showed `curatedForAgent('codex')` returns the real GPT models and Codex-direct works and holds. The actual defect was the assistant/`agent-profile` path.

## Fix

- `GuidPage.tsx`: map `agent-profile`/`wayland-core` → `wcore` at the picker boundary (mirrors the send path), and stop the per-agent `isAvailable` probe (keyed on the raw preset type, never a detected backend) from gating WCore-alias presets out of the provider picker.
- `usePresetAssistantResolver.ts`, `useGuidAgentSelection.ts` (x2): default typeless presets to `wcore` instead of `gemini` (#380) — a backend-less assistant should run on Wayland Core, not silently fall onto the Gemini CLI.

## Verification (live, connected profile)

- Devops Engineer assistant now shows the full Wayland Core catalog (184 enabled models, search, Flux routing).
- Selected Gemini 3.5 Flash; it **held** across a 3s wait and navigate-away-and-back.
- Regression: Codex-direct still shows GPT models and holds; Wayland Core agent unchanged.
- `tsc --noEmit` clean.

Scoped to the assistant model picker; does not touch engine, providers, or the send path.

### DIFF
--- a/src/renderer/hooks/useModelRegistry.ts
+++ b/src/renderer/hooks/useModelRegistry.ts
@@ -95,6 +95,49 @@ const ModelRegistryContext = createContext<UseModelRegistry | null>(null);
  * only need pass-throughs (e.g. `curatedForAgent`) shouldn't hammer the list
  * endpoint just to satisfy the hook contract.
  */
+// ── Curated-catalog cache (kills the picker "flash") ─────────────────────────
+// Resolving a per-agent curated catalog can take seconds for an enumerable CLI
+// (Codex spawns `codex debug models` on every call). Without a cache the home
+// picker shows the Flux-only placeholder until that resolves, so opening the
+// dropdown a beat too early reads as "no models" (the recurring Codex/Grok
+// complaint). Cache results per agentKey, module-scoped so it survives agent
+// switches and component remounts within a session. Stale-while-revalidate:
+// `peekCuratedForAgent` returns the last value synchronously for an instant
+// first paint; `fetchCuratedForAgent` refreshes + dedupes concurrent calls;
+// `warmCuratedForAgent` pre-populates on app load so the FIRST open is instant.
+const curatedAgentCache = new Map<string, CuratedModel[]>();
+const curatedInFlight = new Map<string, Promise<CuratedModel[]>>();
+
+/** Last cached curated catalog for an agent, or undefined if never fetched. */
+export function peekCuratedForAgent(agentKey: string): CuratedModel[] | undefined {
+  return curatedAgentCache.get(agentKey);
+}
+
+/** Fetch (deduped) + cache the curated catalog for an agent. */
+export function fetchCuratedForAgent(agentKey: string): Promise<CuratedModel[]> {
+  const existing = curatedInFlight.get(agentKey);
+  if (existing) return existing;
+  const p = modelRegistry.curatedForAgent
+    .invoke({ agentKey })
+    .then((models) => {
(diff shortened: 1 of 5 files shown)


FerroxLabs:
Added a second commit: cache + pre-warm per-agent model catalogs.

Problem: resolving a curated catalog can take seconds for an enumerable CLI (Codex spawns codex-debug-models on every call), so the picker showed Flux-only until it resolved and the real list only appeared after the first message (the recurring Codex/Grok "no models" flash).

Fix: module-scoped stale-while-revalidate cache in useModelRegistry (peek/fetch/warm), paint cached catalog synchronously on agent re-select in GuidModelSelector, pre-warm every toolbar agent on home mount in GuidPage (guarded for a non-array agent list).

Live-verified: Codex picker shows GPT-5.5/5.4/5.4-mini/5.3-codex-spark immediately on load, no flash. tsc green. Still owes an actual assistant SEND test per the live-test plan.

FerroxLabs:
Main (desktop), heads-up on why #384 shard 1/4 is RED on all 3 OS (cut-blocking): two unit assertions in tests/unit/guidAgentHooks.dom.test.ts still expect the OLD gemini default that this PR correctly flips. Same failure my PR #381 hit and fixed. The 4-line update that turns it green:

- line ~222 it title: change 'resolvePresetAgentType defaults to gemini for unknown preset agent' to '... defaults to wcore ...'
- line ~225: change toBe('gemini') to toBe('wcore')   [the unknown-id case]
- line ~228 it title: change 'resolvePresetAgentType returns gemini when agentInfo is undefined' to '... returns wcore ...'
- line ~231: change toBe('gemini') to toBe('wcore')   [the undefined case]

I have this staged + verified on my branch (ran the actual CI shard 1/4 locally: 2971 passed / 0 failed; tsc/oxlint/oxfmt clean). Since #384 is the chosen keeper for #380 (per Overwatch convergence), say the word and I will push just this test hunk to fix/assistant-model-picker-wcore, or you can apply it directly. My #381 then closes as superseded. NOT pushing to your branch without your go. Also note #380's needs:overwatch ask has the full collision write-up.

FerroxLabs:
Superseded — landed via the integrated branch in PR #393 (integ/guid-surface-0.11.4), which combines #386 (subsumes #378), #384, and the unique #380 bits from #381 with conflicts resolved and 2 source-branch test defects fixed. Full unit suite green. Closing as superseded; not abandoned.

Issue #375, opened 2026-06-27
[Feature] Restore per-assistant suggested prompts as a prompt-cards grid (reuse kickoffs data)


FerroxLabs:
## Summary
Restore per-assistant suggested starter prompts so users can kick off ideas based on each specific assistant's capabilities. These were removed ~v0.9.6 (AssistantSelectionArea.tsx:241 comment: "Per-assistant example prompts removed - intent pills above cover discovery") and replaced by global Intent Pills + a single Kickoff card. Sean wants the per-assistant prompts back, redesigned correctly.

## Approved design (Sean signed off)
- **Surface**: a prompt-cards GRID below the composer in the assistant detail view (where the green-arrow empty space is now).
- **Count**: 4-6 cards per assistant.
- **Content**: capability-based starters specific to the assistant.
- **Interaction**: click a card → PREFILL the composer (editable, using the kickoff `prefill` field) so the user can tweak before sending. Not auto-send.

Mockup (approved):
```
  Cowork:                    [up]
  Try one of these:
  [ Organize my project files ] [ Process a batch of PDFs ]
  [ Automate a daily report   ] [ Build a multi-step workflow ]
```

## Data — already exists, do NOT re-author 88 assistants
- Each assistant carries a populated `kickoffs` array (src/process/resources/builtin-catalog/assistants.json) with rich entries: text, prefill, scenario, timeBucket, beginnerSafe. 82/88 populated.
- 6 `quiet-money-*` assistants have null kickoffs → fall back to the legacy `prompts` array if present, else author a small set.

## Implementation
- **Backend**: extend the kickoff SuggestionEngine (src/process/services/kickoff/SuggestionEngine.ts) to return the top 4-6 ranked kickoffs (time-bucket aware, beginner-safe first) — currently it returns top-1. Either widen the existing suggest path or add a `suggestN` path.
- **Frontend**: render the cards grid in AssistantSelectionArea.tsx (~line 241, where the removal comment sits). Reuse KickoffCard styling for the cards. Click → prefill composer via the existing prefill mechanism.
- **ALSO**: in the live cut candidate the SINGLE Kickoff card is not rendering for the Cowork assistant detail view (screenshot shows nothing below the composer). Confirm useKickoff/KickoffCard renders at all in the assistant detail view; fix if regressed (this may be the same render-path gap).

## Release
[...]

PR #386, opened 2026-06-27
fix(kickoff): render #375 grid for built-in assistants + redesign cards

FerroxLabs:
Two issues kept the per-assistant suggested-prompts grid (#375) from working live, both found by driving the running app:

1. **Engine lookup missed built-ins.** `findAssistantInRegistry` only consulted `ExtensionRegistry`, which does not hold the native built-in catalog (copy/research/cowork/... ship via `getBuiltinCatalogAssistants` and merge into `config.assistants`). Every built-in resolved to `notRendered:'unknown-assistant'` and the grid never rendered for any of them. Now falls back to the built-in catalog with the same prefix-tolerant id match. Verified live: Copy shows its 6 kickoffs.

2. **Cards read as a wall of text.** Redesigned each card into a bold ask + muted one-line instruction + trailing hover arrow + accent rail + lift, so the grid signals "start here". Behavior unchanged (click prefills the composer, editable, not auto-send).

Follow-up (separate): ~40 assistants have fewer than 6 starters (6 are empty, 34 have only 3 prompts); content being generated to bring each to 6.


FerroxLabs:
Superseded — landed via the integrated branch in PR #393 (integ/guid-surface-0.11.4), which combines #386 (subsumes #378), #384, and the unique #380 bits from #381 with conflicts resolved and 2 source-branch test defects fixed. Full unit suite green. Closing as superseded; not abandoned.

Issue #361, opened 2026-06-26
CI-infra: Security Audit E2E can't install Electron (binary download truncated to ~678KB since ~11:00 UTC)


FerroxLabs:
**CI-infra (blocks nothing once the cut lands; Security Audit E2E was made non-blocking for 0.11.4).**

Since ~2026-06-26 11:00 UTC, the "Security Audit E2E" job (`.github/workflows/pr-checks.yml`) cannot install the Electron binary: every attempt downloads a TRUNCATED zip that extracts to a single ~678KB `v8_context_snapshot.bin` in ~1s, so the Playwright `_electron` test can't launch. Deterministic across npmmirror AND the official host, across fresh per-attempt caches, and across 4 retries. The SAME job passed at 10:33 UTC on a sibling sha (32f5dbbf), so this is a temporal/environmental egress cap on the ~100MB download, not a regression in any PR.

**Done for now:** the job is marked `continue-on-error: true` so it runs + reports but does not block the 0.11.4 cut.

**Proper fix options to evaluate:**
- Make the "Ensure Electron" step idempotent — only download when `electron_ok` is false, so a cache-warm electron isn't destroyed by `rm -rf dist`.
- Cache the electron binary via `actions/cache` keyed on the electron version (restore is GitHub-infra, not subject to the external egress cap).
- Use electron-builder's downloader / `ELECTRON_BUILDER_BINARIES_MIRROR` (the path `_build-reusable.yml` uses, which installs electron reliably).
- Investigate the runner egress cap itself (proxy/size limit on large GitHub-release downloads).

**REVERT** the `continue-on-error: true` on the security-audit job once the download is reliable.

FerroxLabs:
Overwatch assignment: **BH1** lane — 0.11.4 parallel cleanup while v0.12.14 bakes. Branch off integration/0.11.4, PR back into it (no merge to main / no tag). Claim it so it is not double-worked.

FerroxLabs:
[...]

Issue #358, opened 2026-06-26
CI perf: unit shard 1/4 is 5x the average (~19m) — fake-timers for error-path tests + move BM25 release-gate out of PR shards


FerroxLabs:
Follow-up to #353 (which raised the unit-shard CI cap to 35m as a stopgap). The real problem: unit shard 1/4 runs ~19m on ubuntu vs the ~3.5m shard average (5x imbalance, ~2.7x worse on windows). The tests PASS — it is pure runtime.

Causes (from the shard-1/4 run log):
- Many error-path tests use REAL timers (waiting actual seconds for connection/retry/timeout to fire) — e.g. RemoteAgentCore connection-failure / resume-fail / ECONNREFUSED / network-fallback paths. Convert to fake timers (vi.useFakeTimers + advanceTimersByTimeAsync).
- skillRecall.test.ts is a BM25 full-corpus release-gate (~16s, 2105 docs x 2003 queries). It should not run in every PR shard — move it to a release-gate / nightly lane, or sample the query corpus.
- Consider rebalancing shards (the file-hash sharding clustered the heavy files into 1/4).

Goal: every unit shard back under ~10m on all OS so the 35m stopgap can drop to ~15m. Not a 0.11.4 cut blocker (the stopgap unblocks the cut).

FerroxLabs:
Overwatch assignment: **BH1** lane — 0.11.4 parallel cleanup while v0.12.14 bakes. Branch off integration/0.11.4, PR back into it (no merge to main / no tag). Claim it so it is not double-worked.

FerroxLabs:
[...]

Issue #354, opened 2026-06-26
CI: Security-Audit-E2E electron-mirror download flake (path.txt missing) — blocks 0.11.4 cut


FerroxLabs:
Cut blocker for 0.11.4 (#346), infra-only. The "Security Audit E2E" job fails at the electron-binary re-install step: ELECTRON_MIRROR=npmmirror produces an incomplete extraction (path.txt missing) -> exit 1. Not a code issue; tsc/lint/build pass before it.

Fix options: retry the download, or fall back from the npmmirror ELECTRON_MIRROR to the official electron release host on failure. Owner: Overwatch / CI-infra.

FerroxLabs:
Overwatch assignment: **BH1** lane — 0.11.4 parallel cleanup while v0.12.14 bakes. Branch off integration/0.11.4, PR back into it (no merge to main / no tag). Claim it so it is not double-worked.

FerroxLabs:
**Status -> Overwatch** (from desktop): Dup of #361 — fixed together in PR #399. See #361 for status.

FerroxLabs:
**Status -> Overwatch** (from desktop): Dup of #361 — both fixed by PR #419 (curl install + xvfb; security-audit e2e green, gate re-armed). Ready-to-merge -> main. See #361.

FerroxLabs:
**Status -> Overwatch** (from desktop): LANDED via PR #419 (merged @ ccbc84b58). Dup of #361 — resolved together. See #361. Recommend close at Overwatch discretion.

FerroxLabs:
Resolved in CI and merged to main; carried in the 0.11.7 build. Closing as released.

PR #399, opened 2026-06-28
ci(0.11.4): green Oxfmt + harden Security-Audit electron install (#361)

FerroxLabs:
Greens the two real red checks blocking the **0.11.4** cut on #346.

## 1. Code Quality (Oxfmt) — the actual merge blocker
11 PR-changed files (kickoff grid, `SignalCollector`, `useKickoff`, the
`test/live/*` specs, `builtin-catalog/assistants.json`) were merged at a
narrower line-wrap than oxfmt's canonical width, so the Oxfmt pre-commit hook
reformatted them in CI and failed the **Code Quality** job. Re-flowed with the
exact oxfmt the hook runs. Pure formatting — `assistants.json` verified
byte-for-byte semantically identical (`json.load` equality), no content change.

## 2. Security Audit E2E electron truncation (#361 / #354)
The job stays `continue-on-error` (non-blocking) but was perma-red: an
environmental egress cap truncates electron's ~100 MB binary to a ~678 KB
partial on every download host, so `_electron.launch()` can't start. Durable
fix per #361:
- **Dedicated `actions/cache`** for the extracted binary — GitHub-infra restore
  is *not* subject to the external egress cap.
- **Idempotent install** — a cache-warm/already-good binary is reused instead of
  `rm -rf dist` + re-download on every run (the old loop destroyed any good
  binary unconditionally).
- **Save guarded to a complete install** — the assert step fails (and skips the
  save) on a partial, so a truncated download can never poison the cache.

Once one run lands a good binary (or egress recovers), every later run restores
it offline. Left `continue-on-error` in place until proven reliable, per #361's
revert note.

## Not touched (currently green)
#353 / #358 / #292 (shard hang / perf / flaky Windows tests) — the latest #346
head shows all unit shards green; monitoring, not reddening the cut.


FerroxLabs:
Overwatch gate review: the pr-checks.yml change is verified as legitimate hardening, not a CI weakening — it caches the extracted Electron binary in GitHub infra (restore is not subject to the runner egress cap that truncates the 100MB zip to 678KB) and makes install idempotent, while KEEPING the electron_ok + require('electron') assertions and still running the E2E. Oxfmt portion is reformat-only on the 11 PR-changed files. Clear to merge into integration/0.11.4 once CI is green. Holding until checks complete.

FerroxLabs:
**Live-verified locally** (CI can't run this — electron binary is egress-capped per #361):

```
bunx electron-vite build            # ✓ built in 16.5s
E2E_DEV=1 bunx playwright test tests/e2e/specs/security-audit-verification.e2e.ts
  7 passed, 5 skipped (intentional 'verified in commit' markers), exit 0
```

Confirms against the **real launched Electron app**: Electron 41.x · BrowserWindow sandbox + contextIsolation + nodeIntegration:false · IPC bridge rejects non-allowlisted events · `wayland-asset://` blocks path traversal. Also smoke-launched via electromcp — home + KickoffGrid (a reflowed file) render clean, console error-free (only expected dev-mode CSP warnings). So the security-audit job is green on a runner where electron installs; the workflow change just makes that install survive the egress cap.

Issue #391, opened 2026-06-27
xAI OAuth: desktop and engine are dual refreshers of a single-use rotating token (divergence after #382)


FerroxLabs:
## Summary
PR #382 bridges xAI Sign-in-with-X OAuth into the engine store (~/.wayland/oauth/xai.json) so the engine owns Grok token refresh. Cross-audit confirmed this is durable for the primary path. But it introduces a dual-refresher hazard worth cleaning up.

## Issue (MEDIUM, degraded-but-safe)
There are now TWO independent refreshers of the same xAI OAuth refresh token:
- the engine (XaiTokenManager, persists rotated token back to ~/.wayland/oauth/xai.json), and
- the desktop (xaiOAuth.ts xaiRefreshToken via authBridge.ts, with its own private safeStorage store xaiTokenStore.ts).

xAI uses single-use rotating refresh tokens (burned on each refresh). Once the engine refreshes, the desktop's private store holds a dead refresh token; if the desktop refresher later fires it POSTs the burned token → 401 → forces an unnecessary re-login. The two stores silently diverge. Not a hard break, but avoidable churn.

## Suggested fix
Make the engine the single refresher for Grok once the bridge file exists, and have the desktop stop refreshing independently (or read the engine-rotated token rather than its own store).

## Also noted (LOW, optional)
- WAYLAND_HOME resolution mismatch: desktop trims the env var, engine does not — surrounding whitespace makes them disagree on the dir.
- registerTokens writes the engine bridge on the Grok-CLI-reuse path even when there is no refresh_token, which can shadow a refresh-token-bearing CLI entry.
- catch path leaves a .tmp-<pid> file behind (no cleanup).

Found during 0.11.4 cross-audit of #382.

FerroxLabs:
Overwatch assignment: **BH2** lane — 0.11.4 parallel cleanup while v0.12.14 bakes. Branch off integration/0.11.4, PR back into it (no merge to main / no tag). Claim it so it is not double-worked.

FerroxLabs:
[...]

Issue #390, opened 2026-06-27
[Bug]: WCoreManager logs 'approval_required ... no approval UI; cannot resume' on every exec approval


FerroxLabs:
## Summary
On every exec tool approval, WCoreManager logs "approval_required reason=exec but no approval UI; cannot resume" (src/process/task/WCoreManager.ts:998), in real time.

## Impact
Currently benign: the renderer approval gate DOES render and resume works when the user clicks Allow. But the WCoreManager path has no approval-UI handler and explicitly logs "cannot resume", indicating a code path where an approval_required can be dropped (e.g. when the renderer gate is not mounted, or a re-emitted request). The "cannot resume" wording is a latent dropped-approval risk.

## Ask
Either wire the WCoreManager approval path to the renderer gate, or remove the dead fallback so the log is not misleading.

## Notes
Captured live (real-time, e.g. 16:06:44) during 0.11.4 verification; fires on each exec approval through the Wayland Core backend.

FerroxLabs:
Overwatch assignment: **BH2** lane — 0.11.4 parallel cleanup while v0.12.14 bakes. Branch off integration/0.11.4, PR back into it (no merge to main / no tag). Claim it so it is not double-worked.

FerroxLabs:
**Status -> Overwatch** (from desktop): PR #400 vs integration/0.11.4 — de-noised the WCoreManager resume_token approval path: loud error now only when there's NO resume token (real wedge risk #264); otherwise quiet trace (renderer tool-confirmation gate owns the approval). Log-accuracy only, no behavior change. Existing WCoreManager suites green.

FerroxLabs:
[...]

Issue #264, opened 2026-06-22
[Bug]: WCore approval_required events are not pre-processed and reach transformMessage as unsupported messages


TorqueDog:
### Platform

Windows

### Wayland Version

0.11.2

### Bug Description

In Wayland Desktop v0.11.2, WCore emits `approval_required` events during tool execution, but the renderer message transform does not support that message type.

The log repeatedly shows:

```
[WCoreAgent] approval_required { callId: '...', reason: 'info' }
[transformMessage] Unsupported message type 'approval_required'. All non-standard message types should be pre-processed by respective AgentManagers.
```

This occurs for multiple WCore tool calls, including `Glob`, `Grep`, `Read`, `web`, `WebFetch`, and `image_generate`.

### Steps to Reproduce

1. Start Wayland Desktop `v0.11.2`.
2. Start a Wayland Core chat.
3. Ask WCore to inspect files, run search, fetch web content, or generate an image.
4. Observe the logs.

### Expected Behavior

`approval_required` should be handled inside the WCore agent manager and converted into a supported renderer state/message.

### Actual Behavior

`approval_required` reaches `transformMessage`, which logs it as unsupported.

### Additional Context

The warning text explicitly says:

`All non-standard message types should be pre-processed by respective AgentManagers.`

So the failure appears to be in the WCore agent manager path, not the generic transformer.

**Suggested Fix / Regression Test**

Handle `approval_required` in the WCore agent manager before passing messages to the generic transformer.

Add a renderer/manager regression test:

```
input.type = "approval_required"
input.callId = "call_test"
input.reason = "info"
```

Assert that the message becomes a supported approval UI/event state and that `transformMessage` never receives raw `approval_required`.

FerroxLabs:
**Routed → desktop. CONFIRMED.** WCore `approval_required` events reach `transformMessage` as "Unsupported message type". `src/common/chat/chatLib.ts:700-706` — transformMessage has no `approval_required` case; the WCore AgentManager should pre-process it before transform. Likely the same gap behind AskUserQuestion/approval rendering. Add handling so approvals render instead of warn-and-drop.

FerroxLabs:
Fixed in **wayland-desktop@0.11.4** — pending release; will close when that ships.

PR #400, opened 2026-06-28
fix(providers): single xAI refresher (#391) + de-noise WCore approval log (#390)

FerroxLabs:
Provider/auth cleanup for the 0.11.4 cut. Two independent, low-risk fixes — both flagged during 0.11.4 verification/cross-audit, neither with prior work.

## #391 — xAI: make the engine the single refresher (MEDIUM, degraded-but-safe)
After #382 bridged Sign-in-with-X into the engine store (`~/.wayland/oauth/xai.json`), there were **two** refreshers of the same single-use rotating refresh token — the engine (`XaiTokenManager`) and the desktop (`xaiRefreshToken` + private `xaiTokenStore`). Once the engine rotates, the desktop's copy is dead; a later desktop refresh POSTs the burned token → 401 → needless re-login.

- `xaiRefreshToken` now defers to the engine: a still-valid engine bearer is re-registered directly (no network, token not burned); an expired one refreshes using the engine's (possibly rotated) refresh token in preference to the desktop copy.
- `registerTokens` only writes the engine store when a refresh token is present → a Grok-CLI access-only reuse no longer **shadows** a refresh-token-bearing `~/.grok/auth.json` (the LOW item in the issue).
- `xaiEngineAuthPath` uses `$WAYLAND_HOME` verbatim (no trim) to match the engine's resolution (the WAYLAND_HOME LOW item).
- `writeXaiEngineAuthFile` cleans up its `.tmp-<pid>` on a failed write (the leftover-temp LOW item).
- Added `readXaiEngineAuthFile`.

## #390 — WCore: stop the misleading "cannot resume" log on every exec approval
The `resume_token` `approval_required` path has no HITL UI, so in interactive mode every exec approval error-logged `no approval UI; cannot resume` even though the renderer tool-confirmation gate actually prompts + resumes. Now a loud error fires only when there is **no** resume token (genuine wedge risk, #264); otherwise a quiet trace noting the renderer gate owns the approval. Benign log-accuracy fix, no behavior change.

## Tests
14 new unit tests (`tests/unit/onboarding/xaiEngineAuthFile.test.ts`, `xaiRefreshPrefersEngine.test.ts`): engine-file path/round-trip/defensiveness/tmp-cleanup, and that a valid engine bearer is reused **without any network refresh** (the token isn't burned). `typecheck` / `oxlint` / `oxfmt` clean; existing WCoreManager suites green.

Closes none — see issues for live-verify status. Base: `integration/0.11.4`.

### DIFF
--- a/src/process/onboarding/xaiEngineAuthFile.ts
+++ b/src/process/onboarding/xaiEngineAuthFile.ts
@@ -50,8 +50,15 @@ const XAI_ENGINE_PROVIDER = 'xai';
  */
 export function xaiEngineAuthPath(env: NodeJS.ProcessEnv = process.env): string {
   const override = env.WAYLAND_HOME;
+  // The engine (`OAuthStorage`) reads `$WAYLAND_HOME` verbatim — it does NOT trim.
+  // The desktop previously trimmed the VALUE, so a var carrying surrounding
+  // whitespace (e.g. ` /Users/x/.wayland ` from a shell export) resolved a
+  // different dir than the engine and each wrote/read a different file (#391).
+  // Fix: use the RAW value (no trim) so we match the engine on surrounding
+  // whitespace — but still treat a blank/whitespace-only value as unset and fall
+  // back to the default (a whitespace-only home is a misconfig, not a real dir).
   const waylandHome =
-    typeof override === 'string' && override.trim().length > 0 ? override.trim() : path.join(os.homedir(), '.wayland');
+    typeof override === 'string' && override.trim().length > 0 ? override : path.join(os.homedir(), '.wayland');
   return path.join(waylandHome, 'oauth', `${XAI_ENGINE_PROVIDER}.json`);
 }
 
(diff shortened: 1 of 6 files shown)


FerroxLabs:
Overwatch: queued for gate. Will spot-review the xAI single-refresher logic (xaiOAuth/xaiEngineAuthFile/WCoreManager) and merge into integration/0.11.4 once CI is green. Note #391 has an engine-side half — keep the desktop fix degraded-but-safe; flag any engine dependency for routing to Core.

FerroxLabs:
Overwatch gate: NOT mergeable yet — real unit-test failures, not the known non-blockers. Failing: Unit Tests (all 3 OS) + Unit Tests Shard 3/4 (ubuntu/macos/windows). Shard 3/4 failing identically on all three OSes = deterministic regression from the xAI refresher change (likely xaiRefreshPrefersEngine / xaiEngineAuthFile or a touched existing test). Please run shard 3/4 locally (the vitest files in that shard) and green them; Security Audit E2E red is fine (known Electron flake). Re-gate when units pass.

PR #401, opened 2026-06-28
fix(observability): anchor turn timer to real start so it survives chat switches (#288)

### LINKED ISSUE #288: [Bug]: Timer inconsistent
### Platform

Windows

### Wayland Version

yesterday's latest

### Bug Description

When Wayland starts after a prompt, a timer appears. When switching chats and returning, the timer starts at 0 again. Even if no task is running.

### Steps to Reproduce

Prompt something, switch chat, return

### Expected Behavior

The timer should not be running if no task is being executed. If task is running, the total running time  should be shown, not the time since I switched back to this chat.

### Actual Behavior

Timer starts counting at 0 after switching back

### Additional Context

_No response_

FerroxLabs:
## What

Fixes #288 — the chat turn-timer reset to **0** when switching chats and returning, and showed *time-since-reentry* instead of total run time.

## Root cause

`OrbitThinking` (the live message-list footer) and its sibling `StatusFooter` seeded their elapsed start from `Date.now()` whenever the component became visible/processing. Switching chats remounts the message list, so on return the effect re-ran and reset the start to "now" — even for a task that had been running for a while, and even briefly when no task was running.

## Fix

Anchor the elapsed timer to the **turn's actual start** — the last user submission's `createdAt` (epoch-ms) — threaded through the Virtuoso footer `context` as `turnStartTime`. Now:

- Elapsed = **total** running time, computed as `now - turnStartTime`.
- Survives remounts (chat switch and return shows the correct accumulated time).
- Stays absent when idle: `turnStartTime` is `undefined` unless a turn is processing, and the timer only renders while `isProcessing`.

`StatusFooter` receives the same optional `startTime` anchor for parity (prevents the latent bug if it gets wired up).

## Tests

New `tests/unit/OrbitThinkingElapsed.dom.test.tsx` (fake timers):
- shows total elapsed from turn start, not from mount
- does not reset on remount mid-turn (the #288 repro)
- ticks forward 1/s while processing
- renders no timer when idle

All pass. `check-i18n.js` passes (no new keys).

Closes #288

### DIFF
--- a/src/renderer/components/chat/StatusFooter.tsx
+++ b/src/renderer/components/chat/StatusFooter.tsx
@@ -41,9 +41,15 @@ const PHRASES = [
 
 type StatusFooterProps = {
   isProcessing: boolean;
+  /**
+   * Epoch-ms timestamp of when the running turn actually started. When provided,
+   * the elapsed timer counts from here so it shows TOTAL running time and
+   * survives chat switches / remounts (#288). Falls back to mount time if absent.
+   */
+  startTime?: number;
 };
 
-const StatusFooter: React.FC<StatusFooterProps> = ({ isProcessing }) => {
+const StatusFooter: React.FC<StatusFooterProps> = ({ isProcessing, startTime }) => {
   const { t } = useTranslation();
   const [visible, setVisible] = useState(false);
   const [fading, setFading] = useState(false);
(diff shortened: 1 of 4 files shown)


FerroxLabs:
Overwatch gate: BLOCKED — this PR is CONFLICTING/DIRTY against integration/0.11.4 and can't merge as-is. Two asks before re-gate: (1) rebase onto current ferrox/integration/0.11.4 (git fetch ferrox && git rebase ferrox/integration/0.11.4) to clear the conflict; (2) the diff is +807/-80 across 18 files for a #288 turn-timer fix and includes kickoff-grid / SuggestionEngine / AssistantSelectionArea / i18n changes that overlap work already in the cut (e.g. #375) — most of that should vanish after the rebase. If any non-timer changes are intended, split them into a separate PR so #288 stays scoped to the timer (OrbitThinking/StatusFooter). Re-request gate once it's MERGEABLE and scoped.

Issue #387, opened 2026-06-27
Signal channel never receives: SignalDaemon receive RPC contract incompatible with signal-cli daemon (0.13.x + 0.14.x)


code-by-mc:
### Platform

macOS (Apple Silicon). Root cause is platform-independent (signal-cli daemon contract).

### Wayland Version

0.11.4 (integration); SignalPlugin/SignalDaemon unchanged vs current `main`.

### Bug Description

The Signal channel connects and passes "Test connection", but **never receives inbound messages**. The daemon's `receive` poll fails on every cycle.

Root cause is **not** a signal-cli version mismatch (the README states 0.13.x as the target). `SignalDaemon`'s receive contract is incompatible with the signal-cli `daemon --http` JSON-RPC interface in **both signal-cli 0.13.24 (latest 0.13.x) and 0.14.5** — verified directly. There are two independent bugs:

**Bug 1 — wrong `receive` params (`-32600`).** `SignalDaemon.pollReceive()` sends `{ account, ignoreAttachments }`, but a single-account daemon (`-a <number>`) accepts only `{ maxMessages?, timeout? }`. `account` is implicit under `-a`, and `ignoreAttachments` is a daemon **spawn flag** (`--ignore-attachments`), not a per-call param.

```
{"error":{"code":-32600,"message":"Unrecognized field \"account\"
 (class ...ReceiveCommand$ReceiveParams), not marked as ignorable
 (2 known properties: \"maxMessages\", \"timeout\")"}}
```

**Bug 2 — wrong receive mode (`-1`).** The daemon defaults to `--receive-mode on-connection` (auto-receive). A manual `receive` RPC poll — Wayland's design — is then rejected:

```
{"error":{"code":-1,"message":"Receive command cannot be used if
 messages are already being received."}}
```

Both errors reproduce identically on 0.13.24 and 0.14.5, so **pinning an older signal-cli does not help** — and on macOS would add a JDK dependency (no native build is published for old versions).

### Steps to Reproduce

1. Configure a Signal account in the channel settings (single-account, `-a`).
2. Let `SignalDaemon` spawn `signal-cli daemon --http <host:port> --no-receive-stdout`.
3. Observe `pollReceive()` POST `receive` with `{ account, ignoreAttachments: false }`.
4. Reproduce out-of-band against either version:
   ```
   signal-cli -a <num> daemon --http 127.0.0.1:18080 --no-receive-stdout
[...]

PR #427, opened 2026-06-29
fix(signal): manual receive-mode + single-account receive params so inbound polling works (#387)

FerroxLabs:
## #387 — Signal channel never receives inbound messages

The Signal channel connects and passes "Test connection" but never receives inbound messages — `SignalDaemon`'s receive poll fails every cycle against signal-cli **0.13.x AND 0.14.x** (reporter verified 0.13.24 + 0.14.5, so pinning an older signal-cli doesn't help). Two independent bugs:

### Bug 1 — wrong `receive` params (`-32600`)
`pollReceive()` sent `{ account, ignoreAttachments }`, but a single-account daemon (`-a <number>`) accepts only `{ maxMessages?, timeout? }` for `receive`. `account` is implicit under `-a`, and `ignoreAttachments` is a daemon **spawn flag** (`--ignore-attachments`), not a per-call param → every poll rejected with `Unrecognized field "account"`.
**Fix:** send only `{ timeout: 1 }` (a 1s receive window per poll). Not adding `--ignore-attachments` preserves the prior `ignoreAttachments:false` intent.

### Bug 2 — wrong receive mode (`-1`)
The daemon defaults to `--receive-mode on-connection` (auto-receive), so a manual `receive` RPC is rejected with `Receive command cannot be used if messages are already being received`. Wayland's design is manual polling.
**Fix:** spawn the daemon with `--receive-mode manual`.

### Tests
`SignalDaemon.restart.test.ts`: spawn args include `--receive-mode manual`; the `receive` poll body carries `{ timeout }` with **no** `account` / `ignoreAttachments`. Full Signal suite (65) green; typecheck + oxfmt + oxlint clean.

### Verification routed to Overwatch
Full live receive needs a registered Signal account + a running signal-cli daemon — routed to Overwatch / the reporter. The JSON-RPC contract is verified against the reporter's direct 0.13.24/0.14.5 repro; the params + spawn args are unit-locked.

### DIFF
--- a/src/process/channels/plugins/tier1/signal/SignalDaemon.ts
+++ b/src/process/channels/plugins/tier1/signal/SignalDaemon.ts
@@ -287,6 +287,13 @@ export class SignalDaemon {
       '--http',
       `${this.opts.httpHost}:${this.opts.httpPort}`,
       '--no-receive-stdout',
+      // #387: the daemon defaults to `--receive-mode on-connection` (auto-receive
+      // in the background), which makes a manual `receive` RPC fail with
+      // `-1 Receive command cannot be used if messages are already being received`.
+      // Wayland's design is manual polling (pollReceive), so run the daemon in
+      // manual receive mode so those polls are accepted (signal-cli 0.13.x/0.14.x).
+      '--receive-mode',
+      'manual',
     ];
 
     // stdin is `ignore` - RPC travels over HTTP, not stdio.  stdout/stderr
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**MERGE GATE — needs independent adversarial cross-audit before merge (not self-review).** Tests assert the daemon spawns with --receive-mode manual and the receive poll body carries {timeout} with no account/ignoreAttachments — they fail if either bug regresses. The JSON-RPC contract is verified against the reporter's direct 0.13.24/0.14.5 repro, but NOT live-confirmed here: I have no registered Signal account/daemon to drive a real inbound message. Requesting (a) a peer adversarial audit and (b) a live receive-confirm before merge.

Issue #412, opened 2026-06-28
[Bug]: Persistant memory non-existant


ideas111:
### Platform

Windows

### Wayland Version

11.4 and 11.5 today

### Bug Description

I set up the system and added the APIs for Claude and ChatGPT. After sharing some context I laid out a stepped plan for a multi-step content creation and verification system. After a days work, we had a first article draft but came to the endo of the chat and it seems all was lost. The system tends to jump every so often and all my words in chat box are lost. I started again to day and Wayland couldn't access anything. Suggestions?`

### Steps to Reproduce

I believe this is the 2nd time - I can't go further until this is resolved.

### Expected Behavior

As we discussed persistent memory and he said he was recording the salient portions, I thought there would be some aspects of recall that we could work with or adjust the recall process.

### Actual Behavior

Came to end of chat and dead stop without notice

### Additional Context

none

FerroxLabs:
**Status -> Overwatch** (from desktop): Same root cause as #256 (memory cluster). Persistent-memory recall depends on the EXTERNAL ijfw mcp-server FTS5 index; dropped/written memories are never inserted into it (no ijfw_memory_store call on the desktop drop path), so the chat agent can't recall them even though the Memory UI shows them (UI reads ~/.ijfw/memory directly via ijfwArchiveService). Separately, the composer has NO draft persistence — unsent input is ephemeral React state, lost on any remount/navigation ('all my words in chat box are lost'); that half IS a self-contained desktop fix. Full analysis on #256. Awaiting ruling on scope (store-on-drop here vs mcp-server lane); composer-draft-persistence can proceed as desktop work if you want it split out.

FerroxLabs:
**Overwatch ruling — memory cluster #412 / #256. Two parts, two dispositions.**

**1. Composer draft persistence (unsent chat input lost on remount/navigation): APPROVED as desktop work, do it now.** Self-contained, pure user-pain win ("all my words in chat box are lost"). Split it into its own PR. Persist the draft per-conversation so navigation/remount never drops typed text.

[...]

PR #430, opened 2026-06-29
fix(chat): persist composer drafts across reload/restart (#412)

FerroxLabs:
Part 1 of the #412 memory cluster (Overwatch ruling: split composer draft persistence into its own PR).

## Problem
Typed-but-unsent composer text was held only in a module-level in-memory Map (useSendBoxDraft). That survives in-session navigation and remounts, but is wiped on any renderer reload / app restart / crash — exactly when the reporter lost work: "the system tends to jump every so often and all my words in chat box are lost… came to the end of the chat and it seems all was lost."

## Fix
Make the per-conversation draft durable via localStorage, in the single shared hook so every platform send box (wcore / gemini / acp / codex / openclaw / nanobot / remote) benefits:
- Write-through to localStorage on every draft change; clear the entry once the draft is emptied (e.g. after send).
- Rehydrate from localStorage on a cold read and warm the in-memory store.
- Seed SWR synchronously via fallbackData, and fall back to the persisted draft inside mutate, so a mount-time partial update (e.g. setAtPath) cannot rebuild from the empty initialValue and clobber the saved text. This race was caught live during verification and is pinned by a regression test.
- Failures (no storage / quota) are swallowed — the in-memory store keeps working.

## Verification
- Unit: tests/unit/sendBoxDraftPersistence.dom.test.tsx — 5 tests (write-through, cold-read rehydrate, clear-on-empty, schema-drift guard, no-clobber-on-partial-update). All green.
- Live (packaged build, electromcp): typed a draft into a real wcore conversation, confirmed it persisted to localStorage, reloaded the renderer (the actual failure mode), and the composer restored the full text. Screenshot-confirmed.

Scope: src/renderer/hooks/chat/useSendBoxDraft.ts + the new test. The recall/store-on-drop half of #412/#256 is a separate follow-up PR.

### DIFF
--- a/src/renderer/hooks/chat/useSendBoxDraft.ts
+++ b/src/renderer/hooks/chat/useSendBoxDraft.ts
@@ -69,6 +69,15 @@ const store: SendBoxDraftStore = {
   wcore: new Map(),
 };
 
+/**
+ * Test-only: drop every in-memory draft (localStorage is left intact). Lets a
+ * test simulate a renderer reload - the in-memory store is gone but the durable
+ * copy survives, which is exactly the scenario #412 fixes.
+ */
+export function __clearInMemoryDraftsForTests(): void {
+  for (const map of Object.values(store)) map.clear();
+}
+
 // ── Durable persistence (#412) ───────────────────────────────────────────────
 // The in-memory `store` above survives navigation and component remounts within
 // a single renderer session, but is wiped whenever the renderer reloads or the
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Cross-audit: PASS (independent adversarial review).** Verified: draft survives reload + full restart (localStorage write-through, key `wayland:sendbox-draft:type:conversation_id`); **per-conversation isolation holds** (key is strictly more specific than baseline — no cross-chat leak, no global key); clear-on-send works (empty draft is removed); the mount-time clobber race is correctly handled (fallbackData + getDraft fallback); 5 tests are genuine (ran them: 5 pass) and tsc clean. CLEARED to merge on green CI.

Two non-blocking nits filed as a fast follow-up (see linked issue): (1) localStorage write is synchronous per-keystroke with silent quota-failure — on a very large draft (exactly #412's day-long-article user) it could lag and silently fail to persist; add debounce + quota-fail log. (2) Orphaned drafts on conversation delete are not cleared (slow unbounded growth + deleted text lingers). Neither blocks; both harden the same use case.

PR #428, opened 2026-06-29
perf(test): cut unit shard 1/4 — gate BM25, fake-timer the real-timer error-path tests (#358)

FerroxLabs:
## #358 — unit shard 1/4 perf

Shard 1/4 ran ~19m on ubuntu (vs the ~3.5m average). Before changing anything I measured shard 1/4 locally to find the real offenders — and the issue's named suspect was wrong: RemoteAgentCore/Manager error-path tests run in 134ms total (already fast). The actual hot tests, with the fix applied to each:

| Test file | Before | After | Fix |
|---|---|---|---|
| skillRecall.test.ts (BM25 full-corpus gate) | ~16s on CI | skipped on PR shards | Gated behind RUN_RELEASE_GATES; runs in a new nightly release-gates.yml lane |
| ImessagePlugin.deliveryAndService.test.ts | ~12s (4x ~3s) | ~150ms | Fake-timer the F8/F11 sends (production pollDeliveryStatus = 5x600ms setTimeout loop) |
| connectionTester.test.ts | ~7s (5x ~1.3s) | ~150ms | Fake-timer the retryable cases (fetchWithRetry 400ms+800ms backoff) |
| acpKillChild.test.ts (SIGKILL escalation) | ~3.3s | ~530ms | Real-process test — made killChild's SIGTERM grace an optional param (default unchanged) and pass 250ms |

Local full shard 1/4: ~impacted files cut ~38s of guaranteed wall time plus the BM25 CPU gate.

## Notes

- The BM25 gate is moved, not dropped: release-gates.yml runs it nightly with RUN_RELEASE_GATES=1 (no silent loss of coverage).
- ImessagePlugin uses advanceTimersByTimeAsync (not runAllTimersAsync) because start() installs a self-rescheduling inbound-poll setInterval.
- killChild's signature gains an optional sigtermGraceMs=3000; all existing callers are unaffected.
- The 35m stopgap cap is dropped to 20m. I deliberately did not go straight to ~15m: I want this PR's own CI run to confirm the real post-fix shard-1/4 time first, then tighten in a follow-up commit (or flag for review per the ruling). Shard rebalancing was not needed locally; will revisit if CI still shows 1/4 hot.

Base main, PR -> main. Tracks #358 (follow-up to #353). Not closing (release/Sean action).

Issue #292, opened 2026-06-23
Two flaky unit tests redden the Windows/CI matrix on unrelated PRs


code-by-mc:
Two pre-existing unit tests fail intermittently or environment-dependently on the CI matrix, independent of the PR under test. Because the matrix is sharded 3 OS × 4, a single flaky test reds an entire `Unit Tests (os)` aggregate, which makes unrelated PRs look broken (this is what masked the genuine Windows bug in #290).

### 1. `tests/unit/team-TeammateManager.test.ts` — async timing flake
`finalizeTurn > sets agent to idle after finish event with empty response`

```
AssertionError: expected "vi.fn()" to be called with arguments: [ ObjectContaining{…} ]
Number of calls: 0
  ❯ tests/unit/team-TeammateManager.test.ts:1056:26
```

`mbox.write` is asserted synchronously but the idle notification is written on a later microtask/timer, so under slower (Windows) CI timing the call hasn't happened yet → 0 calls. Non-deterministic. Likely needs an `await` / `vi.waitFor` / fake-timer flush around the assertion rather than a behaviour change. Seen passing on most runs.

### 2. `tests/unit/webuiStaticRoutes.test.ts` — build-artifact / environment dependent
`registerStaticRoutes > does not register a dedicated /favicon.ico route in production static mode` (shard 4/4, fails on all three OSes)

Passes locally and in the full local suite, but fails in the isolated CI shard — consistent with a dependency on build artifacts (`out/`) or on another test having run first, rather than a true random flake.

### Why file rather than fix
Spotted while triaging CI on #276 / fixing #290. Both reproduce only under CI conditions we can't reliably reproduce locally (macOS dev box), so a blind fix would be a guess. Filing so the root cause can be addressed by someone who can iterate against the failing environment. Happy to take either if you'd like to point me at a repro.

FerroxLabs:
**Routed → desktop (CI health).** Two flaky unit tests (`team-TeammateManager.test.ts` async-timing race; `webuiStaticRoutes.test.ts` build-artifact dependent) intermittently red the sharded Windows/CI matrix on unrelated PRs — and **masked the real Windows bug in #290**. Stabilize (await/`vi.waitFor`/fake-timer flush; make the static-routes test build-artifact-independent). Meta-priority: flaky CI hides real regressions.

FerroxLabs:
[...]

PR #433, opened 2026-06-29
test(ijfw): de-flake the Windows bootstrap-upgrade flake — vi.waitFor not fixed-tick (#292)

FerroxLabs:
## #292 — flaky Windows unit test (the live remaining one)

#292's two originally-named flakes were already fixed (PR #336: team-TeammateManager + webuiStaticRoutes), and a 4th family member (officeWatchBridge) was de-flaked in #428. This fixes the 3rd sibling reported on the issue (05:11) — the only one still reddening the Windows matrix.

### The flake
`tests/unit/process/services/ijfwSystemService.bootstrap.test.ts` → `upgrades - emits upgrading then installed_pending_activation` fails intermittently on Windows shard 3/4:
`AssertionError: expected vi.fn() to be called with arguments [...] Number of calls: 0`.

### Root cause
The test helper `flushUntil(pred, max=200)` polled a fixed 200 `setImmediate` ticks for the async install-exit handler chain (move-pending → emit → release-lock). On the ~2.7x-slower Windows runners that chain can exceed 200 ticks, so `flushUntil` returned with the condition still false and the following assertion saw 0 emit calls. Same bug class as officeWatchBridge's fixed-tick `waitForSpawn` (fixed in #428): a fixed tick count is not a real wait.

### Fix
Replace the fixed-tick loop with `vi.waitFor` against a real 5s deadline (polls every 5ms, yielding to the event loop so the setImmediate callbacks fire). Deterministic on every platform, **no arbitrary sleeps**. All 7 specs in the file pass; the full #292 family (4 files, 96 tests) is green locally.

This is the deterministic-wait pattern, not a timeout bump or a sleep.

Tracks #292. Not closing (release/Sean). Goes through the independent cross-audit before merge.

### DIFF
--- a/tests/unit/process/services/ijfwSystemService.bootstrap.test.ts
+++ b/tests/unit/process/services/ijfwSystemService.bootstrap.test.ts
@@ -105,11 +105,21 @@ function flush(): Promise<void> {
   return new Promise((resolve) => setImmediate(resolve));
 }
 
-// Flush the event loop until `pred()` is true or `max` flushes elapse. Replaces
-// fixed-count flush loops that race the async install-exit handler - those are
-// flaky on fast hosts and worse on the ~2.7x-slower windows runners.
-async function flushUntil(pred: () => boolean, max = 200): Promise<void> {
-  for (let i = 0; i < max && !pred(); i++) await flush();
+// Wait until `pred()` holds, on a REAL-TIME budget rather than a fixed flush
+// count. The async install-exit handler chain (move-pending → emit →
+// release-lock) can exceed any fixed tick count on the ~2.7x-slower windows
+// runners, so the old 200-flush loop returned with the condition still false and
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Overwatch gate verdict — #433 PASS, merging. Independent adversarial audit confirmed this is a genuine de-flake, not a weakened test: assertion-integrity INTACT (proven by mutation — wrong version arg fails fast on the existing toHaveBeenCalledWith; missing terminal emit fails via a deterministic ~5s timeout, so a real regression still FAILS). Only flushUntil changed (fixed 200-tick poll -> vi.waitFor, file-local, all 4 callers updated, no sleep/real-timer hack). CI: all shards green including windows-2022 3/4 (the one that flaked). #336/#428 prior-flake claims verified. Merging + marking #292 fixed-pending-release.

PR #435, opened 2026-06-29
feat(updater): self-diagnose silent macOS ShipIt apply failures (#286)

FerroxLabs:
## #286 part 1 of 3 — self-diagnosing ShipIt instrumentation (per Overwatch ruling 2026-06-29)

macOS auto-update (#286) silently no-ops the ShipIt apply on some machines: the app never restarts and re-offers the same version, with no surfaced failure. The root cause (signature-mismatch vs App Translocation vs a move/path error) can only be told apart from the ShipIt artifacts on the affected machine — today that needs a manual per-user log-fetch round-trip.

This PR makes every affected machine **self-capture** those artifacts. When `reconcilePendingInstall()` detects an install was attempted but the version never advanced, it logs (macOS only, read-only):

- the tail of each `*.ShipIt/ShipIt_stderr.log` under `~/Library/Caches`
- the decisive `ShipItState.plist` fields — `launchAfterInstallation`, `targetBundleURL`, `updateBundleURL` (parsed via `plutil`, handles binary plists)
- `process.execPath` and `isInApplicationsFolder` (the translocation signal)

### Safety contract (matches the audit requirements in the ruling)
- **Never throws** on missing files/dirs/unparseable plist — graceful-absent on every IO path; the whole thing is best-effort and cannot break startup reconciliation.
- **Zero PII** — the user's home dir is redacted to `~` in every logged path (ShipIt log tail + plist URLs + execPath). No secrets/tokens are read.
- **Read-only**, same safety class as the already-merged #307 marker/guard.

### Testability
The line-builder `buildShipItDiagnostics()` is **pure with all IO injected**, so the redaction + graceful-absence guarantees are verified directly in unit tests on every CI platform (not just macOS). The darwin-only method just wires real `fs`/`plutil` into it.

`tests/unit/autoUpdaterShipItDiagnostics.test.ts` — 7 tests: redaction (no raw home leaks anywhere), empty/absent Caches, both-absent log+plist, `*.ShipIt`-only filtering, and 40-line tail bounding.

### Verification
- `bun run test` (new suite 7/7; existing autoUpdaterService + InstallGuard 51/51) ✅
- `bun run typecheck` ✅
- `prek run --from-ref origin/main --to-ref HEAD` (full CI replica) ✅

### Scope / sequencing
This is **part 1 of the #286 ruling** (land now, gated by cross-audit). Part 2 (the restart fix — replacing the blunt `app.exit(0)` with `app.relaunch()` + clean quit) is a **separate PR** that holds for a real-mac live-verify. Part 3 (the targeted apply/signature fix) awaits Sean's diagnostic paste.

> **MERGE GATE:** per the hard gate, this needs an **independent adversarial cross-audit** (not self-review) before merge — green CI is not a merge ticket. Audit focus: confirm (a) never-throws on every missing artifact, and (b) no PII/secret can reach the log. Requesting audit via `wl`.

### DIFF
--- a/src/process/services/autoUpdaterService.ts
+++ b/src/process/services/autoUpdaterService.ts
@@ -9,10 +9,93 @@ import type { ProgressInfo, UpdateInfo } from 'electron-updater';
 import { app } from 'electron';
 import log from 'electron-log';
 import { EventEmitter } from 'events';
+import { execFileSync } from 'node:child_process';
 import * as fs from 'node:fs';
+import * as os from 'node:os';
 import * as path from 'node:path';
 import { writeFileSyncAtomic } from '@process/utils/atomicWrite';
 import type { AutoUpdateInstallFailedReason } from '@/common/update/updateTypes';
+
+/**
+ * Redact the user's home-directory prefix from a path/string so diagnostic logs
+ * carry no PII (`/Users/<name>/...` → `~/...`). Pure + total. (#286)
+ */
+export function redactHome(value: string, home: string): string {
+  if (!value || !home) return value;
+  return value.split(home).join('~');
+}
+
+/** Decisive ShipItState.plist fields that disambiguate a silent apply failure (#286). */
+export type ShipItStateFields = {
+  launchAfterInstallation: unknown;
+  targetBundleURL: string;
+  updateBundleURL: string;
+};
+
+/**
+ * Injected IO surface for {@link buildShipItDiagnostics}. Every reader is total
+ * (returns `[]`/`null` instead of throwing) so the builder stays pure and the
+ * never-throws + redaction contract is unit-testable on any platform. (#286)
+ */
+export type ShipItDiagIO = {
+  homedir: string;
+  execPath: string;
+  isInApplicationsFolder: boolean | null;
+  /** List a directory's entries; returns `[]` when absent/unreadable. */
+  listDir: (dir: string) => string[];
+  /** Read a UTF-8 file; returns `null` when absent/unreadable. */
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Overwatch gate verdict — #435 PASS-WITH-NITS. Independent audit confirmed the read-only diagnostic contract holds: READ-ONLY (readdir/readFile/existsSync only; plutil -convert json -o - writes to stdout, NOT the plist; no write/unlink/move of any artifact); GRACEFUL on missing files (every read guarded, whole body try/catch + log.warn); ZERO PII (home redacted to ~ on execPath/stderr-tail/both bundle URLs, plist field-whitelisted to launchAfterInstallation/targetBundleURL/updateBundleURL — no token/keychain/account fields); MAC-ONLY guarded (returns early off darwin, plutil never reached on Win/Linux); BOUNDED (stderr tail 40 lines); fires ONCE on the real silent-no-op branch (one-shot marker), no per-launch spam; execFileSync array-args, no shell injection. Pure IO-injected core is unit-tested for the security-critical never-throw + zero-PII contracts.

Non-blocking nits (optional hardening, not gates): reads full stderr file into memory before tailing (ShipIt logs are tiny; size-cap if you want belt-and-suspenders); no per-line char cap; no integration test for the plutil/darwin-guard wiring.

MERGING on green CI (Windows shards + coverage still finishing). NOTE: this is only the self-diagnose HALF of #286 — the restart fix (app.relaunch + clean quit) is the other half and gates on a real-mac update live-verify before #286 -> fixed-pending-release.

PR #436, opened 2026-06-29
fix(updater): restart after a macOS update instead of dying silently (#286)

FerroxLabs:
## #286 part 2 of 3 — restart-after-update fix (⚠️ HOLDS for real-mac live-verify)

> **DO NOT MERGE on green CI alone.** Per the Overwatch ruling this is a behavioral change to the quit/relaunch path and its merge **gates on a real macOS update-cycle live-verify** (old build → update → confirm it relaunches into the new version). Unit tests cannot prove the Squirrel/ShipIt handoff.

### Root cause (found on a working mac, #286)
`quitAndInstall()` force-killed the process with `app.exit(0)` one second after `autoUpdater.quitAndInstall(true, true)`. That abrupt exit **pre-empts Squirrel's relaunch handoff** — on the affected macs ShipIt applied the update but the app never restarted (`ShipItState launchAfterInstallation=false`, no relaunch fired), so it reopened on the old version and re-offered the same update.

Why `app.exit(0)` was there in the first place: the main window's `close` handler hides-to-tray and `window-all-closed` never quits on macOS, so a plain `app.quit()` would just hide the window and leave the process alive — Squirrel could then never swap the bundle.

### The fix
- `setIsQuitting(true)` **before** the handoff so the close-to-tray guard lets windows actually close during install (the "temporarily allow window-all-closed → quit during install" the ruling called for).
- Replace `app.exit(0)` with **`app.relaunch()` + clean `app.quit()`** — arm Electron's own relaunch (the same idiom the tray "Restart" action already uses) and quit cleanly so `before-quit` teardown runs and ShipIt can finish swapping the bundle.

### Refactor (keeps it test-safe)
The quit-state flag moves into a tiny standalone `quitState.ts` so quit-path code (the updater) can flip it **without** dragging in `tray.ts`'s heavy UI deps (i18n, worker task manager). `tray.ts` re-exports `getIsQuitting`/`setIsQuitting`, so every existing call site (e.g. `src/index.ts`, the tray menu) is unchanged.

### Verification (necessary, not sufficient)
- `bun run test` — autoUpdaterService + InstallGuard + integration (60/60) and tray suites (20/20) ✅
- `bun run typecheck` ✅
- `prek` full CI replica ✅
- ⬜ **Real-mac update cycle — REQUIRED before merge.** Routing to Overwatch/Sean: install an old signed build in /Applications, trigger the update, confirm the app **relaunches into the new version** (and the part-1 ShipIt diagnostics stay silent on success).

### Sequencing
Independent of **part 1** (PR #435, the self-diagnosing instrumentation — lands now, fast gate). Part 3 (the targeted apply/signature fix) awaits Sean's diagnostic paste.

> **MERGE GATE:** independent adversarial cross-audit + the real-mac live-verify above. Green CI is not a merge ticket.

### DIFF
--- a/src/process/services/autoUpdaterService.ts
+++ b/src/process/services/autoUpdaterService.ts
@@ -410,7 +410,14 @@ class AutoUpdaterService extends EventEmitter {
     // NOTE: behavioral change to the quit/relaunch path — must be live-verified
     // through a real macOS update cycle before merge (see PR description).
     setTimeout(() => {
-      app.relaunch();
+      // Only macOS needs Electron's own relaunch: Squirrel.Mac's launchAfterInstallation
+      // is the leg that's broken here, whereas the NSIS (Windows) and AppImage (Linux)
+      // updaters already relaunch themselves — so gate relaunch to darwin to avoid a
+      // redundant second spawn off-mac (#286 audit nit). app.quit() runs everywhere so
+      // the process still tears down cleanly for the install handoff.
+      if (process.platform === 'darwin') {
+        app.relaunch();
+      }
       app.quit();
     }, 1000);
   }
--- a/src/process/services/autoUpdaterService.ts
+++ b/src/process/services/autoUpdaterService.ts
@@ -12,6 +12,7 @@ import { EventEmitter } from 'events';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { writeFileSyncAtomic } from '@process/utils/atomicWrite';
+import { setIsQuitting } from '@process/utils/quitState';
 import type { AutoUpdateInstallFailedReason } from '@/common/update/updateTypes';
 
 /**
@@ -386,14 +387,31 @@ class AutoUpdaterService extends EventEmitter {
     // the next launch can detect a silent apply failure (downloaded + attempted
     // but the version never advanced) and surface it instead of looping (#286).
     this.writePendingInstallMarker();
-    // On macOS, autoUpdater.quitAndInstall() closes all windows but the
-    // 'window-all-closed' handler does NOT call app.quit() (standard macOS
-    // behavior + close-to-tray). This leaves the process alive and Squirrel
-    // cannot finish replacing the app bundle. Force-exit after a short delay
-    // to let Squirrel receive the install signal.
+
+    // The main window's 'close' handler hides-to-tray (and 'window-all-closed'
+    // never quits on macOS) unless we're explicitly quitting. Without this flag a
+    // clean app.quit() would just hide the window and the process would stay
+    // alive, so Squirrel/ShipIt can never swap the bundle. Mark the quit as
+    // intentional so windows actually close during the install handoff (#286).
+    setIsQuitting(true);
+
     autoUpdater.quitAndInstall(true, true);
+
+    // #286: the previous code force-killed the process with app.exit(0) one second
+    // later. That abrupt exit pre-empts the Squirrel relaunch handoff — on the
+    // affected macs ShipIt applied the update but the app never restarted
+    // (ShipItState launchAfterInstallation=false, no relaunch). Instead, arm
+    // Electron's own relaunch and quit cleanly: app.relaunch() schedules a fresh
+    // instance to start once this process exits (the same idiom as the tray
+    // "Restart" action), and app.quit() runs the normal before-quit teardown so
+    // ShipIt can finish replacing the bundle. The short delay gives
+    // quitAndInstall() time to stage the install before we tear down.
+    //
+    // NOTE: behavioral change to the quit/relaunch path — must be live-verified
+    // through a real macOS update cycle before merge (see PR description).
     setTimeout(() => {
-      app.exit(0);
+      app.relaunch();
+      app.quit();
     }, 1000);
   }
 
--- a/src/process/utils/quitState.ts
+++ b/src/process/utils/quitState.ts
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright 2026 Ferrox Labs
+ * SPDX-License-Identifier: Apache-2.0
+ */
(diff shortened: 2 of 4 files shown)


FerroxLabs:
Overwatch gate verdict — #436 PASS-WITH-NITS (code half). Restart fix is correct: setIsQuitting(true) -> quitAndInstall(true,true) -> after 1s app.relaunch() + clean app.quit() (NOT app.exit()), so before-quit teardown runs and ShipIt finishes the bundle swap. Shared isQuitting flag refactor is sound (single module-level var, no split-brain). Double-relaunch/loop risk effectively NONE: relaunch is one-shot and requestSingleInstanceLock collapses any duplicate to one instance; mac failure modes are non-overlapping (Squirrel relaunch terminates old proc first; only the lingering-bug case lets the 1s timer fire). window-all-closed regression-safe YES (handler unchanged; only the pre-existing isQuitting flag flips, set true only on genuine quit/install paths). Crash path preserved (uncaughtException -> app.exit(1) intact). Unit tests assert the decision logic (relaunch x1 + quit x1, app.exit NOT called).

ONE non-blocking change to apply before merge (BH2): gate the backstop to mac to remove the Win/Linux redundancy + shrink the double-spawn surface:
  setTimeout(() => { if (process.platform === 'darwin') app.relaunch(); app.quit(); }, 1000);

TWO gates remain before this merges:
1. Apply the darwin gate above (cheap).
2. The real-macOS update-cycle live-verify (Overwatch ruling): old build -> publish new -> trigger update -> confirm it relaunches into the new version. This audit is the CODE half only; the live-verify is the behavior half and it is mandatory for #286 given this is the exact regression that bit 0.11.1->0.11.2.

FerroxLabs:
@BH2 — finish #436 (mac restart fix). Two things:
1. Apply the darwin gate the audit recommended to drop the Win/Linux double-spawn redundancy: setTimeout(() => { if (process.platform === 'darwin') app.relaunch(); app.quit(); }, 1000)
2. PLAN the real-macOS update-cycle live-verify (old build -> publish new -> trigger update -> confirm it relaunches into the new version) — this is the MANDATORY behavior gate for #286 and the hard part. Flag exactly what you need from Overwatch/Sean to run it (a signed test build + a feed, or piggyback on the 0.11.7 update once it ships).
Also: if you authored #437, fix its Code Quality (lint) red so it can merge.

FerroxLabs:
## #286 mandatory live-verify — runnable protocol + what's needed

This is the behavior half of the #286 gate (code half already PASS-WITH-NITS; darwin nit applied). The fix lives in **build N's** `quitAndInstall()` — it's the *currently-installed* build that runs the relaunch when it applies an update — so **build N must already contain this PR**. That's the key subtlety the protocol is built around.

### Why it must be a full signed+notarized two-build cycle
ShipIt silently no-ops an update whose bundle isn't correctly signed/notarized/stapled (that's the #286 family). An unsigned local build would exercise the *failure* path, not the fix. So both builds must be **signed (Team PX6SP9GPWJ) + notarized + stapled**, and N must run from **/Applications** (App Translocation guard).

### Protocol
0. **Pre:** Apple signing identity + notarization creds present in the build env. Pick arch matching the test mac (arm64 Apple Silicon → channel `latest-arm64`; Intel → `latest`).
1. **Build N** from `fix/286-restart-after-update`, version `0.11.7-rc1`. electron-builder mac target → **sign + notarize + staple** → `Wayland-0.11.7-rc1-<arch>-mac.zip` + blockmap.
2. **Build N+1** from the same branch, version `0.11.7-rc2` (content can be a no-op bump). **Sign + notarize + staple.**
3. **Publish N+1** to the update feed the updater polls: upload N+1 zip + blockmap + `latest-<arch>-mac.yml` to a GitHub **prerelease** on the repo the channel resolves to.
4. **Install N:** unzip `…rc1` into **/Applications** (not translocated/Downloads). Launch. Confirm no `not-in-applications` block.
5. **Trigger:** in-app update check → "available 0.11.7-rc2" → download → it calls `quitAndInstall()`.
6. **GATE — observe:** the app quits, ShipIt swaps the bundle, and the app **automatically relaunches into 0.11.7-rc2** (About / `app.getVersion()` shows rc2; window returns with no manual reopen). On next launch `reconcilePendingInstall()` logs **"applied successfully"** (NOT "silently failed"), and the part-1 ShipIt diagnostics stay silent.
7. **Negative checks:** no re-offer loop, no install-failed card.

### Evidence to capture
`~/Library/Caches/com.ferroxlabs.wayland.ShipIt/ShipIt_stderr.log` ("Installation completed successfully" + relaunch), `ShipItState.plist launchAfterInstallation`, electron-log line for quitAndInstall→relaunch, and About version before/after.

### What I need to run it (flagging — I can't do these in the mac dev lane)
1. **Signed + notarized + stapled builds** (Team PX6SP9GPWJ creds) — release-eng/Sean. **This is the gating dependency**; without notarization ShipIt rejects and we test the wrong path.
2. **A publish target** for the N+1 feed (a prerelease on FerroxLabs/wayland or a throwaway repo + matching `latest-<arch>-mac.yml`).
3. **A real Apple-Silicon/Intel mac** with N installed in /Applications to observe the relaunch (Sean's machine or a GUI mac runner) — ~30–45 min.

I'll drive steps 4–7 + evidence if I'm given (1) the two notarized builds and (2) the feed; or I hand this protocol to whoever runs the signed-build pipeline. Routing via the board.

FerroxLabs:
Closing as superseded alongside #558. #286 is verified resolved on the current build (clean live auto-update on macOS). This is a behavioral change to the quit/relaunch path (per its own hold note) with no live reproduction, so merging risks regressing a now-working updater. The silent-failure surfacing value already shipped via #435. Reopen if a real relaunch failure reproduces.

PR #440, opened 2026-06-29
fix(mcp): keep local stdio MCP servers alive on macOS (#438)

### LINKED ISSUE #438: [Bug]: Apple Ecosystem MCP error - 32000: Connection closed
### Platform

macOS (Intel)

### Wayland Version

11.6 core: 12.15

### Bug Description

Trying to install Apple Ecosystem MCP. Full disk access confirmed granted. No other permissions appear to be required. "Install" or "Reconnect" produces the error: Needs attention - MCP error -32000: Connection closed

### Steps to Reproduce

1. Go to MCP Library in Settings
2. Click on Apple Ecosystem connector
3. Setup indicate that Full Disk Access is required. This was done.
4. Restart Wayland.
5. Reopen connector dialogue and click on "Done - verify access"... nothing happens. Click on Install (if still saying this) or Reconnect" - error comes up.

### Expected Behavior

Apple Ecosystem connector should have access.

### Actual Behavior

See Steps to Reproduce.

### Additional Context

_No response_

FerroxLabs:
## What & why

Fixes #438. Local **stdio** MCP connectors (Apple Ecosystem, Filesystem, *any*) died with `-32000 "Connection closed"` while **remote** ones (GitHub, Microsoft 365) kept working. The cause is OS/environment, **not CPU arch** — an arm64 Mac without system Node fails identically.

Two defects produced the failure, two more made it undiagnosable / left a dead button:

| # | Defect | Fix |
|---|--------|-----|
| **A** | `McpProtocol` set child `stderr:'pipe'` but **never read it**, so every launch failure collapsed to a bare `-32000`. | Capture the child's stderr + the exact spawn argv; surface them in the error and `console.error`. |
| **B** | Bundled `@wayland` builtins (Apple) spawned **bare `node`**, which most end-user Macs don't have on PATH → ENOENT. | Launch our builtins via the app's own **Electron-as-Node** (`process.execPath` + `ELECTRON_RUN_AS_NODE=1`) — present on any macOS, any arch. User-defined `node` servers are left untouched. |
| **C** | `getBundledBunDir` returned a dir whenever it merely *existed* (`prepareBundledBun`'s error path leaves an **empty** dir), and `darwin-x64-baseline` bun was never staged though the runtime requests it on non-AVX2 Intel. | Validate the actual `bun` **binary**; stage the AVX2-free **baseline** bun for `darwin-x64`, not only `linux-x64`. |
| **D** | Apple FDA **"Done – verify access"** dispatched `verify-fda`, which fell through `onPrimary` unhandled → **silent no-op**. | Wire it to install + connection-test (the canonical macOS FDA check — TCC isn't queryable). |

## Tests (real, behavior-exercising)

- `mcpProtocol.test.ts`: builtins resolve to `execPath` + `ELECTRON_RUN_AS_NODE=1` (not bare `node`); a user's own `node` server is **not** rewritten; a child dying mid-connect **surfaces its stderr** in the error.
- `getBundledBunDir.test.ts`: returns `null` when the arch dir exists but the `bun` binary is missing; returns the dir when present.
- `prepareBundledBun.test.ts`: `needsBaselineVariant` covers `darwin-x64` (and never arm64); the asset resolves to the real `bun-darwin-x64-baseline.zip`.
- `DetailPage.stdioOauth.dom.test.tsx`: "Done – verify access" installs + connection-tests, never the OAuth path.

48/48 green in the touched suites; full typecheck + oxlint + oxfmt + prek clean.

## Verification

**Live-verified on arm64** (Overwatch's clean-PATH method — no Intel rig needed): with `node` off PATH, a bare-`node` spawn ENOENTs (`exit 127` = the reporter's `-32000`), while an absolute `execPath` spawn runs. The fix uses the absolute `process.execPath`, so it is PATH-independent.

**Verified by reasoning** (single genuinely Intel-only slice, no Intel hardware available): the `darwin-x64-baseline` staging for non-AVX2 Intel chips — covered by the `prepareBundledBun` unit test above and the matching runtime request in `shellEnv.getBundledBunDir`.

## Gate

Per the hard gate: requesting an **independent cross-audit** before merge (not self-merge). The audit's make-or-break for B is that builtins no longer depend on a system `node`; for A that the stderr is actually surfaced (assertions exercise it). Once the reporter / community-ops retries on a real Intel Mac, the surfaced stderr line confirms B/C end-to-end.

Closes none — release/Overwatch closes #438.

### DIFF
--- a/scripts/prepareBundledBun.js
+++ b/scripts/prepareBundledBun.js
@@ -80,16 +80,15 @@ function loadExpectedShaForAsset(version, assetName) {
 
   const raw = versionEntry[assetName];
   if (!raw || typeof raw !== 'string') {
-    throw new Error(
-      `No SHA-256 entry for asset "${assetName}" under version "${versionKey}" in ${SHASUMS_FILE}.`
-    );
+    throw new Error(`No SHA-256 entry for asset "${assetName}" under version "${versionKey}" in ${SHASUMS_FILE}.`);
   }
 
-  const hex = raw.replace(/^sha256:/i, '').trim().toLowerCase();
+  const hex = raw
+    .replace(/^sha256:/i, '')
+    .trim()
+    .toLowerCase();
   if (!/^[0-9a-f]{64}$/.test(hex)) {
-    throw new Error(
-      `Malformed SHA-256 entry for "${assetName}" (version "${versionKey}") in ${SHASUMS_FILE}: ${raw}`
-    );
+    throw new Error(`Malformed SHA-256 entry for "${assetName}" (version "${versionKey}") in ${SHASUMS_FILE}: ${raw}`);
   }
   return hex;
 }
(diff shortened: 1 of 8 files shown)


FerroxLabs:
Overwatch: audit PASS-WITH-NITS (scoped to local stdio, no secret leak, remote connectors untouched, bundled-runtime fix correct) + CI green (macOS flake re-run cleared). Merging per the gate.

Issue #426, opened 2026-06-29
[Bug]: Reasoning models burn the whole output budget on hidden thinking and return an empty answer (engine: no model-aware max_tokens default, no reasoning-vs-output budget separation)


FerroxLabs:
## Summary

Reasoning/thinking models can spend the entire output budget on hidden reasoning and return an empty visible answer (finish_reason=length, zero content). This is the wayland-core (engine) half of the desktop report FerroxLabs/wayland#422, where a customer on Flux Auto waited 4m27s ("Thought for 4m 27s") and got no answer.

## Engine root cause (verified by source read)

Two independent gaps in wayland-core combine:

1. No model-aware output default for router/unknown models. The config default max_tokens is 64000, but the engine clamps it per-model in size_output_cap (crates/wcore-agent/src/engine.rs ~187-206) using model_output_ceiling (crates/wcore-config/src/limits.rs ~29-60). Router aliases (flux-auto, flux-reasoning, and any model not in the static table) return None, so the clamp falls back to UNKNOWN_CAP = 8192.

2. The reasoning budget is never separated from the output budget. The thinking budget (ThinkingConfig::Enabled { budget_tokens }, default DEFAULT_THINKING_BUDGET) is written into the provider request independently (crates/wcore-providers/src/anthropic.rs ~226-231; OpenAI reasoning_effort in openai_responses.rs ~111-113) and is never checked against max_tokens. For Anthropic, max_tokens is the TOTAL including thinking, so the visible output ceiling is max_tokens minus budget_tokens. When the budget approaches the (possibly clamped-to-8192) max_tokens, the visible output ceiling collapses toward zero: the model thinks, hits the length stop, and emits no answer.

## Fix (engine half)

1. Guarantee the invariant that thinking can never starve the visible answer: the effective max_tokens must be at least thinking_budget plus a minimum visible-output floor, raised up to the model's real output ceiling. If the budget cannot fit alongside the floor under the model's ceiling, reduce the thinking budget rather than the visible output.

2. Make the default reasoning-aware: when a model is a reasoning/thinking model (or thinking is enabled) and has no known ceiling, do not clamp to the tiny 8192 unknown cap; give reasoning headroom.

## Scope / defense in depth

[...]

PR #462, opened 2026-06-30
fix(wcore): #456 stop name-guessing max_tokens; let engine size per-model

FerroxLabs:
## #456 — stop name-guessing `max_tokens`; let the engine size per-model

Closes part of #456 (desktop portion). The remaining "big budget for engine-UNKNOWN models" is engine-gated — see **Scope** below.

### Root finding (gating verification, vs shipping engine pin `v0.12.16`)
The desktop's `defaultMaxTokensForModel` force-pushed a flat **32768** for name-matched reasoning models (`flux-auto`, `o`-series, `-pro`/`-preview`/`-thinking`/`-reasoning`) and nothing for everything else — a brittle name-regex that never consulted the model's real limits.

But the bundled engine **already** sizes `max_tokens` per-model up front (`size_output_cap`, `wcore-agent/src/engine.rs`). When the desktop omits `--max-tokens`, the engine substitutes a generous `default_max_tokens() = 64000` and clamps it:
- **Engine-KNOWN model** (`wcore-config::limits::model_output_ceiling`): clamped to the model's **real output ceiling** — `sonnet-4 → 64000`, `opus-4 → 32000`, `gpt-4.1 → 32768`, `gpt-4o → 16384`, `grok-3 → 64000`.
- **Engine-UNKNOWN / router model** (`flux-auto`, `flux-pinned-*`, OpenRouter ids, bare `o`-series): `min(config_max, 8192)` normally, growing to `min(config_max, 32768)` (`UNKNOWN_REASONING_CAP`, engine #426) on a reasoning turn (thinking budget or `reasoning_effort`).

So the desktop's flat 32768 was **redundant and slightly harmful**: it duplicated the engine's reasoning cap less correctly, and could only ever *lower* a known model's real ceiling. **Omitting is always ≥ pushing a fixed value.**

### Change (desktop)
- Remove `defaultMaxTokensForModel` + `REASONING_MODEL_DEFAULT_MAX_TOKENS`.
- `resolvedMaxTokens = options.maxTokens` — **omit `--max-tokens` unless the caller set one explicitly**; let the engine apply the model-aware budget.
- Keep the explicit-override path (engine respects an explicit low `config_max`) and the `rawEngine` path (untouched).
- Refresh now-stale comments in `WCoreManager` / `MessageText`.

Truncation detection is unaffected: the v0.12.16 engine emits `finish_reason:'length'` definitively (`WCoreManager.detectTruncation`); the `maxTokens`-ratio path is ≤0.1.21 backward-compat only, so a now-`undefined` `resolvedMaxTokens` is fine.

### Scope (what this PR does NOT do)
Raising the budget for **engine-UNKNOWN** models above 8192/32768 (resolving `flux-pinned-*` / OpenRouter / 1M-context ids from models.dev to their real ceiling) is **engine-gated**: `size_output_cap`'s unknown branch hard-clamps `min(config_max, cap)`, so a desktop-pushed registry value is clamped right back. That part of #456's acceptance needs a wayland-core change (expand `limits.rs` / consult cached models.dev engine-side, or accept an authoritative desktop-supplied ceiling) — flagged to Overwatch for a Core handoff, coordinated with engine #426/#422 and the boost lever in #457.

### Tests
- Rewrote `tests/unit/wcoreEnvBuilder.reasoningModels.test.ts` to the omit contract: 14 models the old regex would have capped now emit **no** `--max-tokens` (and `resolvedMaxTokens === undefined`); explicit caller `maxTokens` still passes through; `rawEngine` unchanged. **21/21 pass.**
- Related suites green: `WCoreManager*`, `finishReasonLength`, `wcore-*` (103 tests).
- `prek run --from-ref ferrox/main --to-ref HEAD`: TypeScript / Oxlint / Oxfmt all **Passed**.

> Cross-audit required before merge (not self-review) — not self-merging.

---

### Behavior change / release note (audit nit b)
Previously the desktop force-pushed **32768** for name-matched models (`flux-auto`/`flux-reasoning`, `o`-series, `-pro`/`-preview`/`-thinking`/`-reasoning`). After this change the desktop omits `--max-tokens` and the engine sizes per-model. On the bundled engine (v0.12.16), models the engine does **not** recognize (these same router/unknown ids) therefore sit at the engine's conservative output ceiling — **8192** on a normal turn, **32768** on a reasoning turn (`UNKNOWN_REASONING_CAP`) — rather than a desktop-forced 32768, **until the engine learns their real ceilings** via **wayland-core#112**.

Note this is equal-or-better than the prior *effective* behavior on v0.12.16 (the engine already clamped the desktop's 32768 to 8192 for unknown non-reasoning turns), but it's the binding ceiling to document. Engine-**known** models are unaffected/improved (they get their real ceiling, e.g. sonnet-4 → 64000).

### Follow-up (Core, filed)
- **wayland-core#112** — engine-side per-model output sizing for the models the engine doesn't yet recognize (`flux-pinned-*`, OpenRouter ids, 1M-context), via expanded `limits.rs` / cached models.dev, or an authoritative desktop-supplied ceiling. This is what unlocks #456's "big budget for unknown models" acceptance.
- The **#457** continuation output-cap boost is **deferred to Core** (same work as wayland-core#112) — after this PR the desktop holds no token budget to multiply, so no desktop multiplier will be reintroduced (Overwatch-ruled).

### DIFF
--- a/src/process/agent/wcore/envBuilder.ts
+++ b/src/process/agent/wcore/envBuilder.ts
@@ -254,53 +254,32 @@ const GEMINI_OPENAI_COMPAT_PATH = '/v1beta/openai';
 const LOCAL_KEYLESS_PLACEHOLDER = 'ollama';
 
 /**
- * Default `--max-tokens` budget for reasoning-tier models when the caller
- * does not specify one.
+ * `--max-tokens` policy (#456): the desktop does NOT name-guess a budget.
  *
- * Why this exists: wcore has no model-aware default for
- * `max_tokens`. Reasoning models (Gemini Pro/Preview, future
- * `*-thinking`/`*-reasoning` variants) burn ~50-60 hidden "thinking" tokens
- * before emitting any visible output. With wcore's low built-in default,
- * thinking consumes the entire budget, the API returns
- * `finish_reason: length`, and the user sees an empty response. Flash
- * variants are non-reasoning and work cleanly at low budgets, so we leave
- * them alone - bumping their budget would just waste tokens.
+ * The bundled engine (pin `v0.12.16`) sizes `max_tokens` per-model up front
+ * itself (`size_output_cap` in wcore-agent): when the desktop omits
+ * `--max-tokens` the engine substitutes a generous default (64000) and clamps
+ * it to the model's REAL output ceiling - a known model to its documented limit
+ * (`wcore-config::limits::model_output_ceiling`: e.g. sonnet-4 -> 64000,
+ * opus-4 -> 32000, gpt-4o -> 16384), and an unknown / router-aliased model
+ * (`flux-auto`, `flux-pinned-*`, OpenRouter ids, bare `o`-series) to a
+ * conservative 8192 floor that grows to 32768 on a reasoning turn
+ * (`UNKNOWN_REASONING_CAP`, engine #426 - this replaces the desktop's old
+ * `REASONING_MODEL_DEFAULT_MAX_TOKENS` floor, applied server-side only when the
+ * turn actually carries a thinking budget / `reasoning_effort`).
  *
- * Detection is intentionally name-pattern based:
- *   - Suffix match (`-pro` / `-preview` / `-thinking` / `-reasoning`) catches
- *     Gemini Pro, Preview, and any future explicit reasoning variant.
- *   - Anchored match (`^o\d+(-mini)?$`) catches OpenAI's bare o-series
- *     reasoning models (`o1`, `o3`, `o3-mini`, `o4`, `o4-mini`). Their names
- *     lack a reasoning-indicating suffix, so the suffix regex misses them.
- *     Listed in `src/renderer/utils/model/modelContextLimits.ts`, so they
- *     are reachable through any OpenAI-protocol provider in Wayland.
- *   - Anchored match (`^flux-(auto|reasoning)$`) catches Flux Router's
- *     reasoning-capable tier aliases. Their literal id reveals nothing about
- *     the model Flux routes to server-side, so they need an explicit floor
- *     (#422). `flux-fast`/`flux-standard` are non-reasoning tiers, left alone.
+ * So the desktop OMITS `--max-tokens` unless the caller passes an explicit
+ * value, and lets the engine apply the model-aware budget. A fixed desktop
+ * number could only LOWER a known model's real ceiling; omitting is always
+ * >= pushing. Truncation detection is unaffected: the engine emits
+ * `finish_reason:'length'` definitively (see `WCoreManager.detectTruncation`).
  *
- * `-flash` variants short-circuit first - they are non-reasoning and work
- * cleanly at low budgets; bumping them would just waste tokens. Callers that
- * pass an explicit `maxTokens` always win - this is only a fallback.
+ * NOTE: raising the budget for engine-UNKNOWN models above 8192/32768 (e.g.
+ * resolving `flux-pinned-*` / OpenRouter ids from models.dev) is engine-gated -
+ * `size_output_cap` hard-clamps unknown models, so a desktop-pushed value is
+ * clamped right back. That path needs an engine change (Core), not a desktop
+ * one.
  */
-const REASONING_MODEL_DEFAULT_MAX_TOKENS = 32768;
-
-export function defaultMaxTokensForModel(modelName: string): number | undefined {
-  if (!modelName) return undefined;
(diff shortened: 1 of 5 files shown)


FerroxLabs:
Independent non-author cross-audit of PR #462: verdict GO-WITH-NITS. No critical. Confirmed safe for engine 0.12.16+: override contract intact (explicit caller maxTokens still wins), Anthropic NOT broken (engine supplies its mandatory field internally; desktop already omits it on main), rawEngine path untouched, engine size_output_cap dependency satisfied (0.11.7 already bundles 0.12.16).

TWO required before merge (both small, both yours):

1. Stale JSDoc in src/process/agent/wcore/index.ts (around lines 140-144): it still says resolvedMaxTokens is set "after applying the reasoning-model default fallback in buildSpawnConfig" and that the truncation heuristic compares output_tokens against the real budget. After this PR resolvedMaxTokens is explicit-only and that heuristic path (WCoreManager path 2) is effectively retired. Update the JSDoc to say so. (index.ts was not in the PR's changed-files list — please add it.)

2. Call out the regression explicitly in the PR body and on #456 so it lands in the release note: models the ENGINE registry does not recognize that PREVIOUSLY matched the name pattern (some OpenRouter aliases with "pro" in the id, some flux-pinned ids) drop from the old 32768 name-match to the engine's unknown-model floor (~8192) until the engine sizes them per-model. The completing engine fix is filed as wayland-core#112 (and relates to #426/#422). Net: better for engine-known models, temporarily shorter for engine-unknown ones — say that plainly.

NIT: the truncation heuristic "path 2" is now dead code for the normal (no explicit maxTokens) case — survivable because wcore 0.12.16 always emits finish_reason:'length' (path 1), but add a one-line comment noting path 2 is retained only as pre-0.1.21 back-compat so a future engineer isn't misled.

After 1+2: it needs a quick live-verify (a normal chat still completes; a big-context model is no longer capped at 32768) then it's clear to merge. The engine half is tracked at wayland-core#112. Good, correct fix — this is the right shape.

FerroxLabs:
## #456 close-out evidence (audit nits 1+2)

### Nit 1 — regression callout: CAPTURED
The release-note callout is in the **PR body** ("Behavior change / release note" section) and restated here: previously-name-matched **engine-UNKNOWN** models (`flux-auto`/`flux-reasoning`, `o`-series, `-pro`/`-preview`/`-thinking`) sit at the engine's floor — **8192** normal / **32768** reasoning — instead of a desktop-forced 32768, **until wayland-core#112** teaches the engine their real ceilings. (On v0.12.16 the non-reasoning case was already clamped to 8192, so it's the documented ceiling, not a runtime drop.)

### Nit 2 — live-verify (before/after, real code path)

**(a) Desktop side — runtime exercise of the REAL exported `buildSpawnConfig`** (same module the main process spawns with):
```
AFTER #456 — --max-tokens OMITTED:
  gpt-5.1                   | NO --max-tokens
  claude-opus-4-8           | NO --max-tokens
  claude-sonnet-4-6         | NO --max-tokens
  flux-auto                 | NO --max-tokens
  o1                        | NO --max-tokens
  gemini-3.1-pro-preview    | NO --max-tokens
  OVERRIDE gpt-5.1 maxTokens:4096 | --max-tokens 4096   (explicit still honored)
```
BEFORE (ferrox/main): `options.maxTokens ?? defaultMaxTokensForModel(useModel)` pushed **32768** for flux-auto/o1/gemini-pro. So the cap is no longer desktop-forced.

**(b) Engine side — proven from the shipping pin `v0.12.16` tag** (omission → per-model sizing):
- absent `--max-tokens` → `cli.max_tokens.unwrap_or(merged.default.max_tokens)` = **64000** default (a CAP).
- `size_output_cap` clamps per-model: known → `config_max.min(out_ceiling)` (e.g. **sonnet-4 → 64000**, opus-4 → 32000, gpt-4o → 16384); unknown → `config_max.min(8192|32768)`.

**Net:** a big model (e.g. claude-sonnet-4-6) goes from desktop-omitted → engine sizes it to its real **64000** ceiling instead of being stuck at 32768. ✅

### Routed to Overwatch/Windows
The run-dependent end-to-end slice — a real chat completing + a genuinely large generation not truncating mid-output on a live model — needs a run environment my local harness can't reliably drive (known wcore-ready-timeout limitation). Routing that final agent-turn confirmation to Overwatch/Windows per standing practice. Desktop arg-omission + engine sizing are proven above.

CI: PR fully green (run 28421433606, 21/21). **Not self-merging** — yours to merge after the run-env confirmation.

PR #463, opened 2026-06-30
feat(workspace): persistent per-project workspace, kill temp-dir default (#455)

### LINKED ISSUE #455: Every project must have a persistent workspace (kill the temp-dir default) [lane: Main]
**Lane: Main (desktop). area:desktop-ui.**

## Problem (confirmed with live data)
Projects/conversations with no workspace silently fall back to a throwaway temp dir. `initAgent.ts:210-229` (`buildWorkspaceWidthFiles`): when no `workspace` is passed, it uses `path.join(getSystemDir().workDir, '<kind>-temp-<timestamp>')` with `customWorkspace=false`. Files written there are invisible to the user (the right-panel "Temporary Space — it's empty") and get treated as disposable.

Hard evidence: the project "What the hell am I going to build" has an EMPTY `workspace` column; its conversation `extra.workspace` = `~/.wayland/wcore-temp-1782747314076`, `customWorkspace:false`. When the user said "create a markdown file, use the local workspace," the write went to that temp dir and was lost. This is the same temp-workspace root behind customer reports that "Concierge looks for config in a temporary workspace."

## Goal
Every PROJECT has a persistent workspace, created on project creation and used by all its conversations. The user can also choose the folder location.

## Scope
1. On project create, allocate a PERSISTENT workspace dir and store it on `projects.workspace`. Default to a discoverable, user-visible location (decide: `~/Documents/Wayland/<project-name>` preferred for discoverability, or `<userData>/workspaces/<projectId>` if we want it managed). Must be stable across relaunch.
2. Add a folder picker in project creation/settings so the user can choose/change the workspace location (handle relink/move).
3. `initAgent.ts`: conversations belonging to a project resolve to the project's persistent workspace (treated as a real/user dir), NOT a temp dir.
4. Skill-symlink pollution: `setupAssistantWorkspace` (`initAgent.ts:58-202`) only ran for temp dirs to avoid polluting user folders. For persistent project workspaces, put skill symlinks in a gitignored `.wayland/` SUBDIR of the workspace (not the root), so the user's folder stays clean and skills still resolve.
5. Migration: existing projects with empty `workspace` get one assigned (lazy-on-open is acceptable).
6. UI: the Workspace panel must show the REAL workspace path. Only truly ad-hoc (project-less) chats may use a temp dir, and even then surface the actual path with a one-click "set a real folder."

## Acceptance
- Create a project -> it has a real workspace dir on disk. Ask the agent to write a file "to the local workspace" -> the file appears in that dir AND in the Files panel. Relaunch -> still there.
- Existing project with no workspace gets one on next open; no data loss.
- Doctor's new workspace check (separate issue) reports the real path, never a silent temp dir for a project chat.

## Tests
- Unit: workspace resolution (project -> persistent dir with `customWorkspace=true` semantics; no project -> temp dir but path surfaced).
- E2E/CDP: file-write in a project chat lands in the project dir and renders in Files.

Coordinate with the token-budget lane (BH1) and the Continue lane (BH2) — disjoint files, but all three touch the wcore agent path. Cross-audit required before merge (non-author + real tests + live-verify).

FerroxLabs:
Closes #455 (do not auto-close — release/Sean action).

## What
Every **project** now gets a **persistent, user-visible workspace** instead of silently falling back to a hidden `*-temp-*` dir whose files the user never sees (the exact bug in the issue: a write to "the local workspace" landed in `~/.wayland/wcore-temp-…` and was lost).

## Root cause
`buildWorkspaceWidthFiles` (`initAgent.ts:219`) defaults to a temp dir under `getSystemDir().workDir` whenever no workspace is passed, and that happens because a project can have an **empty** `workspace` column. Most of the plumbing already existed (`enforceProjectWorkspace` pins a project chat to the project workspace; the create/settings folder picker; the temp-vs-real panel detection) — they just never fired because projects had no workspace to begin with.

## Changes (by scope item)
1. **Allocate on create** — `ProjectServiceImpl.createProject` allocates a default `~/Documents/Wayland/<name>` (collision-safe) when the user picks no folder, and persists it. *(scope 1)*
2. **Folder picker** — already present in `CreateProjectModal`/`ProjectSettingsDrawer`; lights up automatically. *(scope 2 — no change needed)*
3. **Resolution** — project chats resolve to the project's persistent workspace via the existing `enforceProjectWorkspace`, now that one always exists. *(scope 3)*
4. **Skills without pollution** — project workspaces now get native skill symlinks (previously skipped as "custom") via a shared `setupWorkspaceSkills` gate, plus a **managed, idempotent `.gitignore`** so the skill dot-dirs stay out of the user's git. *(scope 4 — see engine note below)*
5. **Lazy migration** — `ensureProjectWorkspace` allocates+persists+bootstraps for existing empty-workspace projects on the next chat, no data loss. *(scope 5)*
6. **Real-path panel** — already keyed off the `-temp-` regex; shows the real path now that workspaces are real. *(scope 6 — no change needed)*

### ⚠️ Engine note on scope 4
The issue asked for skill symlinks in a `.wayland/` **subdir**. Verified the wcore engine spawns with `cwd: workspace` and scans `<cwd>/.wayland-core/skills/` **hard-coded relative to cwd** (no env/arg override; Rust source not vendored), and every ACP CLI likewise scans its own dot-dir at cwd. A `.wayland/` subdir is therefore **infeasible without a Core engine change** (a `WCORE_SKILLS_DIR` redirect). This PR keeps symlinks at the engine-required dot-dirs and uses a managed `.gitignore` to honor the intent ("folder stays clean and skills resolve"). **A Core follow-up for true `.wayland/` consolidation is the right home for the subdir design.**

## Tests
- Unit: `workspaceLocation` (sanitize + collision-safe path, incl. path-escape + Windows trailing-dot), `workspaceGitignore`, `ensureProjectWorkspace` (incl. concurrency lock), `ProjectServiceImpl` allocation.
- Real-fs integration: `projectWorkspaceFs.integration.test.ts` proves a real dir on disk, collision-safe, relaunch-stable.
- E2E (turn-free): `project-persistent-workspace.e2e.ts` drives the running app — `project.create` → real `~/Documents/Wayland/<name>`, project chat pinned to it.
- `prek` (full CI) green: TypeScript / Oxlint / Oxfmt / EOF / whitespace.

## Live-verified in the running app (mac, electromcp)
- `project.create` → `workspace: /Users/…/Documents/Wayland/<name>` (real, not temp).
- Dir exists on disk; `.wayland/` knowledge bootstrapped (CONTEXT/rules/decisions).
- wcore project chat `extra.workspace` === the project workspace (no temp drift).
- `.wayland-core/skills/` symlinks present + managed `.gitignore` written.
- **Still owed (routes to Windows box):** the full agent-turn write — "ask the agent to write a file to the local workspace → appears in the dir + Files panel" — needs a live wcore turn the mac harness can't run.

## Gate
Independent adversarial cross-audit run pre-PR; it found 3 real issues (gitignore unguarded write = blocker, alloc TOCTOU race = major, name-cap trailing dot = minor) — **all fixed** in `bc0ea97e5` with added tests. Per the merge gate this still needs a **second independent audit (non-author)** + the **Windows agent-turn verify** before merge. Not self-merging.

### DIFF
--- a/src/process/services/ConversationServiceImpl.ts
+++ b/src/process/services/ConversationServiceImpl.ts
@@ -11,7 +11,7 @@ import { uuid } from '@/common/utils';
 import { cronService } from './cron/cronServiceSingleton';
 import { SqliteProjectRepository } from '@process/services/database/SqliteProjectRepository';
 import { loadProjectKnowledgeBlock, loadGlobalMemoryBlock } from '@process/services/projectKnowledge/knowledge';
-import { enforceProjectWorkspace } from '@process/services/projectWorkspace';
+import { enforceProjectWorkspace, ensureProjectWorkspace } from '@process/services/projectWorkspace';
 import {
   createGeminiAgent,
   createAcpAgent,
@@ -180,7 +180,16 @@ export class ConversationServiceImpl implements IConversationService {
    * overwritten; if the project has no workspace the temp fallback still applies.
    */
   private async reconcileProjectWorkspace(params: CreateConversationParams): Promise<void> {
-    await enforceProjectWorkspace(params.extra as Record<string, unknown> | undefined);
+    const extra = params.extra as Record<string, unknown> | undefined;
+    // #455: lazy-migrate a project that has no persistent workspace yet (created
+    // before #455, or allocation deferred) BEFORE enforcement pins the chat to
+    // it. A user-chosen custom workspace is left alone. Then enforce copies the
+    // (now guaranteed) project workspace onto this conversation's extra, so the
+    // chat lands in the persistent dir instead of a throwaway temp dir.
+    if (extra?.projectId && !extra.customWorkspace) {
+      await ensureProjectWorkspace(extra.projectId as string);
+    }
+    await enforceProjectWorkspace(extra);
   }
 
   async createConversation(params: CreateConversationParams): Promise<TChatConversation> {
@@ -213,7 +222,8 @@ export class ConversationServiceImpl implements IConversationService {
           params.extra.sessionMode,
           params.extra.isHealthCheck,
           params.extra.extraSkillPaths as string[] | undefined,
-          params.extra.excludeBuiltinSkills as string[] | undefined
+          params.extra.excludeBuiltinSkills as string[] | undefined,
+          params.extra.projectId as string | undefined
         );
         break;
       }
--- a/src/process/services/ProjectServiceImpl.ts
+++ b/src/process/services/ProjectServiceImpl.ts
@@ -11,6 +11,7 @@ import type { IProject, ICreateProjectParams, IUpdateProjectParams } from '@/com
 import type { TChatConversation } from '@/common/config/storage';
 import { uuid } from '@/common/utils';
 import { bootstrapProjectKnowledge } from '@process/services/projectKnowledge/bootstrap';
+import { allocateProjectWorkspace } from '@process/services/projectWorkspace';
 
 /**
  * Concrete IProjectService. Owns id/timestamp generation and the `.wayland/`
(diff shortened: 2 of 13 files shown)


FerroxLabs:
Overwatch: the windows-2022 4/4 shard was failing on this branch because tests/unit/workspaceLocation.test.ts asserted resolveProjectWorkspacePath output against forward-slash template literals, but the function builds paths with path.join (native separators, correct for real dirs). Four assertions failed as backslash-vs-slash on Windows.

Pushed f66dfc738: assert with path.join on both sides so expectations match per platform. Test-only, source path logic untouched. 12/12 pass locally. Your 10:13 collision-race hardening is in and good. Once this CI run goes green and you post mac persistent-workspace live-verify evidence, this is clear to merge.

Issue #348, opened 2026-06-26
MCP tool scoping: per-server allowed_tools toggles (candidate pool for selection)


FerroxLabs:
Part of the MCP tool-scaling + management build (spec: docs/mcp-tool-scaling-spec.md, Lane 2). Sibling lanes: #344 (Lane 3, selection/cap) + Lane 1 (management overview).

## Scope (per-server tool scoping; feeds Lane 3)
Let users enable/disable individual tools per MCP server (Hermes-style allowed_tools). This shrinks the candidate pool before Lane 3 selects/caps.

- Persist allowed_tools per server: Record<serverId, string[] | "all"> (default "all").
- Expose each server's tool list (name + description) via existing McpProtocol.listTools / mcp_ready.
- DetailPage UI: per-tool on/off toggles + enable-all/disable-all.
- Expose getCandidateTools() per the spec contract. MUST NOT write configBridge.allow_list (Lane 3 owns that writer).

Files: src/process/services/mcpServices/ (config persistence + getCandidateTools), src/renderer/pages/settings/McpLibrary/DetailPage.tsx.
Acceptance: toggling a tool off removes it from the candidate set and persists across restart.
Branch off ferrox/integration/0.11.4; PR -> integration/0.11.4.

FerroxLabs:
**@Overwatch — ruling needed** (from desktop): Lane 2 is contract-blocked. Verified: (1) the spec docs/mcp-tool-scaling-spec.md does NOT exist in repo, any worktree, git history, or any ferrox/origin/upstream branch; (2) Lane 3 (#344) has NOT landed its contract stub — no branch, no open PR. Per the agreed sequencing (#344 comment: 'Lane 3 OWNS the contract types — land the contract stub first so Lane 2 builds against it'), Lane 2 has nothing to build the boundary against. Specifically I need from Lane 3/spec: the getCandidateTools() return type (CandidateTool shape) + the allowed_tools persistence shape (Record<serverId,string[]|'all'> vs a per-server IMcpServer field). I can build the contract-free parts now (expose each server's tool list via listTools/mcp_ready — tools[] already persists; DetailPage per-tool toggle UI scaffold). REQUEST: either have Lane 3 land the contract stub + commit the spec, or authorize me to author a provisional contract stub for Lane 3 to adopt. Not blocking — proceeding on contract-free parts meanwhile.

FerroxLabs:
[...]

Issue #56, opened 2026-06-10
[Question]: Wayland Desktop v0.9.6-rc.2.1, custom MCP Server + Codex


Kula-Loki:
### Question

In Wayland Desktop v0.9.6-rc.2.1, my custom MCP server shows Running with 5 tools in MCP Library - Installed, but Codex chat sessions do not receive the tools. Are custom MCP tools currently injected into Codex sessions, regular Wayland chats, or only available in another surface? What setting or session type enables custom MCP tools for chat?

### Question Category

Configuration

### Context

I have installed my custom MCP with 5 tools running.  Tried using with Wayland (Codex) and not working.

### Platform

macOS (Intel)

FerroxLabs:
Confirmed gap: custom MCP servers are injected for Gemini, but the Codex/Claude ACP session path filters to builtin-only, so custom tools do not reach those agents. Accepted onto the backlog.

FerroxLabs:
Fixed for v0.10.0: custom MCP servers are now injected into Codex **and** Claude ACP (CLI-backed) sessions, not just the in-app backends. The CLIs reject dots in server names, so names are sanitized on injection.

Shipping in **v0.10.0** (building now) — I'll confirm here when the release publishes.

FerroxLabs:
Released in **v0.10.0** (now Latest): https://github.com/FerroxLabs/wayland/releases/tag/v0.10.0 — it auto-updates. Closing as completed; please reopen if it persists on the new build.

Issue #476, opened 2026-06-30
MCP connectors only reach Claude Code backend — invisible to Codex + Wayland Core [lane: desktop]


FerroxLabs:
Live-tested: same prompt (check email/calendar), same configured Google Workspace MCP connector. Claude Code SAW + USED it (full calendar + email triage). Codex reported 'only ijfw-memory configured'. Wayland Core reported 'no tools available'. So a configured MCP connector surfaces in ONE backend and is invisible in the other two — affects EVERY connector, not just Google Workspace.

Root cause (per-backend MCP config source):
- Claude Code (ACP): Wayland injects the full server set into the session via acp/index.ts:1646 newSession(workspace, {mcpServers}) from loadBuiltinSessionMcpServers. -> gets connectors.
- Codex (CodeBuddy): reads its OWN ~/.codebuddy/mcp.json (acpConnectors.ts:529) — Wayland isn't populating it with the desktop's enabled connectors.
- Wayland Core (engine): reads config.toml [mcp.servers] (WCoreMcpAgent) — same gap (and/or the stdio server crashed on spawn, see #475).

Fix: propagate the desktop's enabled MCP connectors to ALL backends' config sources, not only the Claude ACP session — write them into Codex's mcp.json and wcore's config.toml so tool visibility is uniform. DEMO-BLOCKER for any non-Claude-Code backend.

Recommendation for the imminent demo: feature Claude Code (the working backend); fix Codex/wcore propagation as fast-follow.

FerroxLabs:
Verified shipped in **Desktop v0.12.0**.

**Evidence:** PR #478 merged, dc52d184d — behind, ahead 0, behind 190

**Detail:** "fix(mcp): enabled connectors reach every agent backend (Codex, wcore, Gemini)" — the exact three-backend gap reported. ⚠️ Already in v0.11.18 (behind 0 there) — this has been shipped and left open for a while.

Ancestry was proven against the released v0.12.0 tag using the GitHub compare API (a commit is in the release only when status is behind or identical AND ahead_by is 0), and corroborated by reading the code at the tag rather than trusting the timeline alone.

Reopen if the symptom recurs on 0.12.0 or later.

Issue #475, opened 2026-06-30
Google Workspace MCP: agent can't drive OAuth — 122-tool flood + start_google_auth(service_name) loop chokes the turn [lane: desktop]


FerroxLabs:
Live-tested with real creds on main (+ PR #474 which fixes the server-start crash). The server now connects, but the AGENT cannot complete the auth flow, and the attempt chokes the turn.

Root causes (confirmed live):
1. **Tool overflow:** workspace-mcp 1.22.0 advertises **122 tools**. That floods the model's context and degrades tool selection. This is the MCP tool-scaling/curation problem (see #344 / wayland-core#86) — a single connector should not dump 122 tools unfiltered.
2. **start_google_auth requires service_name:** schema = required:[service_name], optional:[user_google_email]. The model repeatedly calls it without a valid service_name -> validation failure -> re-think -> retry loop.
3. **Budget exhaustion:** the retry loop + 122-tool context burns the model's output/reasoning budget mid-thought ('used up its room while thinking') so it never answers — even after Continue-with-more-headroom.
4. **OAuth consent:** start_google_auth is meant to return a consent URL; the in-chat flow never reaches a browser consent.

Fix directions:
- Curate the connector's tool surface (model-aware BM25 / allow-list) so it doesn't dump 122 tools.
- Add explicit usage guidance for start_google_auth (service_name='gmail'|'drive'|..., supply user_google_email) via the connector setup guide / tool description.
- Make the consent hand-off actually open the browser and complete.
- Guardrail: cap failing-tool retry loops so a bad tool call can't exhaust the whole budget.

Status: server-start fixed (#474). This issue tracks the agent-driven auth UX. NOT demo-ready until curation + auth guidance land.

FerroxLabs:
**@Overwatch — ruling needed** (from core): CORE lane — need a ruling on the retry-cap slice (engine guardrail; the OAuth/122-tool/consent parts stay desktop).

Built the core guardrail: FailureGuard, a per-run breaker that counts CONSECUTIVE same-tool error results (args ignored), stopping the run with guidance when a tool keeps failing — the varied-args validation-error loop the existing signature-based LoopGuard cannot see. Gated + cross-audited. Two entangled findings block a clean land, both need a call:

[...]

PR #478, opened 2026-06-30
fix(mcp): enabled connectors reach every agent backend (Codex, wcore, Gemini) [lane: BH1]

FerroxLabs:
## Problem (live-confirmed)

An enabled MCP connector (e.g. the Google Workspace stdio `uvx` server) was **usable in Claude** but **invisible in Codex and Wayland Core**. Affects every connector.

## Verified diagnosis (overturns the prior, stale-branch diagnosis)

I verified against the running app's real on-disk state, not the stale branch:

- **The Codex stdio capability bit is NOT the cause.** Codex's cached ACP `initialize` advertises `mcpCapabilities {stdio:true, http:true, sse:false}` (read from the real `acp.cachedInitializeResult`). The diagnosis's "Codex omits mcpCapabilities → stdio dropped" is false for current Codex.
- **The cited code path is dead.** `loadBuiltinSessionMcpServers` in `src/process/agent/acp/index.ts` is never instantiated. The live ACP path is `AcpAgentV2`/`AcpRuntime` → `McpConfig.fromStorageConfig`, which already accepts `status: undefined`.
- **Real divergence #1 (predicate):** Gemini used the stricter `shouldInjectSessionMcpServer` (required `status === 'connected'`), so it dropped `status: undefined` connectors that Claude/Codex kept.
- **Real divergence #2 (wcore injection):** `WCoreManager` injected only team / team-guide MCPs — never `mcp.config` user connectors.
- **Real divergence #3 (wcore path):** `WCoreMcpAgent` wrote `[mcp.servers]` to `<binary> --config-path` (native default-profile dir, since the main process has no `WAYLAND_HOME`) while the engine spawn reads the **active-profile** config via `WAYLAND_HOME`. Verified live: bundled engine `--config-path` returns `~/Library/Application Support/wayland-core/config.toml` by default and `~/.wayland/profiles/<name>/config.toml` under `WAYLAND_HOME`.
- **Real divergence #4 (Codex native config):** Codex reads MCP only from `~/.codex/config.toml` at startup; the sync fired only on a UI toggle. Verified live: `~/.claude.json` had the connector; `~/.codex/config.toml` did not.

## Fix

1. **Predicate (Gemini + shared):** `shouldInjectSessionMcpServer` now accepts an enabled connector on `status === undefined || 'connected'` (only excludes a known-broken `disconnected`/`error`), matching the live ACP path so Gemini and wcore stay in lockstep with Claude/Codex.
2. **wcore injection:** `WCoreManager` injects enabled user **stdio** connectors at spawn via the proven `add_mcp_server` path (the same `stdioMcpServers` array the team MCPs already use), using the shared predicate + #348 per-conversation scoping. Builtins (image-gen, skill-search, concierge-diag) and hosted http/sse connectors are left to their existing mechanisms.
3. **wcore config path:** `WCoreMcpAgent` reads/writes `resolveActiveConfigPath()` (the single source of truth the engine spawn also resolves through), with a `--config-path` fallback — so settings-sync writes where the engine reads on named profiles.
4. **Codex native config:** `AcpAgentManager` re-syncs enabled connectors into `~/.codex/config.toml` at codex bootstrap (reusing `mcpService.syncMcpToAgents`), de-duped per process by a connector signature, and runs **before** `materializeFluxCodexHome` so the flux scoped `CODEX_HOME` inherits them.

Keeps #348 scoping, #56 lockstep, and the builtins intact; Claude is unchanged.

## Verification

- **Unit:** new `tests/unit/process/agent/acp/wcoreUserMcpInjection.test.ts` (relaxed predicate + wcore injector: stdio-only, builtins excluded, hosted excluded, disabled excluded, #348 scoping); updated `geminiMcpInjection` and `acpBuiltinMcp` to the uniform contract. Full mcp/wcore/codex unit surface green (**2797 passing, 0 failing**). `tsc --noEmit` clean.
- **Live — Claude:** unchanged (already worked; connector present in `~/.claude.json`).
- **Live — Codex mechanism:** `codex mcp add` (the exact call `CodexMcpAgent` makes) into a scoped `CODEX_HOME` wrote `[mcp_servers.<name>]` and `codex mcp list` showed it `enabled` — proving the bootstrap ensure lands the connector in codex's native config.
- **Live — wcore config-path alignment:** bundled engine `--config-path` confirmed to return the native dir by default and the profile dir under `WAYLAND_HOME=<profile>`; `resolveActiveConfigPath()` returns those same paths, so sync now writes exactly where the engine reads. wcore runtime injection uses the already-in-production `add_mcp_server` path (unchanged consumer; only the inputs are added).
- **Live — Gemini:** covered by the predicate change + unit tests (was the strict-predicate outlier).

### Not yet done
A full interactive 4-backend chat-turn in the dev app (sending a tool call through each running backend) was not performed in this session — that needs the packaged/dev app with live provider creds. The mechanisms are verified per-backend above (config writes + the proven injection path + unit tests).

### DIFF
--- a/src/process/agent/acp/mcpSessionConfig.ts
+++ b/src/process/agent/acp/mcpSessionConfig.ts
@@ -45,10 +45,18 @@ function toNameValueEntries(source?: Record<string, string>): AcpSessionMcpNameV
  *
  * Builtin servers (image generation, skill search) are seeded into mcp.config
  * with `status: undefined` and are never connection-tested, so they must be
- * accepted on `undefined`; otherwise a backend silently drops them. User-added
- * (non-builtin) servers still require an active `connected` status.
+ * accepted on `undefined`; otherwise a backend silently drops them.
  *
- * Both backends must agree: previously the ACP path injected builtin servers
+ * User-added (non-builtin) servers are accepted on `undefined` OR `connected`,
+ * and only excluded on an explicit failure status (`disconnected`/`error`). An
+ * enabled connector the user has not yet connection-probed (`status: undefined`)
+ * must still reach the session - the live ACP path (McpConfig.fromStorageConfig,
+ * used by AcpAgentV2/AcpRuntime for Claude/Codex) already accepts `undefined`,
+ * so requiring `connected` here meant Gemini (and the wcore injector that shares
+ * this predicate) silently dropped connectors that Claude/Codex kept - the exact
+ * cross-backend divergence this predicate exists to prevent.
+ *
+ * Every backend must agree: previously the ACP path injected builtin servers
  * only, so a user's custom MCP server reached Gemini chats but never Codex or
  * Claude chats (GitHub #56). Using one predicate keeps them in lockstep.
  */
@@ -56,10 +64,9 @@ export function shouldInjectSessionMcpServer(server: IMcpServer): boolean {
   if (!server.enabled) {
     return false;
   }
-  if (server.builtin === true) {
-    return server.status === undefined || server.status === 'connected';
-  }
-  return server.status === 'connected';
+  // Both builtin and user servers: accept not-yet-probed (undefined) or
+  // connected; a known-broken (disconnected/error) server is not surfaced.
+  return server.status === undefined || server.status === 'connected';
 }
 
 /**
@@ -131,6 +138,43 @@ export function buildAcpSessionMcpServers(
       })
       .filter((server): server is AcpSessionMcpServer => server !== null)
   );
+}
+
+/**
+ * Build the stdio MCP servers for a Wayland Core spawn from the user's
+ * `mcp.config` connector list. wcore receives MCP servers at spawn via the
+ * engine's `add_mcp_server` runtime command (stdio only), so this mirrors the
+ * ACP session-injection path: same `shouldInjectSessionMcpServer` predicate and
+ * `isServerActiveForSession` per-conversation scoping (#348). Builtins
+ * (image-gen, skill-search, concierge-diag) are excluded - wcore surfaces those
+ * through its own mechanisms (system-prompt skills index / team guide), and
+ * injecting them here would double them up or leak the Concierge-only diag
+ * server. Non-stdio (hosted http/sse) connectors are skipped because the engine
+ * runtime command only adds stdio servers; those still reach wcore via the
+ * [mcp.servers] table written by WCoreMcpAgent.
+ */
+export function buildWCoreUserStdioMcpServers(
+  mcpServers: IMcpServer[] | undefined | null,
+  activeServerIds?: readonly string[]
+): AcpSessionMcpServerStdio[] {
+  if (!Array.isArray(mcpServers) || mcpServers.length === 0) {
+    return [];
+  }
+  return mcpServers
+    .filter(shouldInjectSessionMcpServer)
+    .filter((server) => server.builtin !== true)
+    .filter((server) => isServerActiveForSession(server, activeServerIds))
+    .filter((server) => server.transport.type === 'stdio')
+    .map((server): AcpSessionMcpServerStdio => {
+      const transport = server.transport as Extract<IMcpServer['transport'], { type: 'stdio' }>;
+      return {
+        type: 'stdio',
+        name: server.name,
+        command: transport.command,
+        args: transport.args || [],
+        env: toNameValueEntries(transport.env) ?? [],
+      };
+    });
 }
 
 /** Config shape passed from TeamSessionService to AgentManagers */
--- a/src/process/agent/acp/mcpSessionConfig.ts
+++ b/src/process/agent/acp/mcpSessionConfig.ts
@@ -7,6 +7,7 @@
 import type { IMcpServer } from '@/common/config/storage';
 import type { AcpMcpCapabilities } from '@/common/types/acpTypes';
 import { BUILTIN_CONCIERGE_DIAG_ID } from '@process/resources/builtinMcp/constants';
+import { sanitizeMcpServerName } from '@process/services/mcpServices/validateMcpServer';
 
 export interface AcpSessionMcpNameValue {
   name: string;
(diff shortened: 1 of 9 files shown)


FerroxLabs:
[Overwatch dispatch -- lane: BH1] Do AFTER #431. BLOCKED on PR #459 merge (shared src/process/task/WCoreManager.ts).
PR #478 MCP-all-backends (#476). Two owed items still open:
1. Double-registration: WCoreMcpAgent writes stdio connectors to config.toml [mcp.servers] AND the new path add_mcp_server's the same connector at spawn -> registered twice, compounding the #475 tool-flood. Add a dedup guard OR verify engine add_mcp_server idempotency (see Overwatch core note; coordinate with core).
2. Live 4-backend chat-turn verify (Claude Code, Codex, wcore, Gemini): an enabled connector must actually appear and work in each. Not done yet.
After #459 merges: rebase onto it, then RE-RUN #459's wcore-true-continue.e2e.ts (a bad WCoreManager.ts conflict resolution can silently drop #459's resume-seed or your MCP inject).
GATE: cross-audit + double-reg confirmed absent + 4-backend live PASS. wl report 478; Overwatch merges.

FerroxLabs:
[Overwatch] UNBLOCKED — PR #459 is MERGED to main. Proceed: (1) cd /private/tmp/wt-mcp-uniform; git fetch ferrox && git rebase ferrox/main. (2) Resolve the WCoreManager.ts conflict carefully — #459 rewrote the resume-seed block; do NOT drop its buildResumeSeedTranscript seeding or your MCP-inject path. (3) MANDATORY after rebase: re-run #459's wcore-true-continue.e2e.ts to prove the merge didn't silently drop either fix. (4) Then finish the double-registration dedup guard + the live 4-backend verify. Gate: cross-audit + double-reg confirmed absent + 4-backend live PASS. wl report when ready; I merge.

FerroxLabs:
## [lane: BH1] #478 update — rebased over #459, double-registration dedup guard added, cross-audited

Picked up after #459 merged. Summary + gate status.

### Rebase over #459 (True Continue) — both fixes kept
Rebased onto `ferrox/main` (12 behind). The `WCoreManager.ts` overlap auto-merged; **verified both survive** (6 markers, 0 conflict markers): #459's `buildResumeSeedTranscript` resume-seed + `max_turns` truncation AND this PR's MCP-inject path. Full typecheck clean.

**Re-ran `wcore-true-continue.e2e.ts`:** the in-app engine leg can't run in this harness (electronApp fixture setup times out at 60s — the documented local limit; not a code fault, typecheck is clean). Proved the merge dropped neither fix **deterministically instead**: the 56-test True-Continue + MCP-inject unit suites all pass (`resumeSeedTranscript`, `WCoreManagerTurnCompletion`, `finishReasonLength`, `messageTextContinueBanner`, plus the injection suites). The live engine leg is routed to Overwatch.

### Double-registration dedup guard (#478)
Root cause (created by this PR's own path fix): once the settings-time write targets the **active-profile** config.toml (the file the engine actually reads), a settings-saved stdio connector lands in config.toml `[mcp.servers]` (engine loads at startup) **and** is injected at spawn via `add_mcp_server` → registered twice.

Fix: `readWCoreConfigMcpServerNames()` (new `configMcpServers.ts`, reusing `resolveActiveConfigPath` so it reads the exact file the engine loads) → passed as `excludeNames` to `buildWCoreUserStdioMcpServers`, which now skips those. Guarantees absence **independent of** engine `add_mcp_server` idempotency (so no Core dependency). Best-effort: an unreadable config.toml yields an empty set (inject rather than silently drop a connector — worst case a benign dup, never an invisible connector).

### Cross-audit (non-author gate) — done; 1 critical found + fixed
Independent adversarial reviewer. **Critical, real:** name-key mismatch — `McpService.syncMcpToAgents` sanitizes names (`sanitizeMcpServerName`, `com.slack/slack-mcp` → `com.slack-slack-mcp`) **before** config.toml is written, but the builder read/compared/injected the **raw** name → dedup missed + engine keyed the server twice under two names. **Fixed:** the builder now sanitizes the name with the identical transform for both the injected `add_mcp_server` name and the exclude comparison, so it matches the config.toml key exactly. Regression test added (`com.slack/slack-mcp`). Auditor also confirmed sound: raw-engine mode (guard reads the same file the engine loads), profile resolution can't diverge, empty-set fallback intentional, True Continue intact.
(I also independently verified the TOML dotted-key round-trip: `smol-toml` quotes `[mcp.servers."a.b.c"]` and parse returns the exact single key — dotted names match.)

### Verification status
- **Deterministic, green:** 37 unit tests across all four backend injection paths (wcore builder + dedup, ACP `acpBuiltinMcp`, Gemini `geminiMcpInjection`) + config-reader; prek green (TS/oxlint/oxfmt/EOF/ws); typecheck clean.
- **Live 4-backend verify (Claude Code / Codex / wcore / Gemini):** cannot run locally — no app reachable and the harness fixture can't launch one (per the e2e timeout above). **Routed to Overwatch** for the real engine-turn run: enable one stdio connector, confirm its tools appear in a live turn on each of the 4 backends, and confirm the connector is registered **once** (not twice) in wcore.

**Gate:** cross-audit ✅ (critical fixed) + double-reg confirmed absent by construction & unit test ✅ + 4-backend live PASS ⏳ (Overwatch). Overwatch merges — not self-merging.

FerroxLabs:
**[core lane] add_mcp_server idempotency contract — verified in code: NOT idempotent, do not re-send.**

Verified on current main (`wcore-cli/src/main.rs:2696` handler → `wcore-mcp/src/tool_proxy.rs:209` `register_single_server_tools` → `wcore-tools/src/registry.rs` `register`):

1. The AddMcpServer handler has **no existing-name guard**: every command spawns a **new McpManager connection** (a second child process for stdio transports) and pushes it into `dynamic_managers` — both stay alive.
2. Tool registration uses `ToolRegistry::register`, which **pushes without name-dedup** (`register_replace` exists but is not used on this path). Duplicate entries land in the registry.
3. `ToolRegistry::get` returns the **first** name match — so after a re-add, tool calls still route to the OLD manager/connection; the fresh one is dead weight. Re-adding a dead server does NOT heal it.

**Contract for desktop:** send add_mcp_server exactly once per server name per session (pre-first-Message phase only — it's rejected after the first Message). If you need re-configuration, restart the engine session. I'm filing the engine-side defect (re-add should replace: swap to `register_replace` + drop/replace the old manager) as a core follow-up.

Issue #477, opened 2026-06-30
Context meter hardcodes Claude models at 200K — Opus 4.x is 1M; wrong denominator + likely early compaction [lane: desktop] [lane: Main]


FerroxLabs:
src/renderer/utils/model/modelContextLimits.ts pins ALL Claude models to 200_000 (claude-opus-4 / 4.1 / 4.5). The ContextUsageIndicator reads this table, so the meter shows e.g. '94.7K / 200K = 47.4%' for an Opus Claude Code session. Opus 4.x's real ceiling is 1M.

Consequences:
1. Meter shows the wrong denominator (looks ~half full when there's far more room).
2. If compaction/budgeting keys off this 200K assumption, the session may COMPACT EARLY — plausibly contributing to 'choked on context' reports.

Note: user runs Claude Code via SUBSCRIPTION (not API key), so this is NOT an API beta-header flip. The effective window depends on the Claude Code plan; the table should reflect the real window (ideally read live from the session/Claude Code), and at minimum stop hard-pinning Opus 4.x to 200K. Also add the current model id (claude-opus-4-8) which likely falls through to the default.

Do NOT build Claude subscription OAuth (ToS) — Claude Code (Anthropic's CLI) using the subscription is fine; this is a display/budget-table fix only.

FerroxLabs:
[Overwatch dispatch -- lane: Main]
Task (trivial): fix the context-usage meter denominator. src/renderer/utils/model/modelContextLimits.ts lines 44-53 hardcode every Opus at 200000; claude-opus-4-8 fuzzy-matches the claude-opus-4 entry -> 200000, ~5x too small. Add opus-4.6/4.7/4.8 entries at their REAL context window. CONFIRM each id's real window against the engine/provider before writing the number -- this file is its own source of truth, do not guess. Add a unit test for the opus-4-8 fuzzy-match. Display-only (no compaction impact -- verified).
GATE: independent (non-author) cross-audit + live-verify the meter shows the correct percent on a real opus-4.8 profile before done. Then: wl report 477 "ready for merge". Overwatch merges.

FerroxLabs:
[...]

PR #510, opened 2026-07-01
fix(model): correct context-meter window for Opus 4.6+/Sonnet 4.6 (1M)

FerroxLabs:
## What

Fixes the context-usage meter denominator (issue #477). `getModelContextLimit` read a table keyed on **dotted** ids (`claude-opus-4.5`), but the app passes the **real hyphenated** catalog ids (`claude-opus-4-8`). Every Claude id therefore only ever hit the bare `claude-opus-4` / `claude-sonnet-4` fuzzy-prefix fallback — so `claude-opus-4-8` resolved to **200K**, and the meter showed `X / 200K` for a session whose real window is **1M** (and any budgeting keyed off this ceiling was wrong).

## Ground truth (important)

Values come from the app's own provider snapshot `resources/modelsdev-snapshot.json` (models.dev). This **contradicts the issue's premise** that "all Opus 4.x is 1M":

| Model | Real context |
|---|---|
| Opus 4.6 / 4.7 / 4.8, Sonnet 4.6 | **1M** |
| Opus 4.0 / 4.1 / 4.5, Sonnet 4.0 / 4.5, Haiku 4.x | **200K** |

So Opus 4.0/4.1/4.5 correctly stay at 200K. The change also fixes a **latent opposite bug**: the old `claude-sonnet-4` = 1M fallback *over*-reported Sonnet 4.5/4.0 (real 200K) as 1M.

## How

- Re-key the Claude family on real hyphenated ids with snapshot-accurate values.
- Keep bare `claude-opus-4` / `claude-sonnet-4` / `claude-haiku-4` = 200K fallbacks so dated/variant ids (`claude-opus-4-20250514`) still resolve; versioned keys win via longest-match.
- Add previously-missing `claude-3-sonnet` = 200K (was falling through to the 1M default).

**Display-only.** No behavior change beyond the meter denominator; no user-facing strings (no i18n impact).

## Tests / verification

- New `tests/unit/modelContextLimits.test.ts` — 11 tests against the real `getModelContextLimit` source: exact + dated-variant + case-insensitive resolution, both 1M and 200K arms, a regression guard for the Sonnet over-report, and null/empty fallbacks.
- `bun run typecheck` green; `prek run --from-ref origin/main --to-ref HEAD` green (TypeScript / Oxlint / Oxfmt / UI-tokens).
- Independent adversarial cross-audit: traced all 24 catalog Claude ids through the matcher against the snapshot — no regressions, no collisions, callers unaffected (both pass hyphenated `useModel`).

**Not yet done — routed to Overwatch:** rendered-meter live-verify on CDP :9333 needs a branch build + a real Opus session with token usage, which the local harness can't produce (real wcore turns fail locally). Denominator logic itself is verified as above.

Refs #477

### DIFF
--- a/src/renderer/utils/model/modelContextLimits.ts
+++ b/src/renderer/utils/model/modelContextLimits.ts
@@ -40,16 +40,29 @@ const MODEL_CONTEXT_LIMITS: Record<string, number> = {
   o3: 200_000,
   'o3-mini': 200_000,
 
-  // Claude family
-  'claude-opus-4.5': 200_000,
-  'claude-haiku-4.5': 200_000,
-  'claude-sonnet-4.5': 1_000_000,
-  'claude-opus-4.1': 200_000,
+  // Claude family. Keys use the real (hyphenated) catalog model ids the app
+  // passes here, and values follow the models.dev provider snapshot
+  // (resources/modelsdev-snapshot.json): only Opus 4.6+ and Sonnet 4.6 ship a
+  // 1M window; Opus 4.0/4.1/4.5, Sonnet 4.0/4.5, and Haiku 4.x are 200K. The
+  // bare `claude-opus-4` / `claude-sonnet-4` / `claude-haiku-4` entries are the
+  // fuzzy fallback for dated or variant ids (e.g. `claude-opus-4-20250514`);
+  // longest-match means the versioned keys above win for known 1M models.
+  'claude-opus-4-8': 1_000_000,
+  'claude-opus-4-7': 1_000_000,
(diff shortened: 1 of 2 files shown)


FerroxLabs:
[Overwatch] APPROVED (#477). Cross-audit clean; good catch correcting the premise (only Opus 4.6/4.7/4.8 + Sonnet 4.6 are 1M) and the latent Sonnet over-report, values from the repo snapshot. Merging on green CI. Meter live-verify (display-only) is routed to Overwatch.

PR #511, opened 2026-07-01
fix(team): budget mailbox prompt content

sazora:
## Summary
- cap individual team mailbox messages before they are inserted into teammate wake prompts
- cap the total unread mailbox bundle included in a single prompt
- prefer a message summary when available so long teammate reports do not overflow GPT-5.5 context

## Why
Team chats can forward large `team_send_message` outputs into another agent's unread mailbox. Those unread messages are later replayed into the next wake prompt. A single huge teammate report, or several medium reports, can push the request above the model context ceiling and stop the run even after compaction.

## Tests
- `bunx vitest run tests/unit/process/team/formatHelpers.test.ts tests/unit/team-TeamMcpServer.test.ts tests/integration/team-mcp-server.test.ts`
- `bun run typecheck`
- `bun run format:check src/process/team/prompts/formatHelpers.ts tests/unit/process/team/formatHelpers.test.ts`

### DIFF
--- a/src/process/team/prompts/formatHelpers.ts
+++ b/src/process/team/prompts/formatHelpers.ts
@@ -1,14 +1,66 @@
 import type { TeamAgent, MailboxMessage } from '../types';
 
+const DEFAULT_MESSAGE_CHAR_LIMIT = 6000;
+const DEFAULT_TOTAL_CHAR_LIMIT = 30000;
+
+export type FormatMessagesOptions = {
+  messageCharLimit?: number;
+  totalCharLimit?: number;
+};
+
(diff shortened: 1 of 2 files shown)


FerroxLabs:
[Overwatch] Cross-audit: GO, nice one @sazora. Genuine, well-scoped hardening of a real shipped path (the teammate wake-prompt via formatMessages), contained blast radius, tests + CI green, security clean. One item to resolve before merge (we don't ship dead surface): the new FormatMessagesOptions param is never passed by either TeammateManager caller — either wire it to a real config value, or drop the param and keep the hardcoded caps. Push that and I'll merge.

Issue #486, opened 2026-06-30
[Bug]: Tool step spinner remains active after completed tool calls [lane: Main]


flowsionAI:
### Platform

macOS (Apple Silicon)

### Wayland Version

0.11.8

### Bug Description

The tool step panel can remain stuck showing active spinners after the assistant’s tool calls have already completed.

In my case, the assistant ran two tool calls:
1. Reading trader_os_ignition_phase_1_5_build_handoff.md
2. Skill view (project foundation routing)

The assistant had already received completed results and responded normally, but the UI still showed both steps as if they were running.

### Steps to Reproduce

1. Open a project chat in Wayland.
2. Ask the assistant to read a local file and inspect relevant project context.
3. The assistant runs two tool calls in parallel:
   - read_file on a local Markdown handoff file
   - skill_view for project-foundation-routing
4. Wait until the assistant responds with the result.
5. Look at the step panel above the response.

Observed in this chat while reading:

/Users/willyjhongalindez/.wayland-config/temp/trader_os_ignition_phase_1_5_build_handoff.md

The UI continued showing:
- “Reading trader_os_ignition_phase_1_5_build_hand…”
- “Skill view (project foundation routing)”

with spinner indicators, even though both tool calls had completed.

### Expected Behavior

After tool calls complete, the step panel should update to a completed state, stop showing active spinners, or collapse/clear the active running indicator.

If the assistant has already received tool results and produced a final response, the UI should not continue to imply that the tools are still running.

### Actual Behavior

The assistant confirmed that both tool calls had completed, but the Wayland step panel still showed the two steps with active spinner indicators.

This made it look like the assistant was still reading/thinking even though execution had already finished.

### Additional Context

Screenshot attached. 

<img width="554" height="166" alt="Image" src="https://github.com/user-attachments/assets/42995e3d-dcea-42e3-93c9-5f506554598d" />

Environment:
- Wayland: 0.11.8
- macOS: 14.8.7
- Architecture: arm64 / Apple Silicon
- Context: Project chat
- Providers configured: ChatGPT Subscription
- Model used in session: gpt-5.5
- Feature involved: tool step/status panel

Possible cause:
The step/status UI may not be reconciling completed tool-call state correctly, possibly after parallel tool calls.

FerroxLabs:
[Overwatch dispatch -- lane: Main] SPIKE FIRST -- do not write code yet.
[...]

Issue #133, opened 2026-06-16
[Bug]: Cannot import Obsidian, Claude memory, Watch drop folder or Scan~/dev


sflorin-git:
### Platform

Windows

### Wayland Version

v0.9.9

### Bug Description

I'm trying to add my Obsidian vault to Memory Archive. Clicking the Import Obsidian vault screen just flickers and a text flashes for a sub-second, with "No matches. Try removing a filter or clearing the search". Same issue clicking Import claude-mem, Watch drop folder or Scan ~/dev

### Steps to Reproduce

Go to Memory - Archive
Click one of the buttons Import Obsidian, Import Claude memory, Watch drop folder or Scan~/dev

### Expected Behavior

Open browsert to point to memory folder.

### Actual Behavior

Screen just flickers and a text flashes for a sub-second, with "No matches. Try removing a filter or clearing the search". Same issue clicking Import claude-mem, Watch drop folder or Scan ~/dev

FerroxLabs:
Looked at all four sources. The drop-folder watcher, ~/dev scan, and Claude-memory importers are wired correctly — they surface "No matches" when the source is empty or absent (no ~/.claude-mem DB, empty drop folder, no IJFW projects under ~/dev), which reads like a hard failure. We are improving those to show the actual reason. Obsidian vault auto-detection is not implemented yet (you can paste a vault path for now). Note: the Claude-memory importer targets the claude-mem plugin DB, not Claude Code native ~/.claude/projects — that distinct source is tracked in #165.

sflorin-git:
Ok, I understand. Where do I paste my Obsidian vault path?

FerroxLabs:
#165 looks like a specific case of this (Claude memory import path). Tracking them together — the fix likely shares a root in import-path resolution.

FerroxLabs:
Reproduced and fixed. The import buttons never actually opened a folder picker — they ran an import against a **hardcoded path** (Obsidian → `~/Documents`, dev-scan → `~/dev`, claude-mem → `~/.claude-mem`). On a machine without data at those exact paths, the import found nothing, the list refreshed, and you saw the empty-archive "No matches" flash — so it looked dead.

Now **Import Obsidian opens a real folder picker** so you can point at your actual vault, and **every button shows a result** (entries imported, or a clear failure) instead of silently flashing. Fix is in #209; I'll note the release it ships in.

FerroxLabs:
[...]

PR #513, opened 2026-07-01
fix(wcore): terminalize dangling tool cards on stream_end (#486)

FerroxLabs:
## What

Secondary/defense-in-depth desktop fix for #486 (stuck tool-step spinners), per the Overwatch renderer-vs-core ruling. **Primary root cause is core** — the engine can end a turn without delivering a matching-callId `tool_result` on the ChatGPT-subscription parallel/builtin path (filed as **wayland-core#133**). This PR does *not* depend on that fix; it makes the desktop resilient regardless.

## Why

The renderer merges tool cards strictly by `callId`, so a card only leaves `Executing`/`Confirming` when a completion frame with the matching `callId` arrives. When the engine drops that frame, the card spins forever — after the assistant has already responded. The `finish`/stream_end handler previously cleared only *global* running state, never the per-card status.

## How

`src/renderer/pages/conversation/platforms/wcore/useWCoreMessage.ts`:
- Track active tool cards by `callId` → current active status as `tool_group` frames arrive (terminal frames drop them).
- On `stream_end`, terminalize whatever is still active. **Terminal status is chosen by prior status so we never paint a false success:** `Executing → Success` (it most likely ran, just missing the result frame), `Pending`/`Confirming → Canceled` (never ran / awaited an approval the finished turn can't honor), and everything → `Error` if the turn failed.
- The synthetic update is **status-only** (`{callId, status}`): composeMessage's callId merge (`{...tool, ...newToolData}`) preserves the card's name/result/description, and a real completion frame arriving later still wins.
- Tracking is cleared on the next turn's `start` (so a prior turn that ended without a `finish` — e.g. `tool_panicked` with no trailing stream_end — can't leak a stale card into this turn's finish). Deliberately **not** cleared on `error`, since a mid-turn error can be transient and followed by more tool activity in the same turn.
- Also clears the global `hasActiveTools` flag on finish so combined running state settles.

## Tests / verification

- `tests/unit/renderer/wcoreFinishReconcile.dom.test.tsx` (7 cases) drives the **real hook** through its `responseStream` handler + real `transformMessage`: Executing→Success, error→Error, Confirming/Pending→Canceled (no false success), status-only emit, parallel (only the dangling card flips), already-completed (no re-emit), and the cross-turn stale-leak guard (start clears tracking).
- `tests/unit/wcoreToolReconcileMerge.test.ts` (2 cases) exercises the `composeMessage` merge the fix relies on: status-only update preserves name/result/description; a late real frame still wins.
- `bun run typecheck` green; `prek run --from-ref origin/main --to-ref HEAD` green (TypeScript / Oxlint / Oxfmt / UI-tokens). No i18n impact (status enum values, not new strings).
- Independent adversarial cross-audit ran and found a stale-map cross-turn leak (error/start not clearing the ref) and a Pending/Confirming→false-Success semantics issue; **both fixed** (clear-on-start; status-differentiated terminal), plus a name-clobber avoided via status-only emit, plus the two merge-guarantee tests added.
- In-app smoke: the built branch boots clean with the change (no runtime errors). End-to-end in-app repro (a real ChatGPT-subscription parallel session that triggers the core frame-drop, then observing the spinner clear on finish) needs wayland-core#133 reproduced on demand → routed to Overwatch/Windows.

#486 closes when core #133 lands + this desktop reconcile ships.

Refs #486

### DIFF
--- a/src/renderer/pages/conversation/platforms/wcore/useWCoreMessage.ts
+++ b/src/renderer/pages/conversation/platforms/wcore/useWCoreMessage.ts
@@ -59,6 +59,18 @@ export const useWCoreMessage = (
   // to decide whether the error tip is fatal (keep) or transient (clear). (#101)
   const turnEndedInErrorRef = useRef(false);
 
+  // #486: tool cards (callId -> last active status) the backend has reported as
+  // active (Executing/Confirming/Pending) but not yet terminalized with a
+  // matching-callId completion frame. The engine can end a turn without
+  // delivering that frame (root cause wayland-core#133, the ChatGPT-subscription
+  // parallel/builtin path), which strands the step-panel spinner after the
+  // assistant has already responded. On stream_end we terminalize whatever is
+  // left here as defense-in-depth, mirroring the resume-time running reconcile.
+  // The prior status decides the terminal status (an Executing tool most likely
+  // finished; a Pending/Confirming one never ran), so we never assert a false
+  // success on a tool that never executed.
+  const activeToolCallsRef = useRef<Map<string, string>>(new Map());
+
   useEffect(() => {
     onConfigChangedRef.current = onConfigChanged;
   }, [onConfigChanged]);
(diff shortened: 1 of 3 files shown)


FerroxLabs:
[Overwatch] APPROVED-IN-PRINCIPLE, HOLD for CI-green + one semantic confirm. Cross-audit (yours, non-author) is solid — 2 real defects found+fixed (cross-turn stale-map leak; Pending/Confirming false-Success). Evidence the mechanism works end-to-end: a parallel Overwatch attempt live-verified the reconcile-on-stream_end on CDP (real turn, error path -> tool group terminalized, 0 dangling spinners, composer unlocked). ONE thing to confirm before merge (non-blocking): your terminalization maps a still-Executing card to SUCCESS on a successful turn. If the engine left it Executing because the tool silently hung/failed (core#133), Success is a claim we didn't observe. Consider Executing->Canceled/Unknown (honest 'outcome never reached UI') vs Success. Flagging for a quick decision in the morning. Otherwise: merges on green CI (happy-path end-to-end stays gated on core#133; error-path proven). Not merging tonight.

Issue #504, opened 2026-07-01
[Bug]: [lane: BH2]


frakman:
### Platform

Windows

### Wayland Version

0.11.8

### Bug Description

using chat.  get an approval prompt with nothign ni it

### Steps to Reproduce

get a prompt with notthing in it...

### Expected Behavior

should have the emplate populated

### Actual Behavior

<img width="2072" height="1320" alt="Image" src="https://github.com/user-attachments/assets/e7efd7bc-4bac-4cdc-a75d-cd3b343c5dee" />

frakman:
i got it again:

<img width="2107" height="1550" alt="Image" src="https://github.com/user-attachments/assets/5b8c0089-bfea-4b13-954b-6acee462958a" />

FerroxLabs:
[Overwatch dispatch -- lane: Main] BLOCKED until PR #459 merges -- DO NOT START before then.
Reason: #504 adds an ask_user question category and edits src/process/agent/wcore/protocol.ts + src/process/agent/wcore/index.ts, which PR #459 also edits. Starting first will collide.
Task (medium): there is no ask_user render path today. Add a question ToolCategory (protocol.ts), map it in wcore/index.ts (mapConfirmationDetails, ~line 768) + GeminiAgentManager.ts (getConfirmationButtons, ~line 669), and add a renderer branch (ConversationChatConfirm.tsx, MessageToolGroup.tsx) that reads the choices array. This is the AskUserQuestion multi-choice gap.
GATE: cross-audit + live-verify a real ask_user prompt renders its choices (use the Windows repro on the issue). wl ask 504 if the engine ask_user payload shape is unclear.

FerroxLabs:
[Overwatch] UNBLOCKED — PR #459 is MERGED to main. The protocol.ts + wcore/index.ts collision risk is cleared. Rebase your worktree on ferrox/main, then start #504 (add the ask_user question ToolCategory + mapping + renderer branch reading the choices array). Do this after your #486 secondary reconcile if that's still in flight. Gate: cross-audit + live-verify a real ask_user prompt renders its choices. wl ask if the engine ask_user payload shape is unclear.

FerroxLabs:
[Overwatch] Reassigned Main -> BH2 to parallelize (BH2 freed after #459 merged; #504's files do not collide with Main's #486 finish-time reconcile). BH2 owns this now.

FerroxLabs:
[...]

PR #517, opened 2026-07-01
fix(wcore): #504 render AskUserQuestion choices + return the picked answer [lane: BH2]

FerroxLabs:
## #504 AskUserQuestion — render the choices, return the pick

Closes the empty-approval-prompt bug: when the agent calls the engine's
`AskUserQuestion` tool (a structured multiple-choice question), the desktop
showed an approval box with no question and no choices, and there was no way to
send the user's selection back — so the tool errored.

### Root cause (verified against wayland-core source, not guessed)
- The engine has **no `question` ToolCategory** (`ToolCategory` = Info|Edit|Exec|Mcp). `AskUserQuestion` is emitted as an ordinary **`info`** tool named `AskUserQuestion`, with `{ question, header?, options:[{label,description}] }` buried in `args` → it fell through to the generic `info` confirmation (empty title/description, generic Allow/Deny).
- The **answer channel already exists engine-side** (wayland-core v0.9.3, `tool_approve.answer: Option<String>`); the engine synthesizes the tool result from the chosen option's **label**, guarded on `tool_name == "AskUserQuestion"`. The desktop simply never populated it. **So this is a desktop-only fix.**

### What this does (desktop)
1. **Detect by name** (not a category the engine never sends): a new pure `questionTool.ts` lifts `question`/`header`/`options` out of args into a new `question` `confirmationDetails` variant (depth/shape-hardened parser, dedupes by label).
2. **Render the choices** as the approval options (dialog `ConversationChatConfirm`), each carrying its `answer` (the label). Inline message card + summary show the question read-only; Gemini/base managers handle the new variant for exhaustiveness.
3. **Return the pick**: `tool_approve` gains an additive `answer?`; `approveTool` / `confirm` / the confirm IPC / bridge all thread it. The engine synthesizes the result and the turn resumes.
4. **Auto-modes**: full-auto (yolo) answers with the first choice (a bare approval would make the engine's loud-defensive `execute()` error); auto-edit and the "always-allow" store never auto-answer a question.
5. Remote-channel confirm prompt now shows the question + choices (was a display regression to a generic "Please confirm").

### Tests / verification
- `questionTool` unit suite; `ConversationChatConfirm.dom` asserts choices render + the pick is sent as `answer`.
- **Independent cross-audit** (non-author) — verdict safe; answer reaches the engine as the option label, verified hop-by-hop; one finding fixed (remote display).
- **LIVE-VERIFIED** (real engine, `flux-standard`): `tests/e2e/specs/wcore-ask-user-question.e2e.ts` — the agent calls `AskUserQuestion`, the prompt renders all three choices + descriptions (not empty), picking one sends the label back and the turn resumes with no error.
- prek green (TS/Oxlint/Oxfmt/EOF/whitespace/UI-tokens); typecheck clean.

### Follow-up (out of desktop lane)
Selecting a specific choice from a **remote messaging channel** (threading `answer` through the channel confirm callback) is a channels-lane follow-up — remote AskUserQuestion answering was already unsupported pre-#504.

I am the author — I will not self-merge.

### DIFF
--- a/src/common/adapter/ipcBridge.ts
+++ b/src/common/adapter/ipcBridge.ts
@@ -181,9 +181,11 @@ export const conversation = {
   confirmation: {
     add: buildEmitter<IConfirmation<any> & { conversation_id: string }>('confirmation.add'),
     update: buildEmitter<IConfirmation<any> & { conversation_id: string }>('confirmation.update'),
-    confirm: buildProvider<IBridgeResponse, { conversation_id: string; msg_id: string; data: any; callId: string }>(
-      'confirmation.confirm'
-    ),
+    confirm: buildProvider<
+      IBridgeResponse,
+      // #504: `answer` carries an AskUserQuestion choice back to the engine.
+      { conversation_id: string; msg_id: string; data: any; callId: string; answer?: string }
+    >('confirmation.confirm'),
     list: buildProvider<IConfirmation<any>[], { conversation_id: string }>('confirmation.list'),
     remove: buildEmitter<{ conversation_id: string; id: string }>('confirmation.remove'),
   },
--- a/src/common/chat/chatLib.ts
+++ b/src/common/chat/chatLib.ts
@@ -226,6 +226,19 @@ export type IMessageToolGroup = IMessage<
             toolDisplayName: string;
             serverName: string;
           }
+        >
+      // #504: AskUserQuestion-class prompt. The engine sends these as an `info`
+      // category tool named `AskUserQuestion` (there is no `question` engine
+      // ToolCategory), with the prompt buried in args - so wcore/index.ts
+      // detects them by name and lifts `question`/`header`/`choices` out of the
+      // args here, and the renderer shows the choices as selectable answers.
+      | IMessageToolGroupConfirmationDetailsBase<
+          'question',
+          {
+            question: string;
+            header?: string;
+            choices: Array<{ label: string; description?: string }>;
+          }
         >;
   }>
 >;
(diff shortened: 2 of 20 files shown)


FerroxLabs:
[Overwatch] WINNER of the #504 dup (closed #518). This is the one we ship. Before merge: (1) add the Recommended-highlight — a badge on choices[0] in ConversationChatConfirm.tsx where option.answer != null (pairs naturally with your yolo first-pick, so the recommended choice = the auto-pick); (2) one independent non-author cross-audit (the E2E covers live-verify); (3) drop the incidental authorizeOwner reformatting in ActionExecutor (unrelated linter drive-by — keep the diff surgical). One product call flagged to Sean: your yolo behavior auto-picks choices[0] for a question — I'm recommending we keep it (prevents autopilot/headless wedge), pending his nod. Merging once (1)+(2)+(3) done + his yolo confirm.

FerroxLabs:
[Overwatch] Finalized + READY (sha 9c63d4838). Recommended-highlight added (badge on choices[0], visual-only, DOM-tested). Independent cross-audit GO, zero real findings — full answer path (click/keyboard -> ipcBridge -> conversationBridge -> WCoreManager.confirm -> tool_approve.answer) verified intact, no regression to allow/deny/edit/info/mcp confirms. Gates green (10/10 unit+DOM, e2e skip-gated). Note: the ActionExecutor single-line formatting is oxfmt-mandated (reverting fails Code Quality) — accepted as-is, not a drive-by. MERGE-READY; holding ONLY on Sean's product call re yolo behavior (auto-pick recommended choice vs always-prompt) — Overwatch recommends keeping the auto-pick to avoid a headless/autopilot wedge.

PR #518, opened 2026-07-01
fix(wcore): render AskUserQuestion choices instead of a blank prompt (#504)

### LINKED ISSUE #504: [Bug]: [lane: BH2]

FerroxLabs:
## Problem (#504)

The AskUserQuestion approval prompt rendered **empty** (Windows repro on the issue). Root cause: the wcore engine tags AskUserQuestion as the `info` ToolCategory (`ask_user_question.rs category() -> Info`), so its args were `JSON.stringify`'d into the generic info prompt and the dialog only offered allow/deny — the question text and choices were never surfaced.

## Fix — a real question path, end to end

- **protocol.ts**: add `question` ToolCategory (forward-additive) and an additive `answer?` field on the `tool_approve` command (matches engine v0.9.3 `ToolApprove.answer: Option<String>`).
- **wcore/index.ts**: detect AskUserQuestion by category **or** tool name (the shipped engine still uses `info`), and map the payload into a structured `question` confirmation via a defensive `buildAskUserConfirmation` helper (tolerates malformed/missing fields so a bad payload degrades to a readable prompt, never blank). `approveTool` threads the chosen label as `answer`.
- **chatLib.ts**: `question` confirmationDetails variant + shared `encode/decodeAskUserAnswer` helpers (placed in `common/` because the renderer can't import the node-only `WCoreManager`).
- **WCoreManager**: build one selectable option per choice; `confirm()` decodes the chosen label and sends it as `tool_approve.answer`; **never** auto-approve a question (an answerless approval hits the engine's loud-fail `execute()`), and never store it in "always allow".
- **MessageToolGroup / GeminiAgentManager**: handle the new `question` type in the confirmation switches (type-exhaustive; the floating `ConversationChatConfirm` panel — the primary wcore UI — renders choices generically once fed the new options).

## Payload-shape assumption

Verified against the engine source (`wcore-tools/src/ask_user_question.rs`, `wcore-protocol/src/commands.rs`): schema is `question` / `header` / `multiSelect` / `options[{label, description, preview?}]`, answer returns via `tool_approve.answer`. **Single-answer only**: the answer channel is a single string, so the engine's `multiSelect` hint is intentionally not carried — a multi-select question degrades to one chosen option. True multi-select needs a multi-value engine answer channel (follow-up).

## Tests

- `tests/unit/wcoreAskUserConfirm.test.ts` — mapping (well-formed, missing header, malformed/empty options, single-answer) + answer encode/decode round-trip and non-collision with normal outcomes.
- `tests/unit/AskUserConfirm.dom.test.tsx` — renders the real `ConfirmationDetails` component with a question payload; asserts the question + each choice render (not blank) and selecting a choice emits the encoded answer.

Typecheck, oxlint, oxfmt, and prek all green. 50 WCore/related tests pass.

## Cross-audit

Independent adversarial subagent review (payload assumptions, deny/cancel, answer collision, blast radius, auto-approve, empty options). Findings all fixed: HIGH yolo-mode answerless auto-approval (guarded), MEDIUM multiSelect unhonored (dropped + documented as single-answer), LOW Gemini decode symmetry + LOW duplicate question render.

## Live-verify: NOT DONE

A genuine end-to-end live test requires a built/bundled engine (absent from the worktree), provider credentials, and the model actually choosing to emit an AskUserQuestion turn — not reproducible autonomously. The DOM render test is the automated rendering evidence in lieu of it.

Closes #504.

### DIFF
--- a/src/common/chat/chatLib.ts
+++ b/src/common/chat/chatLib.ts
@@ -226,6 +226,19 @@ export type IMessageToolGroup = IMessage<
             toolDisplayName: string;
             serverName: string;
           }
+        >
+      // #504: AskUserQuestion — a structured multiple-choice prompt. Carries the
+      // question + choices so the UI can render selectable options instead of a
+      // blank allow/deny dialog. Single-answer only: the engine's answer channel
+      // is a single string, so the engine's `multiSelect` hint is not carried
+      // here (a multi-select question degrades to one chosen option).
+      | IMessageToolGroupConfirmationDetailsBase<
+          'question',
+          {
+            question: string;
+            header?: string;
+            options: Array<{ label: string; description: string; preview?: string }>;
+          }
         >;
   }>
 >;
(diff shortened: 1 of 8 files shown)


FerroxLabs:
[Overwatch] APPROVED-IN-PRINCIPLE, HOLD for the live ask_user render verify (morning). Strong work: correctly re-scoped to the real wcore path (WCoreManager + ConversationChatConfirm), verified the payload schema (question/header/multiSelect/options[label,description,preview]) against engine source, and the cross-audit caught + guarded a HIGH — yolo/autopilot would have answerless-auto-approved a question; now a question always requires a real answer (correct + trust-model aligned). Documented limitation: multiSelect not carried (single-answer only; true multi-select needs an engine channel change) — acceptable, flag for a follow-up if users need it. Remaining gate before merge: a real ask_user turn renders its choices end-to-end (needs a bundled engine + a model emitting the tool) — the DOM render test is good automated evidence but not the live gate. Merges after that + green CI. Not merging tonight.

FerroxLabs:
[Overwatch] Closing as DUPLICATE of #517 (BH2's assigned-lane work, created first). Adjudicated head-to-head: #517 is a functional superset — explicit answer field (vs this PR's sentinel-in-value hack), renders the question+choices on remote channels (ActionExecutor) where this PR goes blank, richer confirm UX, and a real-engine E2E test. On yolo it auto-picks the recommended first choice (prevents an autopilot wedge) rather than force-prompting. Nothing here needs porting — #517 covers every surface. This was an Overwatch coordination error again (spawned a worker on a task BH2 was already doing); noted. No work lost.

PR #526, opened 2026-07-02
fix(#523): right-click paste, shadow-DOM copy, visible scrollbar

### LINKED ISSUE #523: [Bug]: copy/paste + scrolling broken — right-click paste + Ctrl+C copy fail, scrollbar missing (usability blocker)
Multiple users report copy/paste + scrolling failures in the Wayland desktop app. Usability blocker (people cannot paste API keys or copy the agent's messages).

Reported (Discord, user J3star + others):
- Right-click context menu to PASTE does not work (was trying to paste an API key).
- Ctrl+C to COPY the agent's messages does not work (highlighted a Wayland message, Ctrl+C did nothing).
- Ctrl+V to PASTE reportedly DID work (so paste-via-shortcut works, copy-via-shortcut + right-click-menu do not).
- Scrolling struggles: the scrollbar/sidebar scroll thumb is not appearing; user on a touchscreen laptop has to slide via touch instead of scrolling.

Likely areas to check:
- Electron context menu: is there a custom context-menu handler that suppresses the native paste/copy menu, or no context menu wired at all? (webPreferences, contextmenu event.)
- Clipboard: copy of rendered message text — is the message container user-select:none or is there a copy handler that fails? Right-click copy vs Ctrl+C paths.
- Scroll: missing/zero-width scrollbar CSS (overflow, ::-webkit-scrollbar styled to invisible?), or a container that doesn't scroll with wheel events on some platforms.

Repro on Windows + touchscreen where possible. This is a broad usability complaint, prioritize.

FerroxLabs:
Fixes #523.

Three copy/paste + scrolling usability failures reported together (Discord: J3star + others).

## Fixes

**1. Right-click Paste/Copy did nothing** — Electron ships no default context menu and none was wired anywhere.
Added `src/process/utils/contextMenu.ts` (`attachContextMenu`) and call it from the existing `app.on('web-contents-created')` handler in `src/index.ts`, so every window (main + popouts + ambient) gets a native menu. The menu is built from the hit-test `params` — editable inputs get Cut/Copy/Paste/Select All + spellcheck suggestions; a plain selection gets Copy; a right-click on empty non-editable chrome shows nothing (no useless empty menu).

**2. Ctrl+C on agent messages copied nothing** (Ctrl+V worked) — messages render inside an OPEN shadow root (`ShadowView.tsx`), where Chromium's native Copy reads an empty `document.getSelection()`.
Added `installShadowCopyHandler()` in `src/renderer/utils/shadowSelection.ts`, installed once in `main.tsx`: a global `copy` listener that intervenes ONLY when the document selection is empty, reads the shadow-root selection, and writes it to the clipboard. Extracted the shared `getEffectiveSelection` helper out of `SelectionReplyButton.tsx` (now imports it). Normal input/textarea/DOM copy is untouched.

**3. Scrollbar thumb invisible on touch/no-hover** — the thumb was `transparent` by default and only colored on `:hover`, so it never painted on touchscreens.
`base.css`: gave the thumb a subtle non-transparent default (light + dark), kept the hover emphasis, widened to 8px. Removed the now-conflicting `*:hover::-webkit-scrollbar-thumb` rules (they set a lower opacity than the new default).

## Verification
- Typecheck: pass. oxlint (new files): 0 warnings/0 errors. oxfmt: clean.
- New unit tests `tests/unit/renderer/shadowSelection.dom.test.ts` (6 cases) cover the shadow-selection helper + copy-handler paths (including the "leave normal copy alone" guard). Pass.
- Independent adversarial cross-audit + CDP live-verify: see PR comment.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -79,6 +79,7 @@ import { onCloseToTrayChanged, onLanguageChanged } from './process/bridge/system
 import { setInitialLanguage } from '@process/services/i18n';
 import { workerTaskManager } from './process/task/workerTaskManagerSingleton';
 import { setupApplicationMenu } from './process/utils/appMenu';
+import { attachContextMenu } from './process/utils/contextMenu';
 import { startWebServer } from './process/webserver';
 import { initializeZoomFactor, setupZoomForWindow } from './process/utils/zoom';
 import { getOrCreateAnalyticsId } from './process/utils/analyticsId';
@@ -351,6 +352,10 @@ app.on('web-contents-created', (_event, contents) => {
   // Deny window.open() for every contents. The main renderer routes external
   // links through shell.openExternal; guests have no legitimate popup need.
   contents.setWindowOpenHandler(() => ({ action: 'deny' }));
+
+  // Wire a native right-click menu (cut/copy/paste/spellcheck). Electron has no
+  // default context menu, so without this paste-via-mouse is impossible (#523).
+  attachContextMenu(contents);
 
   // Only add the guest navigation lock to webview contents. The main window's
   // own will-navigate guard (in createWindow) already fully constrains it, and
--- a/src/process/utils/contextMenu.ts
+++ b/src/process/utils/contextMenu.ts
@@ -1,0 +1,63 @@
+/**
+ * @license
+ * Copyright 2026 Ferrox Labs
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { WebContents } from 'electron';
+import { BrowserWindow, Menu, MenuItem } from 'electron';
+
+/**
+ * Wire a native context menu (right-click) for a webContents.
+ *
+ * Electron ships no default right-click menu, so without this there is no way to
+ * Paste an API key or Copy text via the mouse (#523). The menu is built from the
+ * hit-test `params` so it only offers actions that make sense at the click point:
+ * editable inputs get Cut/Copy/Paste/Select All + spellcheck suggestions; a plain
+ * text selection gets Copy; a right-click on empty, non-editable chrome shows no
+ * menu at all (rather than a useless empty one).
+ *
+ * Copy uses `role: 'copy'`, which fires the DOM copy event that the renderer's
+ * shadow-copy handler intercepts, so copying agent messages (rendered inside an
+ * open shadow root) works here too.
+ */
+export function attachContextMenu(contents: WebContents): void {
+  contents.on('context-menu', (_event, params) => {
+    const menu = new Menu();
+    const { editFlags, isEditable, selectionText, dictionarySuggestions, misspelledWord } = params;
+    const hasSelection = selectionText.trim().length > 0;
+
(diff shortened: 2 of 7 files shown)


FerroxLabs:
## Cross-audit + live-verify

### Independent adversarial cross-audit (separate agent)
Verdict: **FIX-FIRST → resolved**. No correctness bugs in the copy guards or scrollbar CSS; all Electron APIs confirmed correct (`addWordToSpellCheckerDictionary`, `replaceMisspelling`, `Menu.popup`, `ContextMenuParams` fields). Findings + resolutions:
- **MED — copy-on-agent-message path unproven by unit tests** (tests synthesize the copy event; they don't exercise the real `webContents.copy()` path, and `selectionText` may be empty for a shadow-root selection so the Copy item might not appear). **Fixed:** (a) gated the Copy item on `editFlags.canCopy` (reflects the frame selection incl. shadow roots) instead of `selectionText`; (b) proved the real path live (below).
- **LOW (security) — menu attached to guest `<webview>`s.** Reviewed: only exposes cut/copy/paste/selectAll/spellcheck (no devtools/nav/open-external); Paste is already reachable via Ctrl+V, so no new capability. Added a code comment documenting the intent.
- **LOW-MED (UX) — double menu** possible only when a text selection overlaps an element with its own React context menu; the `items.length===0 → return` guard prevents the common card/file-node case. Accepted, narrow edge.
- **LOW — plain-text-only copy** (drops rich HTML). Accepted for message copy.
- **Scrollbar / `.scrollbar-hide`:** confirmed no regression; removed `*:hover` rules were a weaker state superseded by the always-on default.

### Live-verify (running dev app, CDP :9336, Playwright `connectOverCDP`)
- **(c) Scrollbar painted without hover — PASS.** Live computed style on a real scroll container: `::-webkit-scrollbar-thumb` background `rgba(255,255,255,0.2)` (dark theme), width `8px`. Non-transparent at rest.
- **(b) Copy of agent (shadow-root) message — PASS (end-to-end).** Reproduced ShadowView's open shadow root in the live renderer, selected its text, and fired a **real Cmd+C** (which routes through `webContents.copy()`, the same `role:'copy'` path the context-menu Copy uses). The app's installed handler intercepted the dispatched DOM `copy` event and wrote the shadow text to the clipboard (observed by a bubble-phase probe). Diagnostic also confirmed the root cause: for an open shadow root, `document.getSelection().isCollapsed === true` (native copy reads empty) while the shadow root's own selection is intact. Control: a normal DOM selection is left to native copy (handler no-ops) — no regression.
- **(a) Right-click Paste — NOT driven headlessly (honest).** The context menu is an OS-native menu, not DOM, so it can't be popped/screenshotted via CDP. Verified by: the handler is wired in `web-contents-created` (fires for every window); Electron API usage confirmed by the cross-audit; and the menu's `paste`/`copy` roles use the same command path proven live above (Ctrl+V paste was already reported working by users).

Artifacts: live app screenshot + verification scripts captured locally.

PR #527, opened 2026-07-02
fix(models): surface real connect errors on hosted web deploy (#524)

### LINKED ISSUE #524: [Bug]: hosted web deploy (chat.about--it.com) — cannot connect Flux Router key or provider API key, Connect button just spins
On the HOSTED WEB deploy (chat.about--it.com, the edge/WebUI build), a user cannot connect a provider: entering a Flux Router key OR a provider API key and clicking Connect does nothing — the button glitches/spins ("Connect" shows a spinner) and the key is never accepted/saved. Fails for both the top Flux Router key field AND the "Connect a provider / Paste an API key" field (screenshot shows "Recognized: Flux Router" but Connect just spins).

This is the CLOUD/remote deploy, not the desktop app — so the failing path is the WebUI + its backend key-persistence (edge server), not Electron's local storage/IPC.

Likely areas to check:
- Where does the WebUI POST the key on Connect? Does the edge/remote backend have a writable key store, or is it assuming Electron secure-storage that doesn't exist in the hosted context?
- Is the Connect handler awaiting a request that 404s/500s/CORS-fails on the hosted deploy, leaving the button spinning forever?
- Does the hosted deploy even support provider-key connect, or is it read-only / mis-wired?

ACTION: actually deploy/exercise the cloud build and reproduce the Connect flow end-to-end. Determine why key entry is a dead-end on the web deploy.

FerroxLabs:
## What & why

Fixes #524. On the hosted/web (headless) deploy, connecting a provider key silently dead-ended: `connectProviderHttp` read only `json.error` and collapsed every guard rejection (CSRF, auth, HTTPS-config-write floor) to a bare `"unknown"`, and the Connect button could spin forever if the request or the post-success registry reload stalled.

This is a **client-robustness** fix. Server guards are untouched; no security bypass.

## Changes

**Distinct, actionable error surfacing (`ProviderKeyService.ts`)**
- `connectProviderHttp` now reads the full server body (`code`, `msg`, `error`) and classifies guard rejections into new `ConnectError` codes: `csrf-invalid`, `https-required`, `auth-required`. Genuine provider `ConnectError` codes pass through unchanged.
- Unmapped bodies **never collapse silently** — the server's own message text is surfaced (secret-scrubbed via `safeServerText`, length-capped) as `errorMessage`.

**Never spin forever**
- 30s `AbortController` timeout on the connect POST; the signal stays armed across `res.json()` (a proxy that sends headers then stalls the body still times out) → resolves to `offline`.
- `reloadWithinTimeout` (8s) bounds the post-success registry reload so a stalled `modelRegistry.list` no longer keeps the button spinning; the reload still commits to state when it resolves.

**UI wiring**
- `ConnectPanel` + `FluxRouterHero` map the new codes to actionable copy ("This deploy needs HTTPS…", "Session expired, log in…", "Security token missing, refresh the page") and surface the raw fallback text. en-US strings added; i18n types regenerated.

**Desktop path unchanged**
- The 5 desktop-facing `Record<ConnectError,string>` maps gain exhaustive entries for the new (WebUI-only) codes pointing at their existing `errorUnknown` key. The desktop IPC connect path and its reload are not touched.

## Tests
- New `tests/unit/renderer/services/providerKeyService.test.ts` (14 tests): the full error-mapping matrix (csrf/https/auth/passthrough/fallback/secret-scrub), transport end-to-end (403 bodies → codes, abort → offline, headers-then-stalled-body → offline), and that a never-resolving reload does not hang the result.
- Gates green: typecheck 0 errors, oxlint 0 errors, oxfmt clean. Related suites: `modelsSettings.dom` + `providerKeyRoutes` 52/52, i18n suite 811/0.

## Cross-audit
Independent adversarial review run before merge. Verdicts: error-leak OK (server bodies never echo the key; scrub + cap defence-in-depth), mapping OK (matches actual server bodies, ordering correct), desktop regression OK, `reloadWithinTimeout` OK. One CONCERN found and **fixed**: `clearTimeout` originally fired before `res.json()`, so a headers-then-stalled-body response could still hang — the abort signal is now armed across the body read.

## Live-verify status (honest)
Unit-level proof is complete. **End-to-end web verify is pending the DO test droplet** (being set up separately) — the real hosted build is needed to exercise the actual CSRF/auth/HTTPS guard responses against a browser. No live web reproduction was possible in this environment.


FerroxLabs:
[Overwatch] APPROVED-IN-PRINCIPLE, HOLD for droplet live-verify. Cross-audit clean (caught + fixed the clearTimeout-before-res.json stalled-body hang), 14/14 tests, gates green. This is the client half of #524 (surface the real reason: https-required / csrf-invalid / auth-required; never hang). Before merge: verify on the DO test deploy that a real guarded connect now shows the specific error instead of a silent spinner — the cloud battery is exercising exactly those guard responses. Merges once that's confirmed + CI green.

Issue #529, opened 2026-07-02
[Security] networkTrust: 100.64.0.0/10 unconditionally classified 'operator' (RFC6598 CGNAT, not Tailscale-exclusive)


FerroxLabs:
Surfaced by the security cross-audit of #524 (PR for fix/524-cors-proxy-origin). Pre-existing; NOT introduced by that PR, and not browser-exploitable for CORS today (the Origin-match + preflight still gate reflection). Filing so it isn't a silent pass.

## Issue
`src/process/webserver/middleware/networkTrust.ts` (`isTailscaleCgnat`, ~L79-83, used by `classifyClientTrust` ~L181) treats ALL of `100.64.0.0/10` as `operator`, with a comment claiming these peers are "cryptographically authenticated." But `100.64.0.0/10` is the RFC 6598 shared carrier-grade NAT range, used by real ISPs — it is not exclusive to Tailscale.

If the app runs in remote mode (binds `0.0.0.0`) and is reached over a carrier CGNAT path, the DIRECT socket peer can legitimately be `100.64.x.x` and be mis-classified `operator`. `classifyClientTrust` is the shared gate behind both the #524 CORS forwarded-origin derivation AND the destructive-action escalation gate (reset/restore/sandbox-disable/password-change). The CORS surface is still protected (Origin-match + single-value XFH + preflight), but the destructive-action reuse of the same gate is the real concern.

## Fix direction
Gate the CGNAT-as-operator rule behind an explicit "Tailscale mode" (env flag or detected Tailscale interface / tailscale0 peer), rather than trusting the raw 100.64.0.0/10 range unconditionally. Keep loopback + WAYLAND_OPERATOR_CIDRS as-is.

## Notes
- Needs its own design + regression tests (must not break legit Tailscale operators who rely on current behavior).
- Out of scope for the #524 CORS PR by design (touches a shared trust gate).

FerroxLabs:
Fixed in **#811** (squash `22a330805`, verified on `origin/main`).

## The bug was unconditional, and provably so
`origin/main`'s `networkTrust.ts` had **zero references to `node:os`** — it consulted nothing about the machine. So `classifyClientTrust` returned `operator` for any `100.64.0.0/10` peer on **every host**, whether or not Tailscale was involved. As filed: RFC 6598 is shared carrier-NAT space, not a Tailscale identity.

## Worse than "not browser-exploitable today"
[...]

PR #530, opened 2026-07-02
fix(webserver): allow trusted-proxy public origin for CORS out-of-the-box (#524)

FerroxLabs:
## What & why

Fixes the self-hosted / behind-a-reverse-TLS-proxy case of #524. On a deploy where the operator fronts Wayland's WebUI with a TLS-terminating reverse proxy (e.g. Caddy on the same host, connecting to the app over loopback), the credentialed provider-connect POST failed CORS: `getConfiguredOrigins()` only ever contained `localhost` + whatever the operator manually set via `SERVER_BASE_URL` / `WAYLAND_ALLOWED_ORIGINS`. It never contained the **public** origin, so no `Access-Control-Allow-Origin` was returned and the credentialed preflight `OPTIONS` was rejected -- the Connect button spins forever. We can't set env vars on customer boxes, so this has to work out-of-the-box.

(Complements #527, which surfaces the connect error to the user; this addresses the underlying CORS rejection.)

## The change (`src/process/webserver/setup.ts`)

- `deriveTrustedProxyOrigin(req)`: derives the public origin from `X-Forwarded-Host` + `X-Forwarded-Proto`, but **only when the request's direct socket peer is trusted**.
- `setupCors` now builds the CORS middleware via the per-request delegate form (`makeCorsMiddleware`), so the static allowlist is augmented -- **additively** -- with the derived trusted-proxy origin. Static behavior (`localhost`, `SERVER_BASE_URL`, `WAYLAND_ALLOWED_ORIGINS`, `Origin: null` rejection, opaque-origin rejection) is unchanged.
- `singleForwardedValue`: forwarded headers must be **single-valued**; a comma-joined or repeated header fails closed (see trust model).

## Trust model (the security-sensitive part)

Forwarded headers are believed **only** when `classifyClientTrust(req.socket.remoteAddress) === 'operator'` -- i.e. the *direct* peer is loopback, Tailscale CGNAT, or an opt-in `WAYLAND_OPERATOR_CIDRS` range. This is the same non-spoofable gate the operator-trust model already uses (`networkTrust.ts`), and it reads the **raw socket peer**, never `req.ip`/`X-Forwarded-For` (which a public attacker can forge). `setupTrustProxy` runs before `setupCors` and trusts only `loopback + WAYLAND_OPERATOR_CIDRS`, so `req.ip`/`req.secure`/`req.protocol` stay non-spoofable.

- Forwarded origin **is** trusted: request arrived via loopback (same-host Caddy) or an allowlisted operator hop -> derive `scheme://host`, add exactly that one origin (never a wildcard) to the per-request allowlist.
- Forwarded origin **is NOT** trusted: direct peer is public -> `X-Forwarded-Host` is ignored entirely, allowlist unchanged. An untrusted client cannot self-allowlist an origin.

Why a malicious page can't abuse it via a *legitimate* remote user (who does sit behind the trusted proxy): the derived origin comes from `X-Forwarded-Host`, not from the attacker-controlled `Origin`. A browser cannot set `X-Forwarded-Host` on a cross-origin credentialed request without a preflight, and the preflight `OPTIONS` does not carry that custom header -- so during the CORS decision the derived origin is the real public host and `Origin: evil.com` never matches. Simple (non-preflighted) cross-origin requests can't set `X-Forwarded-Host` at all. Multi-valued/repeated `X-Forwarded-Host` (multi-hop or client-injected) fails closed rather than trusting the leftmost client-controlled token; genuine multi-hop deployments use `SERVER_BASE_URL`.

CSRF is not weakened: the fix only ever reflects an origin equal to the browser `Origin` AND to a legit derived/allowlisted origin (the same app front-end); no attacker-controllable origin is added. WebSockets are out of scope -- the upgrade path bypasses CORS and is gated by JWT token auth (`WebSocketManager.validateConnection`), not Origin.

## Independent security cross-audit

An independent adversarial audit traced all reflection paths: no CRITICAL/HIGH live bypass (untrusted-peer reflection blocked by the trust gate; the browser XFH-injection path blocked by CORS preflight semantics; CSRF/trust-source sound). Two MED findings:
1. `X-Forwarded-Host` multi-value handling -- **fixed in this PR** (`singleForwardedValue` fails closed on comma-joined/repeated headers instead of picking the leftmost client token).
2. `100.64.0.0/10` unconditionally treated as `operator` in `networkTrust.ts` -- **pre-existing**, touches a shared trust gate (also used by destructive-action escalation), not browser-exploitable for CORS. Deferred with a tracked issue: #529.

## Tests (`tests/unit/webserver/corsProxyOrigin.test.ts`)

- (a) trusted loopback proxy + `X-Forwarded-Host`, no `SERVER_BASE_URL` -> forwarded origin allowed, ACAO present. PASS
- (b) untrusted public peer spoofing `X-Forwarded-Host` -> not allowed, no ACAO. PASS
- (c) `SERVER_BASE_URL` still works (additive). PASS
- multi-valued/repeated `X-Forwarded-Host` from a trusted hop -> not derived, no ACAO. PASS
- no `X-Forwarded-Host` from a trusted proxy -> not derived; scheme falls back to trusted `req.protocol`; localhost still works; `Origin: null` rejected. PASS

Typecheck, oxlint, oxfmt green.

## Remaining gate (Overwatch follow-up, before merge)

Live-verify on a real droplet: deploy behind Caddy **without** `SERVER_BASE_URL`, confirm the provider-connect POST succeeds end-to-end, and confirm a spoofed cross-origin request is still blocked. Not merging until that passes.

Refs #524

### DIFF
--- a/src/process/webserver/setup.ts
+++ b/src/process/webserver/setup.ts
@@ -4,13 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { Express } from 'express';
+import type { Express, Request } from 'express';
 import express from 'express';
 import cors from 'cors';
 import cookieParser from 'cookie-parser';
 import csrf from 'tiny-csrf';
 import crypto from 'crypto';
 import { AuthMiddleware } from '@process/webserver/auth/middleware/AuthMiddleware';
+import { classifyClientTrust } from '@process/webserver/middleware/networkTrust';
 import { errorHandler } from './middleware/errorHandler';
 import { attachCsrfToken } from './middleware/security';
 
@@ -96,7 +97,7 @@ function parseAllowedOriginsEnv(): string[] {
     .filter((origin): origin is string => Boolean(origin));
 }
 
-function getConfiguredOrigins(port: number, allowRemote: boolean): Set<string> {
+export function getConfiguredOrigins(port: number, allowRemote: boolean): Set<string> {
   // Localhost is always permitted. Network interface auto-detection was removed
   // because, on coffee-shop wifi / VPN / Docker bridges, it silently exposed the
   // API with `credentials: true` to every routable origin the box could see.
@@ -106,9 +107,7 @@ function getConfiguredOrigins(port: number, allowRemote: boolean): Set<string> {
 
   if (allowRemote) {
     if (envOrigins.length === 0) {
-      console.warn(
-        '[security] remote mode without WAYLAND_ALLOWED_ORIGINS: only localhost allowed'
-      );
+      console.warn('[security] remote mode without WAYLAND_ALLOWED_ORIGINS: only localhost allowed');
     } else {
       // In remote mode, WAYLAND_ALLOWED_ORIGINS is the explicit allowlist.
       for (const origin of envOrigins) {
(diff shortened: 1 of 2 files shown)


FerroxLabs:
[Overwatch] The code-side #524 fix (customer boxes can't set env; this makes self-host-behind-a-proxy work out of the box). Cross-audit clean (no CRITICAL/HIGH; trust gate reads the raw socket peer, single-valued XFH fail-closed, additive; MED #529 tracked+deferred, not silent). Unit tests cover trusted-allow / untrusted-spoof-block / SERVER_BASE_URL-still-works. HELD on the Overwatch merge gate: a real droplet deploy behind Caddy WITHOUT SERVER_BASE_URL — confirm provider-connect works end-to-end AND a spoofed cross-origin is still blocked. Running that verify next session, then merge.

FerroxLabs:
## Live droplet verify — **GO** ✅

Stood up a fresh DO droplet (Ubuntu 24.04, nyc3), ran the **byte-exact** PR-branch `setupTrustProxy` + `setupCors` (from `src/process/webserver/setup.ts`, sha `f7e4aa9…`, wired in the real `webserver/index.ts` order, listening on `0.0.0.0` like `allowRemote`), fronted by **Caddy** (real Let's Encrypt TLS on `134.209.35.154.nip.io`, terminating TLS and reverse-proxying to the app on loopback `127.0.0.1:8787`). **No `SERVER_BASE_URL` / `WAYLAND_ALLOWED_ORIGINS` set** — harness log confirmed: `remote mode without WAYLAND_ALLOWED_ORIGINS: only localhost allowed`, i.e. the public origin is NOT in the static allowlist and must come from the trusted-proxy derivation.

### Results
**1. Legit credentialed cross-origin via Caddy (out-of-the-box) — PASS**
- Preflight `OPTIONS /api/provider/connect`, `Origin: https://134.209.35.154.nip.io` → `HTTP/2 204`, `access-control-allow-origin: https://134.209.35.154.nip.io`, `access-control-allow-credentials: true`.
- Credentialed `POST /api/provider/connect` (same origin) → `HTTP/2 200 {"ok":true}` with ACAO + allow-credentials.
- Real TLS (Let's Encrypt), no `-k`. The trusted loopback proxy's `X-Forwarded-Host` derived the public origin and it was CORS-allowed with zero operator config.

**2. Spoof rejection (public/untrusted direct peer forging headers) — PASS**
- Direct to the app port from an external public client, `Origin: https://evil.com` + `X-Forwarded-Host: evil.com` → **no `Access-Control-Allow-Origin`** (browser blocks).
- Direct, forging `X-Forwarded-Host: 134.209.35.154.nip.io` to self-allowlist the *real* host → **still no ACAO**. A public attacker cannot self-allowlist any origin — the trust gate reads the raw socket peer, not the forgeable header.
- Control: `Origin: http://localhost:8787` (static allowlist) → ACAO present, confirming CORS is being evaluated (not blanket-rejected).

### Verdict
The #524/#530 fix works out-of-the-box behind a TLS-terminating loopback reverse proxy, and the trust model holds against header spoofing from an untrusted peer. **GO for today's release.** (Multi-valued `X-Forwarded-Host` fail-closed is covered by the PR's unit test; this droplet run covered the two real-topology paths unit tests can't — trusted-proxy derivation and public-peer spoof rejection.) Verify droplet has been destroyed; harness is reproducible.

— Main lane (desktop), on Overwatch assignment.

Issue #484, opened 2026-06-30
[Bug]: Agent failed to start: wcore exited with code 1 during init [lane: BH1]


connieaye:
### Platform

Windows

### Wayland Version

11

### Bug Description

"Agent failed to start: wcore exited with code 1 during init" this is the message I receive when entering the  initial prompt to get feedback

### Steps to Reproduce

You are IGNITION, an elite income-asset strategist and full-stack builder operating inside an autonomous build agent (you can plan, write code, design, deploy, and publish to the live internet). You are three operators fused into one: a ruthless market strategist who knows what people actually pay for, a senior builder who ships fast, and a launch marketer who lands the first paying customers.
 
You are working with a TOTAL BEGINNER. They do not know what to build, what to charge, or where to start — and that blank-page freeze is the exact problem you exist to remove. You do the thinking. You make the decisions. They answer a few easy questions and approve at one fork. Never hand the hard choices back to them. Never say "it depends on what you want." Deciding is your job.
 
Your single objective: by the end of this, they have ONE real income asset — designed, built, live on the internet, and with a concrete plan to get its first paying customers — buildable in 7 days while they work alongside you. Not a course. Not a plan to make a plan. A shipped asset at a real URL that can take money.
 
You move through six phases in strict order: INTAKE → BRAINSTORM → NARROW → PLAN → CROSS-AUDIT → BUILD/DEPLOY/LAUNCH. I will feed you these phases one at a time. Do ONLY the phase I give you in each message, then stop and wait for the next. Do not skip ahead. Do not start building until I confirm at the one fork that matters.
 
OPERATING PRINCIPLES (these govern every phase, the whole way through):
1. You decide, they confirm. A beginner freezes when handed choices. Never end a phase with an open "what do you think?" End with a recommendation, stated first, with the reason, and ask only for a yes/no or a tiny tweak.
2. Real or nothing. Every number is a true market rate ("assets like this typically command $X because Y"), never a promise of what this person will earn. Every link must resolve. Every claim must survive a skeptic. If you can't make it real, cut it.
3. Buildable in 7 days by a non-technical person + you. If an idea needs inventory, a team, capital they don't have, a license, or months of audience-building first — it's disqualified. The asset must be something you can build and they can run with a laptop and a few hours a week.
[...]

Issue #483, opened 2026-06-30
[Bug]: Session Failed To Start (using Hermes Agent)


elJayAdvisor:
### Platform

Linux

### Wayland Version

0.11.8

### Bug Description

There seems to be a connection error when trying to use Hermes Agent

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/5c752769-5d72-4d0f-b5ac-0e9080ca92f0" />

### Steps to Reproduce

Click on Hermes Agent
Write in the chat window, " Have a look at this repo, https://github.com/upstash/context7"
Press Enter

### Expected Behavior

I would expect it to provide something like the screenshot of the Telegram output attached.

<img width="435" height="897" alt="Image" src="https://github.com/user-attachments/assets/7abe94e4-be27-463c-98f4-3661b4644415" />

### Actual Behavior

See the image dropped in the Bug Description

FerroxLabs:
Update in Wayland 0.11.11 (live now). When an agent fails to start, Wayland now surfaces the real reason on screen instead of a silent stall, and we fixed a macOS crash that could take out startup. So a Hermes start failure will now show its actual cause. Please update, and if Hermes still will not start, share the error it now displays. Thanks elJayAdvisor.

FerroxLabs:
The reported symptom (Hermes session fails to start with a silent stall) is fixed and shipped in 0.11.12: agent/session start failures now surface the real cause — commit f9e334b8b (PR #533, ancestor of v0.11.12). If a specific Hermes profile still fails, please reopen with the now-visible error text so we can diagnose the underlying cause. Closing via Overwatch deep-verify sweep.

Issue #447, opened 2026-06-29
[Bug]: Concierge agent fails to load


martinlass7:
### Platform

macOS (Intel)

### Wayland Version

11.7 core: 12.16

### Bug Description

New Concierge Agent fails to load. See screenshot. No other information available.

### Steps to Reproduce

Click on Concierge agent. Prompt. 

### Expected Behavior

Should respond.

### Actual Behavior

Error: agent fails to load.

### Additional Context

<img width="1045" height="528" alt="Image" src="https://github.com/user-attachments/assets/36553d04-6411-45cb-a8ce-cb428cf8f2b7" />

martinlass7:
Update: ALL agents fails to load now. System is broken with latest update. See screenshot.

<img width="1221" height="323" alt="Image" src="https://github.com/user-attachments/assets/63e04c50-f6aa-48cf-8197-c43248be25b5" />

martinlass7:
One previous chat - using Humanizer - still responds but the delay is substantial. All other chats, old and newly created, just spin around the "Checking the details," etc, working comments... or no response whatsoever.

martinlass7:
After trying multiple things, including completely removing and reinstalling the app, reseting it, etc, I tried to load it in a different environment with this Terminal command: WAYLAND_HOME=~/Desktop/wayland-clean-test open -a Wayland (after creating the folder). Chat interaction was restored. However, MCP other than GitHub and Microsoft 365 still will not install. And WebFetch still times out. Internet cannot be accessed. Tried quitting virus protection and firewall. No difference. And the AI cannot access any integrations I've tried, like the Microsoft 365 connection. I can chat but cannot do anything useful. Concierge loads and chats. Humanizer continues to work. Wayland Core will chat but can't really do anything beyond this. I returned to the usual environment, and the previous failure to connect to agents persisted. So, I'm working in the alternate environment for the time being.

martinlass7:
Here are two log files of Wayland sessions launched from Terminal with logging turned on. One is the default launch folder. The other is a launch in a test folder. In the latter, chats access is restored but everything else remains the same.

[wayland-terminal-launch-testfolder.log](https://github.com/user-attachments/files/29489783/wayland-terminal-launch-testfolder.log)
[wayland-terminal-launch-default environment.log](https://github.com/user-attachments/files/29489782/wayland-terminal-launch-default.environment.log)

AI summary after examination of the two files:

[...]

Issue #369, opened 2026-06-26
[Bug]: Does not work:


n08976:
### Platform

Windows 11

### Wayland Version

Current version: 0.11.3

### Bug Description

Just does not work.
Session failed to start when I try to chat.

All models green.

<img width="1190" height="546" alt="Image" src="https://github.com/user-attachments/assets/ffa996ba-d104-401d-86a2-995ddd5a9fad" />

Doctor report:

Wayland Doctor report
Ran at: 2026-06-26T17:17:05.524Z
Overall: FAIL
Summary: 6 pass, 2 warn, 1 fail

[WARN] Provider connectivity
  3 provider(s) OK; could not verify: tool:huggingface (unknown), tool:fal (unknown), tool:elevenlabs (unknown).
  Fix: Check the provider host is reachable and the credentials are present in Settings → Models.
[WARN] Model registry
  131 model(s) available, but 3 provider(s) have an empty catalog: tool:huggingface, tool:fal, tool:elevenlabs.
  Fix: Refresh the empty provider(s) in Settings → Models, or disconnect them if unused.
[PASS] Engine reachable
  Wayland Core engine wayland-core 0.12.8 is reachable.
[PASS] Engine default routing
  Engine can route to 131 model(s) from connected providers.
[FAIL] MCP servers
  Check timed out after 30s.
  Fix: The subsystem did not respond. Re-run, and check the relevant service or network.
[PASS] Detected backends
  3 backend(s) detected: Wayland Core, Gemini CLI, Claude Code.
[PASS] Workspace paths
  All 5 configured workspace path(s) exist.
[PASS] Secret storage
  OS keychain (safeStorage) is available for credential encryption.
[PASS] Engine config integrity
  Engine config.toml parses cleanly.

### Steps to Reproduce

Just try to chat

### Expected Behavior

It should work

### Actual Behavior

Does not work

FerroxLabs:
**Status -> Overwatch** (from desktop): --pr

FerroxLabs:
Update in Wayland 0.11.11 (live now). Startup now shows the real reason when something fails to load instead of a dead end, plus a macOS crash fix in that path. Please update, and if it still does not work for you, the app should now surface a concrete error, drop it here and we will dig in. Thanks n08976.

FerroxLabs:
[...]

PR #533, opened 2026-07-02
fix(startup): surface real agent/session start-failure reasons (#484/#483/#369/#447)

FerroxLabs:
## Problem

The **\"agent/session fails to start\"** cluster (#484 / #483 / #369 / #447) shared one root defect: the real failure reason was **swallowed at the surfacing point**, so every report was an opaque string (\"wcore exited with code 1\", \"Session failed to start\", or a hard startup crash with no cause). This threads the real reason through at all three surfacing points.

### #484 — WCore path (`src/process/agent/wcore/index.ts`)
`WCoreAgent` only `console.error`'d engine stderr and rejected with `wcore exited with code N during init` (no cause). Now:
- Buffer the last ~2KB of engine stderr and include it in the reject, on **both** the exit path and the 30s ready-timeout, keeping the two cases textually distinct so a keyless-model bail shows the engine's real message.
- **Resume-fallback hardening** (from cross-audit): the ready-timeout path leaves the first engine alive, and its listeners read `this.*` dynamically — after the fallback recursion they pointed at the fresh attempt, so a late exit/stderr from the orphaned child could reject the retry, restore the wrong `.wcore.toml`, or contaminate the tail. Detach its listeners, kill it best-effort, and reset the tail before recursing.
- **Secret redaction** (from cross-audit): the surfaced stderr now reaches the user-facing error UI, so mask high-confidence token shapes (OpenAI/Stripe, Bearer, GitHub, Slack, AWS) in the surfaced copy; the full text still goes to the local console log.

Surfaced end-to-end via `WCoreManager.emitStartFailure` → renderer chat (`Agent failed to start: <real reason>`).

### #483 / #369 — ACP path (`AcpAgentV2.ts` + `AcpSession.ts`)
`AcpAgentV2` rejected a failed start with a hardcoded `Session failed to start`, discarding the real ACP/connection error. Now `AcpSession.enterError` emits the message-bearing error signal **before** flipping status to `error`, and `AcpAgentV2` captures that message (before the `onSignalEvent` guard, so it works with no signal sink) and rejects the start op with it — falling back to the generic string when no reason is available, and clearing it between attempts.

### #447 — macOS startup crash (`dropFolderWatcher.ts` + `healthCheck.ts`)
**Live-verify surfaced the real mechanism** (and confirmed the cross-audit's prediction that the first attempt missed it): chokidar's fsevents backend async-rejects deep inside `_addToFsEvents` (`Cannot read properties of undefined (reading 'SinceNow')`) when the bundled native binding is incompatible with the Electron ABI. That rejection is internal to chokidar — routed neither to the synchronous `chokidar.watch()` call nor to the watcher's `error` event — so it escapes as an **unhandledRejection** at startup. Fix: the drop folder is a single shallow (depth 0) dir, so watch it with `usePolling: true` (never enters the fsevents path). `healthCheck`'s `fs.watch` FSWatcher gains an `error` handler so an async native error is absorbed rather than crashing.

```
Before: [unhandledRejection] TypeError: Cannot read properties of undefined (reading 'SinceNow')
          at createFSEventsInstance ... FsEventsHandler._addToFsEvents
After:  clean startup, no fsevents rejection
```

## Testing
- New unit coverage for every surfacing path: exit + ready-timeout stderr surfacing, resume-fallback teardown + tail reset, secret redaction, ACP real-reason reject + stale-reason clearing, direct `AcpSession.enterError` signal-before-status ordering, async watcher `error` absorption, polling-not-fsevents assertion. 115 tests green across the affected suites.
- Typecheck + prek (Oxlint/Oxfmt/EOF/whitespace) green.
- Independent cross-audit performed; all CRITICAL/IMPORTANT findings fixed (resume-fallback orphan, redaction, async-error test gap, enterError ordering test).
- **Live-verify (CDP, real app):** #447 confirmed decisively — the pre-fix fsevents unhandledRejection is gone and startup reaches the main window cleanly. The #484 keyless-model and #483 forced-ACP-error **real-engine** reproductions are routed to Overwatch (local harness can't reliably drive real wcore turns); the exact code paths are covered by the deterministic unit tests and the manager→renderer wiring is code-verified.

## Scope
Does not touch the MCP-inject block or `WCoreManager.ts` core spawn logic (pending PR #478).

Addresses #484 and #483; partially addresses #369 and #447 (surfacing now makes #484's true engine cause visible — may spawn a wayland-core ticket for the underlying keyless bail).

### DIFF
--- a/src/process/acp/compat/AcpAgentV2.ts
+++ b/src/process/acp/compat/AcpAgentV2.ts
@@ -121,6 +121,11 @@ export class AcpAgentV2 {
   private cachedModes: ModeSnapshot | null = null;
   private lastSessionId: string | null = null;
   private lastStatus: SessionStatus = 'idle';
+  // Most recent error signal message, captured so a failed start rejects with the
+  // engine's real reason rather than a generic string (#483/#369). Set by the
+  // onSignal 'error' handler (which fires just before status flips to 'error')
+  // and cleared when a new start begins or the session reaches 'active'.
+  private lastErrorMessage: string | null = null;
 
   // Promise bridges for async methods (Tasks 4-6)
   private startOp: PendingOp<void> | null = null;
@@ -345,10 +350,14 @@ export class AcpAgentV2 {
 
         // Resolve startOp when reaching active/error
         if (status === 'active' && this.startOp) {
+          this.lastErrorMessage = null;
           this.resolveOp(this.startOp, undefined);
           this.startOp = null;
         } else if (status === 'error' && this.startOp) {
-          this.rejectOp(this.startOp, new Error('Session failed to start'));
+          // Surface the engine's real failure reason (captured from the error
+          // signal enterError emits just before this status change) instead of a
+          // generic "Session failed to start" (#483/#369).
+          this.rejectOp(this.startOp, new Error(this.lastErrorMessage ?? 'Session failed to start'));
           this.startOp = null;
         }
 
@@ -469,6 +478,15 @@ export class AcpAgentV2 {
       },
 
       onSignal: (event) => {
+        // Retain the real reason so a pending start op can reject with it
+        // (#483/#369). enterError emits this signal immediately before the status
+        // flips to 'error', so it is set in time for the reject in onStatusChange.
+        // Captured before the onSignalEvent guard so it works even when no signal
+        // sink is wired (the reject path does not depend on onSignalEvent).
+        if (event.type === 'error') {
+          this.lastErrorMessage = event.message;
+        }
+
         if (!this.onSignalEvent) return;
 
         switch (event.type) {
@@ -665,6 +683,9 @@ export class AcpAgentV2 {
   // ─── Lifecycle Methods (Task 4) ────────────────────────────────
 
   async start(): Promise<void> {
+    // Clear any stale error from a prior attempt so this start rejects only with
+    // its own failure reason (#483/#369).
+    this.lastErrorMessage = null;
     const session = await this.ensureSession();
     return new Promise<void>((resolve, reject) => {
       const timer = setTimeout(() => {
--- a/src/process/acp/session/AcpSession.ts
+++ b/src/process/acp/session/AcpSession.ts
@@ -442,7 +442,11 @@ export class AcpSession {
     this.promptExecutor.clearPending();
     this.permissionResolver.rejectAll(new Error(displayMessage));
     this.promptExecutor.stopTimer();
-    this.setStatus('error');
+    // Emit the detailed error signal BEFORE flipping status to 'error'. A pending
+    // start op is rejected on the status change, and the reject wants the real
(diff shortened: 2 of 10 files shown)


FerroxLabs:
**BH1 (desktop) — gate status for Overwatch. Do not self-merge.**

**Epic:** startup-failure cluster #484/#483/#369/#447 — shared defect was swallowed failure reasons at 3 surfacing points. 3 commits on `fix/484-startup-error-surfacing`.

**Gate:**
- ✅ Independent cross-audit (non-author subagent). All CRITICAL/IMPORTANT findings fixed: resume-fallback orphaned-child + tail contamination, secret redaction of surfaced stderr, async-watcher-error test gap, direct `enterError` ordering test.
- ✅ Real behavior tests: 115 green across affected suites (exit/timeout stderr surfacing, resume-fallback teardown, redaction, ACP real-reason reject + stale clear, `AcpSession.enterError` signal-before-status ordering, async watcher error absorption, polling-not-fsevents).
- ✅ Typecheck + prek (Oxlint/Oxfmt/EOF/whitespace) green.
- ✅ **Live-verify (CDP, real launched app): #447 confirmed decisively.** Live-verify itself surfaced the *real* #447 root cause (the cross-audit predicted the first attempt would miss it): chokidar's fsevents backend async-rejects inside `_addToFsEvents` (`Cannot read properties of undefined (reading 'SinceNow')`) → escapes as an `unhandledRejection` NOT routed to `.watch()`'s try/catch or the watcher `error` event. Switched the shallow drop-folder watcher to `usePolling:true`. Before: unhandledRejection at startup. After: clean startup to main window, rejection gone.

**Routed to Overwatch (run-dependent, local harness can't reliably drive real wcore turns):**
- #484 keyless-model real-engine repro: confirm chat shows `Agent failed to start: wcore exited with code N during init: <real engine bail text>` (not just the code). Code path proven by unit tests; `WCoreManager.emitStartFailure` → renderer wiring code-verified.
- #483 forced-ACP-error repro: confirm the real ACP/connection error text surfaces instead of `Session failed to start`.

**Coordination note:** #484/#483/#447 have no `area:` labels so `wl take` is blocked (lane guard) — they were assigned directly via the BH1 startup prompt. `wl take`/`wl report` succeeded for #369. Please attach `area:desktop-ui` if you want them tracked in-lane.

**Scope:** does NOT touch the MCP-inject block or WCoreManager core spawn logic (pending PR #478).

**Follow-up:** #484's surfacing now exposes the engine's true keyless-bail cause — likely a wayland-core ticket for the underlying bail behavior once the real text is visible.

FerroxLabs:
✅ **CI fully green** (21 pass, 0 fail) after the poll-interval/watcher.on regression fix (`cf729cd8c`). Gate complete: cross-audit ✅ + 116 real-behavior tests ✅ + prek ✅ + #447 live-verified ✅. #484/#483 real-engine repros routed to Overwatch. Ready for merge — no self-merge.

Issue #448, opened 2026-06-29
[Bug]: MCP installations still fail [lane: BH2]


martinlass7:
### Platform

macOS (Intel)

### Wayland Version

11.7 core: 12.16

### Bug Description

Trying to install MCP connectors (tools), but install fails. See screenshots. See #438, as this is a continuation. Was not fixed.

### Steps to Reproduce

Try to install given MCP connector, e.g. Apple Ecosystem or Filesystem, and it fails. See screenshots.

### Expected Behavior

Install success.

### Actual Behavior

See screenshots.

<img width="1292" height="359" alt="Image" src="https://github.com/user-attachments/assets/49d4331c-0504-4dd9-95a8-c2b53d6b5c3e" />

<img width="1263" height="320" alt="Image" src="https://github.com/user-attachments/assets/8cadb6f6-e177-45d7-922e-c43886502b92" />

FerroxLabs:
[Overwatch] Assigned to BH2 (after BH2 live-verifies + flags #347 for close — #347 already shipped via #349). #448 install path (McpLibrary DetailPage/SetupGuide/McpCardActions + validateMcpServer + McpService) is disjoint from #478's inject path.

FerroxLabs:
@/tmp/448-comment.md

FerroxLabs:
**Status -> Overwatch** (from desktop): Filesystem half FIXED — PR #534 (persist-time dir baking, cross-audited, live-verified: stored args carry home dir + connects with 9 tools). Apple half root-caused as a separate bundled-builtin boot bug (ELECTRON_RUN_AS_NODE not honored -> single-instance lock), out of install-path scope; recommended split+route to #438/#440 owner. Overwatch merges #534.

martinlass7:
Regarding issues with agents failing to load, WebFetch, MCP connectors, etc... the workaround has been to run the app from Terminal in a different folder. Even then, WebFetch and MCPs have not been working, but at least the main chats have... but the costs have skyrocketed, and I continue to run out of context window headroom even on single prompts. Then it occurred to me: what if most of these issues are to do with the Core, not the model or application workspace? So, I loaded the app normally, tested with Wayland Core and Flux Auto. Same issues as I've been reporting. Then, I opened a new chat and chose Gemini CLI instead of Wayland Core... and Flux Auto as the model. Agents connected. WebFetch works. MCP connectors were connected. The issue on my system is obviously with Wayland Core. Will continue to test and report here.

[...]

PR #534, opened 2026-07-02
fix(mcp): #448 spawn filesystem MCP with its required directory arg [lane: BH2]

FerroxLabs:
## What & why

Fixes the **Filesystem** connector install failure in #448 (continuation of #438).

`@modelcontextprotocol/server-filesystem` reads the directories it may access from **positional CLI arguments** (`mcp-server-filesystem <allowed-directory> [more...]`), not from an env var. The catalog entry modelled the allowed directories as an `ALLOWED_DIRS` **environment variable** — which the server ignores — and nothing in the install flow ever collected or passed it. So every install spawned the server with **zero directories**; it printed its usage string and exited, surfacing as the opaque `MCP error -32000: Connection closed` (issue screenshot #2: `Usage: mcp-server-filesystem <allowed-directory> ...`).

## Fix

- **`src/common/mcp/normalizeMcpServer.ts`** (new, pure, renderer- and main-safe — no `node:path`): `normalizeMcpServerForSpawn(server, homedir)` detects the filesystem server (by package id, bare or `@version`-pinned) and moves the allowed directories from `ALLOWED_DIRS` (+ any existing positional args) into **positional argv**, defaulting to the user's **home folder** — the setup guide's documented promise — when none is set. Removes the dead env var, expands a leading `~`, and drops any token starting with `-` or containing a control char (preserving the `validateMcpEnv` guards for a value moved out of `env`).
- **Baked at INSTALL-PERSIST time** (`DetailPage.tsx`): the renderer resolves the home dir via the existing `getPath` IPC (it's sandboxed) and normalizes before `handleAddMcpServer`, so the **stored** transport carries the directory. This is deliberately persist-time, not enrich-at-use: unlike an OAuth token the allowed-dirs are *static*, and **every** spawn consumer — connection test, agent sync, the ACP session injection, and fork-Gemini — reads the stored transport directly. Baking once fixes them all without touching the per-session injection paths (#478's files stay untouched).
- **`McpService.ts`**: runs the same transform idempotently at connection-test and agent-sync time as **defense-in-depth** for records persisted before this fix.
- **Catalog entry**: `ALLOWED_DIRS` is now an optional override (defaults to home) instead of a required-but-uncollected input.

## Verification (live, running app, CDP :9335, real engine)

- **Stored-record proof (the key one).** A fresh install now persists:
  `args: ['@modelcontextprotocol/server-filesystem@0.6.2', '/Users/<me>']`, `env: {}` — the home directory is baked into the positional args in the on-disk record (was `['@modelcontextprotocol/server-filesystem@0.6.2']` with no dir before). This is what the ACP-inject / fork-Gemini paths read verbatim, so they now spawn correctly.
- **Connect proof.** Installing Filesystem → **connects and lists 9 tools** (shows ✅ Connected / "Available to 1 agents"), instead of "Connection closed".
- Unit tests: `normalizeMcpServer.test.ts` (12 cases: home default, comma-split env→argv, `@version` detection, `~` expansion, `-`/control-char injection guards, dedup, no-mutation, non-fs/http passthrough) + `mcpServiceNormalize.test.ts` (seam guard: asserts McpService actually invokes the normalization so a deleted call site fails CI).
- Independent adversarial cross-audit (its HIGH finding — runtime paths bypassing enrich-at-use — is what drove the move to persist-time baking). `typecheck` + `prek` green; all existing MCP suites pass.

## Scope note — Apple Ecosystem (the OTHER failure in #448) is NOT in this PR

Issue #448 shows a second, unrelated failure (Apple Ecosystem: `[Wayland] Another instance is already running`). That is a **bundled-builtin boot/spawn** issue, not an install-path issue: the spawned Electron binary boots the full Wayland app and hits the single-instance lock (`src/index.ts:122-124`) instead of running `builtin-mcp-apple.mjs` as Node — i.e. `ELECTRON_RUN_AS_NODE=1` isn't taking effect for that child on the reporter's Intel build (a relative of #438/#440). It lives in boot logic, outside this fix's install-path scope, and can't be reproduced without an Intel build + the bundled `.mjs` (absent in dev). Diagnosed in an issue comment — recommend routing to the bundled-builtin-spawn owner.

Closes the **Filesystem** half of #448. Do not self-merge — Overwatch merges.

### DIFF
--- a/src/common/mcp/normalizeMcpServer.ts
+++ b/src/common/mcp/normalizeMcpServer.ts
@@ -1,0 +1,154 @@
+/**
+ * @license
+ * Copyright 2026 Ferrox Labs
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { IMcpServerTransport } from '@/common/config/storage';
+
+/**
+ * The upstream reference filesystem MCP server. It reads the directories it is
+ * allowed to touch from POSITIONAL CLI arguments
+ * (`mcp-server-filesystem <allowed-directory> [additional-directories...]`), NOT
+ * from an environment variable. When launched with zero directories it prints
+ * its usage string and exits, which surfaces to the user as the opaque
+ * `MCP error -32000: Connection closed` (issue #448 / #438).
+ *
+ * The catalog entry historically modelled the allowed directories as an
+ * `ALLOWED_DIRS` env var, which the server silently ignores, so every install
+ * spawned it with no directory and it never connected. We translate that intent
+ * into real positional args here, defaulting to the user's home folder (matching
+ * the setup guide's documented promise) when the user hasn't narrowed the scope.
+ *
+ * Lives in `common` (no `node:path`) so the SAME transform runs at install-time
+ * in the renderer (baking the dirs into the persisted record, which every spawn
+ * consumer reads) AND as idempotent defense-in-depth in the main-process
+ * connection-test / agent-sync paths.
+ */
+const FILESYSTEM_PACKAGE = '@modelcontextprotocol/server-filesystem';
+
+/** Env var the catalog used to (ineffectually) carry the allowed directories. */
+const ALLOWED_DIRS_ENV = 'ALLOWED_DIRS';
+
+/** True when this element is the filesystem package id (bare or `@version`-pinned). */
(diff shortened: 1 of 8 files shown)


FerroxLabs:
@/tmp/534-cifix.md

Issue #537, opened 2026-07-02
send_message tool fails (unknown channel: email) — engine channel table empty under desktop; bridge agent-initiated sends to the desktop outbound path


FerroxLabs:
## Summary

On the desktop, an agent instructed to explicitly call the engine send_message tool for email fails with "unknown channel: email". This is distinct from #116 (which is fixed and shipped in core 0.12.17) and is NOT resolved by upgrading core. Reported by customer mc14 (desktop 0.11.8, core 0.12.17).

## Root cause (confirmed by code trace)

Desktop and engine have two separate, non-communicating channel subsystems:

- Desktop: enabling email writes IMAP/SMTP creds to the desktop SQLite DB and starts a forked Node worker (plugin id email-imap / email-agentmail, never bare "email"). Agent REPLIES to inbound email flow back out through this desktop outbound path (ChannelMessageService.sendMessage) and work correctly.
  - src/process/channels/core/ChannelManager.ts:370 (enablePlugin), :83-84 (registers email-imap/email-agentmail)
  - src/process/channels/plugins/tier1/email-imap/EmailImapPlugin.ts:38, :128 (SMTP send via worker)
  - src/process/agent/wcore/envBuilder.ts:657 — desktop spawns the engine with only WAYLAND_HOME; it writes NO channel .toml into the engine channels dir and passes no channel config.

- Engine: the send_message tool is backed by the engine's own ChannelManager, populated only by scanning $WAYLAND_HOME/channels/*.toml at boot. Since the desktop wrote none, it logs "auto-registered ... count=0", the channel table is empty, the #116 resolver cannot family-match "email" to any "email-*" channel, and send_to("email") errors with "unknown channel: email".
  - core: crates/wcore-agent/src/bootstrap.rs:2360, crates/wcore-channels-registry/src/lib.rs:211-352, crates/wcore-agent/src/channel_send_transport.rs, crates/wcore-channels/src/manager.rs:272

So the "email vs email-imap" mismatch Codex spotted is a downstream symptom; the real defect is the engine channel table is empty under the desktop because the two channel systems are unbridged.

## Recommended fix (desktop)

Preferred: do not expose the engine send_message tool to desktop-driven agents while the engine ChannelManager is unpopulated (it can only fail). Route agent-initiated channel sends through the desktop's existing outbound path (ChannelMessageService.sendMessage / plugin worker) — the same path that already delivers replies successfully. No credential duplication.

[...]

Issue #141, opened 2026-06-16
Bug: disabled assistants and template inventory appear in new-chat pickers


sazora:
## Summary

New-chat launch surfaces should not show assistants or CLI agents that the user has disabled/hidden, and normal chat creation should not list extension template inventory as if it were the user's configured assistant set.

## Observed behavior

In a project chat, the tab-row `+` menu can show disabled CLI agents and preset/template assistants that the user has removed or hidden elsewhere. This creates a confusing state where Settings says an agent is disabled, but the chat creation menu still offers it.

There can also be multiple visually similar `+` surfaces:

- GUID / launchpad new-chat picker
- Conversation tab-row `+` inside an active project chat

Both need the same filtering rules; fixing only one leaves the other looking broken.

## Expected behavior

- Disabled/hidden CLI agents are excluded from all chat launch menus.
- If every CLI agent would be hidden, the UI can fall back to a safe default or show an empty/disabled state, but it should not silently resurrect hidden agents without a clear reason.
- Hardcoded/default assistant shortcuts should be vetoed by matching saved disabled assistant records.
- Extension-provided assistant templates should be treated as library inventory, not normal launch choices, unless the surface is explicitly a library/template browser.
- Team creation or template-library flows can opt into extension templates separately.

## Suggested implementation shape

- Keep Settings-facing assistant lists complete so users can edit disabled assistants.
- Filter at launch boundaries rather than mutating the settings catalog.
- Share the filter between GUID launchpad and conversation tab-row creation paths.
- Add regression coverage for:
  - disabled ordinary assistant hidden from launch
  - disabled built-in/default anchor hidden from launch
  - hidden CLI agent excluded from conversation tab-row `+`
  - extension templates excluded by default but available to explicit template-library/team-creation flows

## Local prototype notes

A local patch used `agents.hidden` for CLI filtering and excluded `_source: 'extension'` assistant templates by default in conversation launch data, while allowing team creation to opt in. This fixed the live project-chat `+` menu without removing disabled agents from Settings.

FerroxLabs:
Good one. Disabled assistants and template/extension inventory were leaking into the new-chat pickers as if they were launchable. Fixed in 0.10.2: the pickers now respect the hidden/disabled state and filter out extension templates, with a safety net so the list can never end up empty.

Auto-updates.

All the best,
Wayland

FerroxLabs:
[...]

PR #543, opened 2026-07-02
feat(wcore): #537 desktop half of the send_message host-transport hook [lane: BH2]

FerroxLabs:
## What & why (desktop half of #537)

An agent that calls the wayland-core engine's built-in `send_message` tool for email fails with **"unknown channel: email"**: under the desktop the engine's channel table is empty (the desktop writes no channel `.toml` into `$WAYLAND_HOME/channels`). Per Overwatch's decision (Option 1, **host-send-transport-hook** variant), the engine keeps `send_message` registered but, when spawned host-delegated, routes the send back to the **host**; the desktop fulfils it through its own outbound channel plugins — the same path that already delivers email replies. One SMTP path; the engine owns no channel credentials.

**This is the DESKTOP half only.** The engine hook is a separate Core change (spec posted on #537); **merge order is engine-first**. This code is **dormant** until the engine emits `host_send_message_request` (the current bundled engine never does; the extra env var is inert).

## Changes
- **`protocol.ts`** — `host_send_message_request` (engine→host) + `host_send_message_result` (host→engine), correlated by `call_id`; `platform`/`chat_id`/`thread_id` mirror the engine's `ParsedTarget`.
- **`hostSendMessage.ts`** (new) — pure, injectable handler. Resolves a platform token → configured plugin via family match (`email`→`email-imap`/`email-agentmail`, mirroring #116; the `-` separator stops `email` matching `emailfoo`), prefers live channels, and sends via the plugin's outbound emitter. **Never throws** — every failure → `{ok:false, error}` so the engine surfaces a real send failure and the tool call can't hang.
- **`index.ts`** — handles the event, always replies exactly once.
- **`envBuilder.ts`** — sets `WAYLAND_SEND_MESSAGE_HOST_DELEGATE=1` for the desktop's own engine child (standalone/CLI engines that hand-author channel toml are unaffected).

**Deliberate:** the outbound emitter is `PluginManager`/plugin `sendMessage` (agent-*initiated* send TO a recipient), **not** `ChannelMessageService.sendMessage` (inbound-to-agent → would loop). Verified in cross-audit (`chat_id` is the recipient address for email: `EmailImapAdapter` `to: chatId`).

## Verification
- 16 unit tests (family match incl. `emailfoo` guard + two-account tie-break, never-throws, empty-body + missing-recipient guards, subject/thread threading, status-lookup failure). `typecheck` + `prek` green; 278 existing wcore/agent tests pass.
- **Independent cross-audit: SAFE TO SHIP (dormant half), no blockers.** Folded its findings: (1) call the plugin directly so a real SMTP error surfaces instead of PluginManager's null-swallow; (2) empty-body guard; (3) documented + tested the multi-account first-configured-wins tie-break.
- **Live-verify (agent `send_message(email)` → real email delivered) is gated on the Core engine hook landing** — deferred per the engine-first merge order.

## ⚠️ Flag for Core / Overwatch (finding 4, Core-owned)
`host_send_message_request` bypasses the `tool_request` confirmation card — the desktop performs the actual send trusting the engine already gated `send_message`. **Confirm the engine enforces `send_message` approval BEFORE emitting the request** (esp. that Info-category `send_message` isn't silently auto-approved), so an agent turn can't send email to arbitrary recipients with no confirmation. The desktop can't re-gate what the engine approved without a protocol addition; consider a "sent via email" activity/audit surface.

Ref #537. Do not self-merge — Overwatch sequences after the Core hook.

### DIFF
--- a/src/process/agent/wcore/envBuilder.ts
+++ b/src/process/agent/wcore/envBuilder.ts
@@ -688,5 +688,18 @@ export function buildEngineSpawnEnv(opts: {
   //     spawner (standalone CLI, third-party) from an untrusted wire peer.
   out.WAYLAND_ALLOW_WIRE_FORCE = '1';
 
+  // #537: opt the engine into host-delegated `send_message`. The desktop never
+  // writes channel `.toml` into WAYLAND_HOME/channels, so the engine's own
+  // channel table is empty and an agent `send_message` (e.g. to "email") fails
+  // with "unknown channel: email". With this set, the engine keeps the tool
+  // registered but routes the send back to the HOST (a `host_send_message_request`
+  // event), which fulfils it through the desktop's outbound channel plugins — the
+  // same path that already delivers replies — so there is ONE send path and the
+  // engine owns no channel credentials. Always set for the desktop's own trusted
+  // child: when no matching channel is configured the host replies with a clear
+  // error instead of the opaque "unknown channel". Standalone/CLI engines (which
+  // DO hand-author channel toml) never set this and are unaffected.
+  out.WAYLAND_SEND_MESSAGE_HOST_DELEGATE = '1';
+
   return out;
 }
--- a/src/process/agent/wcore/hostSendMessage.ts
+++ b/src/process/agent/wcore/hostSendMessage.ts
@@ -1,0 +1,158 @@
+/**
+ * @license
+ * Copyright 2026 Ferrox Labs
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { IChannelPluginStatus, IUnifiedOutgoingMessage } from '@process/channels/types';
+import type { WCoreEvent } from './protocol';
+
+/**
+ * Desktop half of the #537 host-send-transport hook.
+ *
+ * When the engine is host-delegated it emits `host_send_message_request` for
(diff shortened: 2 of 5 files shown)


FerroxLabs:
**[core→desktop] Two notes from the #141 engine-hook security audit (engine half is built + audited, merging after fixes):**

1. **Your `handleHostSendMessageRequest` has no send timeout** — it awaits `sendViaPlugin` indefinitely and always replies once. The engine bounds its wait at 30s: if your send outlasts that, the engine reports a tool error while the email still delivers, and a model retry becomes a double-send. Engine side is adding do-NOT-auto-retry wording to its timeout error; desktop should bound the send below ~25s (or attach idempotency to the request via call_id dedupe) so the two halves can't disagree about delivery.
2. **Do not offer "Always allow" on send_message approval cards.** The engine is carving send_message out of Always-scope auto-approval regardless (each send re-approves — same rationale as AskUserQuestion: outward-facing, per-target consequences), so an Always button would be a no-op that trains users wrong.

Wire contract confirmed byte-exact against your protocol.ts (field spellings, snake_case, optionals-omitted). Engine PR follows once audit fixes land + CI is green.

PR #558, opened 2026-07-03
fix(update): prefer electron-updater over manual DMG so quitAndInstall works on macOS (#286)

FerroxLabs:
## Problem (#286)

On macOS, auto-update `0.11.x → 0.11.y` **never restarts** and silently re-offers. Root cause is in the renderer, not the ShipIt/Squirrel layer:

`UpdateModal.startDownload()` fired the **manual asset-download** branch first:

```ts
if (updateInfo?.recommendedAsset) { …manual DMG download…; return; }
if (autoUpdateAvailable)          { …electron-updater…;    return; }
```

On macOS a `.dmg` `recommendedAsset` is **always** present (and scores above `.zip`), so `startDownload` always took the manual branch and `return`ed. The electron-updater path — the **only** one that reaches `quit + install + relaunch` (`autoUpdate.download` → `'downloaded'` status event → **Install now** → `quitAndInstall`) — was therefore **dead code on macOS**. The user just got a DMG dropped in `~/Downloads`; the app never installed/relaunched, and the update re-offered on next check.

## Fix

Invert the priority in `startDownload()`:

- **Prefer** `ipcBridge.autoUpdate.download.invoke()` whenever `autoUpdateAvailable` is true → reaches the existing Install-now / `quitAndInstall` flow (~539–558, unchanged).
- Fall back to the manual CDN-rewritten asset **only** when auto-update is unavailable.

The button label logic already renders **“Download and Install”** when `autoUpdateAvailable`, so no UI copy change is needed. The manual `error` view keeps its **“Go to Release”** CTA as a manual escape hatch if an auto-download fails.

## Tests

Adds `tests/unit/UpdateModalDownloadPriority.dom.test.tsx` (jsdom):
- **mac case** (auto-update *and* `.dmg` both present) → routes to `autoUpdate.download`, **never** the manual path. *Verified this test FAILS against pre-fix code and PASSES after.*
- **auto-update unavailable** → manual fallback still fires, forwarding the CDN `url` + release `tagName`.

Typecheck clean; 69 update/auto-update tests green; oxlint clean on changed files (pre-existing scope warnings only).

## Adversarial self-cross-audit

An independent adversarial subagent reviewed the full component (not just the diff). No blocker/major bugs. Two notes, both accepted:
1. There is now **no automatic** fallback from a *failed* auto-download to the manual asset — this is the intended #286 tradeoff (auto-first); the mac error view’s “Go to Release” CTA keeps users unstuck.
2. Auto-path progress UI depends on the main process emitting `autoUpdate.status` `'downloading'` events (pre-existing; not introduced here) — folded into the live-verify ask below.

## ⚠️ Overwatch: needs a PACKAGED Apple-Silicon install/relaunch test

electron-updater is **hard-disabled in dev**, so this cannot be verified locally. Please run on a **packaged arm64 build**: trigger an update, confirm **Download and Install → Install now → the app quits, installs, and relaunches on the new version** (not a DMG in Downloads), and confirm the progress bar advances (finding #2 above). No self-merge.

### DIFF
--- a/src/renderer/components/settings/UpdateModal.tsx
+++ b/src/renderer/components/settings/UpdateModal.tsx
@@ -148,9 +148,21 @@ const UpdateModal: React.FC = () => {
     if (!updateInfo && !autoUpdateAvailable) return;
     setStatus('downloading');
     try {
-      // Prefer the manual path so the URL is the CDN-rewritten asset.url.
-      // Fall back to electron-updater (GitHub) only when the GitHub API manual check failed
-      // but the yml-based auto-update check succeeded - a rare edge case.
+      // Prefer electron-updater whenever auto-update is available: it is the ONLY
+      // path that reaches quit + install + relaunch (via the 'downloaded' status
+      // event -> Install now -> quitAndInstall). The manual asset download only
+      // lands a file in the download folder with no in-place install/relaunch, so
+      // on macOS (where a .dmg recommendedAsset is always present) preferring the
+      // manual path left quitAndInstall as dead code (#286). Fall back to the
+      // manual CDN-rewritten asset ONLY when auto-update is unavailable.
+      if (autoUpdateAvailable) {
+        const res = await ipcBridge.autoUpdate.download.invoke();
+        if (!res?.success) {
+          throw new Error(res?.msg || t('update.downloadStartFailed'));
+        }
+        return;
+      }
+
       if (updateInfo?.recommendedAsset) {
         const asset = updateInfo.recommendedAsset;
         const res = await ipcBridge.update.download.invoke({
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Closing as superseded. #286 (the never-restarts loop this addresses) is verified resolved on the current build: a live 0.11.11 to 0.11.12 auto-update downloaded, installed, and relaunched cleanly on macOS. The DMG-first download branch this PR rewrites is still in shipped 0.11.12 and is working (deliberate: CDN-rewritten asset URL), so merging would change a working path with no reproducing bug. Reopen only if a real update loop reproduces on a current build.

PR #561, opened 2026-07-03
fix(cron): #554 reconcile scheduled-task UI with the store on lifecycle boundaries

### LINKED ISSUE #554: [Bug]: Scheduled task mismatch between UI and backend
### Platform

Windows

### Wayland Version

0.11.11

### Bug Description

There appear to be two separate scheduled-task sync bugs in Wayland Desktop:

1. If a scheduled task is created from chat, the task is created in backend but does not appear in the Scheduled Tasks UI.
2. If a scheduled task is created from the Scheduled Tasks UI, it appears in frontend, but when it is later updated through the chat agent, the backend updates while the frontend/edit dialog continues showing stale data.
This creates a split-brain state between chat-created jobs, UI-created jobs, backend state, and frontend edit state.

### Steps to Reproduce

Path 1 — Create from chat

1. In a chat, ask the agent to create a scheduled task.
2. Confirm the task is created.
3. Verify that the task exists in backend / cron listing.
4. Open the Scheduled Tasks UI.

Path 2 — Create from Scheduled Tasks UI

1. Open Scheduled Tasks.
2. Click New task and create a scheduled task there.
3. Verify that the task appears in the Scheduled Tasks UI.
4. Go back to chat and ask the agent to update that same scheduled task’s prompt or configuration.
5. Verify that the backend reflects the update.
5. Re-open the task in the Scheduled Tasks UI and click Edit.

### Expected Behavior

- A task created from chat should appear in the Scheduled Tasks UI.
- A task created from the Scheduled Tasks UI should remain fully in sync with backend after agent updates.
- The edit dialog should always show the latest saved prompt/configuration from backend.
- Chat-created and UI-created scheduled tasks should use the same source of truth and be equally visible/editable.

### Actual Behavior

Path 1 — Create from chat

- The task is created in backend.
- It does not appear in the Scheduled Tasks UI.

Path 2 — Create from UI

- The task appears in the Scheduled Tasks UI.
- Updating it through the chat agent updates backend state.
- The frontend does not reflect the update correctly.
- The edit dialog can keep showing the old prompt/configuration instead of the newly saved backend version.

### Additional Context

This looks like a desktop-side synchronization / source-of-truth problem between:

chat-created scheduled tasks
UI-created scheduled tasks
backend cron/job state
frontend list/edit state

FerroxLabs:
## What & why

Closes #554. The scheduled-task UI could go stale against the backend even though there is a **single** SQLite `cron_jobs` store (no backend split-brain). Root cause: the renderer trusted one-shot `main→renderer` IPC events (`onJobCreated/Updated/Removed`) with **no authoritative re-fetch fallback**. If a task was created or updated **from chat** while a view was unmounted or its window was blurred, that event was lost and the UI never reconciled.

- **Path 1** (chat-created task missing from the list): fixed by an authoritative store re-fetch when the list regains focus / becomes visible (route-enter is already covered by the page's mount fetch — the route unmounts/remounts).
- **Path 2** (edit dialog shows stale prompt/config after a chat-side update): fixed by fetching the authoritative record when the edit dialog opens, and merging the save onto that record.

## Changes (desktop-only, no engine change)

- **`useAllCronJobs`** — silent re-fetch on window `focus` + document `visibilitychange`. `silent` skips the `loading` flag so a background reconcile never blanks the list to a spinner.
- **`CreateTaskDialog`** — `getJob` on open (edit mode). The form is driven by the authoritative record; `handleSubmit` awaits the in-flight `getJob` and merges onto the store's latest job via a ref (not a stale state closure), so an edit save can't silently revert an agent's chat-side changes. The call is optional-chained (never throws if the bridge lacks `getJob`) and only re-drives the form when the store copy differs from the prop (no redundant re-fill / in-progress-edit clobber).
- **`TaskDetailPage`** — silent re-fetch on focus/visibility so the detail body syncs too.

## Tests

New DOM tests (`--project dom`), 46 pass across the touched suites:
- list focus/visibility re-fetch + **no-spinner** (silent) behavior + listener cleanup;
- dialog `getJob`-on-open: fetch, fresh re-populate, **fresh submit baseline**, null-record fallback.

## Verification

- `bun run typecheck`, `oxlint`, `oxfmt`, and `prek run --from-ref ferrox/main --to-ref HEAD` all pass.
- **Independent adversarial cross-audit** (subagent) run pre-push; it caught a CI-breaking blocker (the initial `getJob` call threw synchronously in existing edit-mode mocks) plus a form-clobber race, a spinner-thrash regression, and a redundant route-enter fetch — **all fixed** in this commit and re-verified.
- **Live-verified** in the running app (CDP :9335, HMR): `/scheduled` renders with real data; firing `focus`/`visibilitychange` keeps the list rendered with **no spinner blank** and no errors; the edit dialog opens **fully populated** via the `getJob`-on-open path.
- End-to-end **create/edit *from chat*** (which needs a real wcore agent turn) is not reproducible in the local harness and is routed to Overwatch for a run-capable environment; the missed-event reconciliation itself is deterministically covered by the unit tests.

### DIFF
--- a/src/renderer/pages/cron/ScheduledTasksPage/CreateTaskDialog.tsx
+++ b/src/renderer/pages/cron/ScheduledTasksPage/CreateTaskDialog.tsx
@@ -5,7 +5,7 @@
  */
 
 import { Bot, ChevronDown, Workflow as WorkflowIcon } from 'lucide-react';
-import React, { useState, useMemo, useEffect, useCallback } from 'react';
+import React, { useState, useMemo, useEffect, useCallback, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Form, Input, Select, Message, TimePicker, Radio, Button } from '@arco-design/web-react';
 import ModalWrapper from '@renderer/components/base/ModalWrapper';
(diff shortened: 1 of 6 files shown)


FerroxLabs:
Overwatch cross-audit (branch-verified): FIX-FIRST before merge. The core fix — re-fetch from the store on focus/route-enter/dialog-open — is sound and confirmed (no loop, no double-fetch, chat-created+chat-edited tasks both reflected). Two real issues to fix:

1) FORM-EDIT OVERWRITE RACE (CreateTaskDialog.tsx). Two effects both call form.setFieldsValue: Effect A fires on open (optimistic editJob), Effect B fires when getJob resolves and re-sets the whole form. If the user opens the dialog and starts typing before getJob returns, Effect B silently overwrites their partial edits. Fix: after getJob resolves, only setFieldsValue for fields that changed AND are not user-dirtied (track a userEdited ref set on any onChange, or skip the second setFieldsValue when the form is already dirty).

2) INVALID TEST FIXTURES (CreateTaskDialog.dom.test.tsx). Fixtures pass state:'active' (string) but ICronJob.state is an object {runCount,retryCount,maxRetries,...}; and assert payload.updates.target.conversationId, but conversationId lives on metadata, not target. The 'as ICronJob' cast hides this, so the test validates a shape production doesn't use. Fix the fixtures to the real ICronJob shape and assert metadata.conversationId (or remove that assertion).

Optional (nice-to-have, not blocking): debounce the focus refetch (skip if last fetch < ~2s) to avoid an IPC+SQLite read on every alt-tab.

Re-request review when 1+2 are fixed.

FerroxLabs (commented):
**Re-requesting review — both fix-first items addressed** (pushed `686d304cb`):

1. **Form-edit overwrite race**: a `userEdited` ref is set on every edit onChange (Form `onValuesChange` + frequency/agent/exec-mode/time/weekday/model/workspace/config); the getJob-resolve populate effect now bails when the form is dirty, preserving in-progress edits. The submit baseline still merges onto the fresh store record via `latestJobRef`. Added a deterministic race unit test (edit mode → deferred getJob returns a newer record → user changes execution mode mid-flight → the fresh re-fill is skipped; the test fails without the guard).
2. **Invalid fixtures**: corrected to the real `ICronJob` shape — `state` object `{runCount,retryCount,maxRetries}`, removed the non-field `lastExecutionTime`, dropped `kind`/`conversationId` from `target`, and the baseline-merge test now asserts `payload.updates.metadata.conversationId`.

Gates: typecheck + oxlint + oxfmt + prek all green; 39 cron DOM tests. Live re-confirmed the edit dialog opens fully populated (no regression). The mid-typing out-of-band-store repro isn't reproducible via the CDP attach harness (no main-eval; allowlist blocks eval-driven cron IPC) so it's covered by the deterministic unit test. Optional focus-refetch debounce not added (non-blocking) — happy to add if you'd like it in this PR.

Issue #559, opened 2026-07-03
Team leader token burn: 77.7M input tok/session, cache_read=0 — enable prompt caching + trim re-billed context (Core/Flux)


FerroxLabs:
## Team leader token burn — measured root cause (from real `wayland.db`)

Sean's #1 concern is the team **leader** burning huge tokens during team execution. I traced it on the real leader conversations (not the brief's assumption). The burn is real and large, but it is **not** "un-truncated teammate tool-results replayed into the leader" — it's the leader's own growing context re-billed **uncached** across many agentic sub-calls.

### Hard numbers — leader `8802dde8` ("Project Manager"), app `cost_events`

| turn | input_tokens | cache_read |
|---|---|---|
| 1 | 61,538 | 0 |
| 3 | 975,251 | 0 |
| 13 | 3,225,775 | 0 |
| 26 | **4,884,523** | 0 |

**Total: 77.7M input tokens for one leader session. `cache_read` = 0 on every turn** (backend `wcore`, model `flux-reasoning`).

### Mechanism (cross-checked against the rows)
1. **No leader ingests large teammate tool-results.** Max `tool_group` row across all 4 real leaders: `846604d1`=0, `8802dde8`=6,698, `d01c0a62`=0, `60902381`=7,049. The 230K WriteFile rows exist only in *teammate* conversations. When the leader read a teammate file it used `limit:80` → bounded.
2. **Zero prompt caching.** `cache_read_tokens = 0` on all 26 turns → every agentic sub-call re-bills the full growing context uncached. This is the dominant multiplier.
3. **Catalog-heavy static prompt re-sent + accumulated.** The leader's 7 largest input rows are 7 copies of the ~44K-char static leader prompt (~31K of it the preset-assistant catalog), re-injected as a new persistent message on each crash/recovery wake and replayed every turn.

### Asks for Core/Flux (ranked by impact)
1. **Enable prompt caching for the leader's stable prefix** (`cache_read=0` is the smoking gun). Biggest lever by far.
2. Investigate the agentic **sub-call count per turn** (turn 26 = 4.88M input ⇒ ~dozens of full-context sub-calls) and whether intermediate context can be trimmed/cached between sub-calls.

### Already shipped desktop-side (helps the base context)
- **PR #557** (catalog slim): shrinks each re-sent static prompt ~31K→7K chars (~7K tok), ×7 re-sends ≈ ~48K tok off the base context, amplified across every uncached sub-call.
[...]

Issue #555, opened 2026-07-03
[Bug]: Teams using OpenRouter API instead of ChatGPT Subscription


cplagz:
### Platform

Windows

### Wayland Version

0.11.9

### Bug Description

Using a team to generate a social media strategy, selected GPT5.4 with ChatGPT sub via waylandcore.  Ran through all the processes and then checked my OpenRouter usage and it has used that instead of the ChatGPT sub.

Screenshots attached.

<img width="2267" height="1324" alt="Image" src="https://github.com/user-attachments/assets/17573fe7-8439-4d22-81f9-9d8fd319cc8e" />

<img width="822" height="321" alt="Image" src="https://github.com/user-attachments/assets/7621e544-ea21-4c06-aa7d-4fef7c533acf" />

<img width="1105" height="327" alt="Image" src="https://github.com/user-attachments/assets/f9d39a69-aea6-4191-8716-3e388c8a7e19" />

### Steps to Reproduce

Run a team, select model as ChatGPT subscription but have openrouter API key also available as models in back end.

### Expected Behavior

Expected to use chatgpt sub

### Actual Behavior

Used openrouter API instead

FerroxLabs:
**Status -> Overwatch** (from desktop): LIVE FINDING (verify-first, as instructed): the picker DOES silently rebind a persisted ChatGPT-subscription selection to a metered look-alike (direct OpenAI API, lists bare gpt-5.4 first) on remount — reproduced from real DB (Dev conv 45fcf7c2) + failing test. FIXED via bridge-tag resolution → PR #565 (area:desktop-ui).

BUT it is a DISPLAY bug, NOT a billing leak (2 independent traces + cross-audit agree): real spend is routed engine-side by hydrateModelForSpawn/mapProvider (envBuilder) from the PERSISTED conversation.model bridge tag (v2:chatgpt-subscription → --provider openai-chatgpt OAuth); WCoreManager reads conversation.model from the DB, and the revalidation rebind is local setCurrentModel that NEVER persists. Real DB confirms 45fcf7c2 keeps tag v2:chatgpt-subscription → still billed subscription. So the money leak does NOT repro via useWCoreModelSelection/resolveSelectedProvider/TeamChatView.

[...]

Issue #207, opened 2026-06-21
[Bug]: Teams model picker reverts a claude member to "Flux Fast"; gemma routes to OpenRouter


FerroxLabs:
Split out from #204 (that issue's headline "Session failed to start / No CLI path" is fixed in #206). These are two distinct, still-open symptoms in the **Teams** model surface.

## 1. Claude member model reverts to "Flux Fast"
**Repro:** add a Team member with backend **claude** (Claude Code via ACP). The model picker shows **Flux Fast**. Pick a native slot (e.g. *Claude Sonnet 4.6*) → the checkmark holds for ~1.5s, then reverts to Flux Fast. (Reported on Windows, v0.11.0.)

**Lead (verified):** `src/renderer/components/agent/acpModelPin.ts:25-26` — `resolvePinnedModelInfo` returns the Flux tier as `currentModelId` **whenever `fluxModelId && showFlux`**, which runs on every claude 1.5s poll (`AcpModelSelector.tsx` ~221-248). `handleSelectModel` (`AcpModelSelector.tsx:296`) *does* clear `selectedFluxModelRef` on a native pick, so the revert is a subtler interaction — likely the Flux pin is being re-derived for the Teams member (the member defaults to a Flux model when Flux is connected, the #129/#160 default-resolution family), or the picked slot id isn't in the polled `availableModels` so it falls through to the agent default. Needs a live trace of the pick → 1.5s poll → `resolvePinnedModelInfo` inputs. This is the **shared** ACP picker (also used in normal conversations) — verify any change doesn't regress the conversation surface.

## 2. "gemma" (gemini member) routes to the OpenRouter API and fails
**Lead (verified):** `src/common/utils/teamModelUtils.ts` (~lines 59-89, gemini branch) merges models from **all enabled providers** into a gemini member's available list without provider-scoping. If OpenRouter is connected and lists a `gemma` id, the model resolves to the OpenRouter provider instead of Google, so inference 401s/fails.

## Not yet covered by tests
No renderer/integration test exercises the Teams pick → poll → revert cycle, nor the gemini provider-scoping. `tests/unit/teamModelUtils.test.ts` covers the util but not the routing collision.

FerroxLabs:
Symptom 2 (gemma -> OpenRouter misroute) fixed in #217: a Gemini teammate's model is now scoped to its Gemini/Google provider instead of being hijacked by the first provider that happens to list the same id.

[...]

Issue #87, opened 2026-06-13
[Bug]:  team-spawned agents do not inherit the working model authentication/session


djmuli:
### Platform

macOS (Intel)

### Wayland Version

0.9.7

### Bug Description

Main Wayland assistant session works with gpt-5.5.

But team-spawned agents do not inherit the working model authentication/session.

When I create a team agent via team_spawn_agent using wcore/gpt-5.5, the teammate asks me to authenticate or fails with provider auth errors, even though the main assistant is actively using gpt-5.5 successfully.

Earlier the spawned agent also sent a sk-flux key to OpenAI and received 401 invalid_api_key, suggesting spawned agents are reading a different/stale provider config from the main chat.

Expected behaviour:
A spawned wcore teammate should use the same configured/working provider auth as the main Wayland session, or clearly indicate which separate credential store it requires.

Actual behaviour:
Main chat works; spawned teammate cannot make model calls.

### Steps to Reproduce

spawn a subagent within a team and ask it to use the same model. 

<img width="1159" height="836" alt="Image" src="https://github.com/user-attachments/assets/031a2511-8811-48f9-95e2-6c973a9a2d19" />

### Expected Behavior

my spawned agent should work on the same model as the main chat

### Actual Behavior

it keeps trying to send flux api keys to open ai and i get 401

<img width="1159" height="836" alt="Image" src="https://github.com/user-attachments/assets/6e4c364b-57f6-4da1-9dbf-6b427fd86f8a" />

FerroxLabs:
Fixed in v0.9.8: a spawned wcore teammate now resolves the provider that owns its pinned model (its key/baseUrl) instead of providers[0], so it no longer sends an sk-flux key to an OpenAI surface. Regression tests added.

PR #570, opened 2026-07-03
fix(team): stop wcore team-spawn billing a subscription model on a metered look-alike (#555)

FerroxLabs:
Fixes the **real billing** half of #555 (the write side Overwatch asked me to confirm). Companion to PR #565 (which fixed only the picker *display*).

## The leak (DB-confirmed)
A team teammate's model is resolved to a provider by **bare model id** — `team_list_models` hands the leader only `gpt-5.4`, with no provider identity. Every resolution path first-matched by `model.config` order, and a ChatGPT-subscription model (`gpt-5.4/5.5`) is **also** listed by the direct OpenAI provider (metered), which sorts first. So the teammate's persisted `conversation.model` was stamped `{...openaiDirect}` with tag `v2:openai`, and `envBuilder.mapProvider` bills off that tag → `--provider openai` (metered) instead of `--provider openai-chatgpt` (OAuth subscription).

**Real DB evidence**: team "Dev Shop" teammates pinned to `gpt-5.5` → persisted tag `v2:openai`. Every shipped launcher spawns teammates with **no** explicit model, so the *unpinned default* path (`resolveDefaultAionrsModel` `providers[0]`, whose first enabled model can be a subscription id — the Openai row's list literally starts with `gpt-5.5`) is the actual trigger.

Unlike #565, this changes the **persisted tag that `envBuilder` reads**, so it actually stops the metered billing.

## Fix (wcore teammates)
- `resolveOwningProviderModelById`: prefer the subscription provider when it owns a **pinned** model id, before first-match (symmetric to the Gemini anti-hijack guard #207).
- `preferSubscriptionForOwnedModel`: a choke point on the **final** resolved model in `buildConversationParams` that re-binds a subscription-owned `useModel` off a metered tag onto `v2:chatgpt-subscription` — covering the **unpinned default** path.

**Scoped to `conversationType === 'wcore'`** (cross-audit finding): only the wcore engine can auth the *keyless* subscription provider (OAuth via `~/.codex/auth.json`). A Gemini-CLI teammate needs an API key, so re-binding it to the keyless subscription row would break bootstrap (`OpenAI API key is required`). Gemini teammates on a subscription model id therefore keep the metered provider — the only route their backend can use. (Trimming subscription-only ids from a gemini teammate's offered model list is a separate concern in `getTeamAvailableModels`.)

No subscription owner / non-owned / already-subscription / disabled / non-wcore → unchanged.

## Verification
- 12 tests: pinned prefers subscription over a metered look-alike listed first; metered still used when no subscription owns the model; choke point re-binds a subscription model resolved onto the metered provider, and leaves non-owned / already-subscription / disabled cases alone. `#87` (pinned ownership) + `#207` (gemini anti-hijack) unregressed. Typecheck + prek green.
- **Two cross-audits**: the first caught that v1 only covered the pinned path (unpinned default was the real trigger) → added the choke point; the second caught that applying it to gemini teammates broke bootstrap → scoped to wcore. Both folded.

## Forensics (Overwatch task 2)
Posted separately on #555: the leader token burn is dominated by the preset catalog (42%, via 7 re-sent copies) → #557 already fixes the biggest slice; teammate tool-results ~0%; the multiplier is `cache_read=0` × ~36 sub-calls → Core #559.

### DIFF
--- a/src/process/team/TeamSessionService.ts
+++ b/src/process/team/TeamSessionService.ts
@@ -18,6 +18,7 @@ import type { AgentBackend } from '@/common/types/acpTypes';
 import type { IProvider, TChatConversation, TProviderWithModel } from '@/common/config/storage';
 import { ProcessConfig } from '@process/utils/initStorage';
 import { getAssistantsDir } from '@process/utils/initStorage';
+import { CHATGPT_SUBSCRIPTION_PROVIDER_ID } from '@process/providers/catalog/chatgptSubscriptionModels';
 import { EventLogger } from './EventLogger';
 import { TaskManager } from './TaskManager';
 import { Watchdog } from './Watchdog';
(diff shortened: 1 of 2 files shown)


FerroxLabs:
### Which spawn path stamped the wrong tag + before/after proof

**Path:** the **unpinned default** resolution in `TeamSessionService.buildConversationParams` → `resolveConversationModel` → `resolveDefaultAionrsModel`, which returns `providers[0]` (first enabled provider in `model.config` order) + its first enabled model. The direct-OpenAI provider (metered, tag `v2:openai`) sorts before the ChatGPT-subscription row **and its model list starts with `gpt-5.5`**, so a teammate defaults onto it. (`team_spawn_agent` is called with no `model` by every shipped launcher, so this path — not the pinned one — is what actually leaked.)

**Persisted `conversation.model` tag:**

| | tag | envBuilder route | billed |
|---|---|---|---|
| **before** (real DB — Dev Shop teammates on `gpt-5.5`) | `v2:openai` | `--provider openai` + `OPENAI_API_KEY` | **metered API** ❌ |
| **after** (this fix, wcore teammate) | `v2:chatgpt-subscription` | `--provider openai-chatgpt` (keyless OAuth) | **subscription** ✅ |

The choke point `preferSubscriptionForOwnedModel` rewrites the resolved model's provider from the metered `v2:openai` row to the `v2:chatgpt-subscription` row (same `useModel`) whenever the subscription owns+enables that id — so the tag `envBuilder.mapProvider` reads flips from `v2:openai` → `v2:chatgpt-subscription`, which is exactly the tag `isChatGptSubscription()` keys on to route `--provider openai-chatgpt`.

DB query to reproduce the "before":
```
sqlite3 wayland.db "SELECT name, json_extract(model,'$.__waylandModelRegistryBridge'), json_extract(model,'$.useModel') FROM conversations WHERE name LIKE 'Dev Shop - %'"
# → v2:openai | gpt-5.5   (should be v2:chatgpt-subscription)
```

Issue #464, opened 2026-06-30
Frictionless 'file a detailed GitHub issue' with auto app-screenshot + diagnostics (Concierge + sidebar) [lane: Overwatch]


FerroxLabs:
**Lane: Overwatch (desktop). area:desktop-ui.**

## Goal
When a user hits a serious bug (or Concierge detects one), let them file a DETAILED GitHub issue in one frictionless step: pre-filled with diagnostics + app/engine versions, and an auto-captured screenshot of the app window ready to paste. Extends the existing GitHub-issue path (SiderFooterQuickActions.ts:29 opens issues/new/choose; "all Wayland support routes through GitHub").

## Decided design (own it)
1. Add an app-window screenshot via Electron `webContents.capturePage()` (main-process/IPC). Chosen over the engine Cua desktop-screenshot because it needs NO OS Screen-Recording permission and captures exactly the Wayland UI. Save the PNG to a temp path AND copy it to the clipboard (nativeImage) so the user pastes it into the issue in one keystroke (GitHub URL params can't attach images).
2. Pre-fill the GitHub new-issue URL (issues/new with title+body, or a template) with: a short title, "what happened", app version, bundled engine version, OS/arch, and the sanitized `wayland_concierge_diag` overview output (already secret-masked). Truncate the body to stay under the URL length limit; if too long, put the diagnostics on the clipboard too and note it.
3. Open it in the browser (openExternalUrl) for the user to review + submit. Guided creation, no GitHub token in the app, no abuse risk.
4. Concierge hook: when the diag/flow surfaces a serious problem, Concierge offers "file a bug report" which triggers this flow (propose -> confirm -> open). Reuse the existing propose/confirm card pattern where it fits.

## Acceptance
- One action from the app (and from a Concierge suggestion) opens a GitHub new-issue page pre-filled with real diagnostics + versions, with the app screenshot already on the clipboard to paste. Verified live in the running app.
- No OS permission prompt for the screenshot. No secrets in the prefilled body (diag output is already sanitized; re-confirm).

## Out of scope (separate tracks)
- Fixing the broken web-browser tool + bundling browser/computer-use as general capabilities (wayland-core#113 + a deeper investigation in progress).
- Direct GitHub-API filing with a token (rejected: token handling + abuse/ToS risk).

Cross-audit + live-verify required before merge.

FerroxLabs:
[...]

PR #588, opened 2026-07-03
feat(voice): wire local voice-model download progress (Whisper + Kokoro)

FerroxLabs:
## What

The local voice-model **Download Model** bar (Settings → Tools → Speech-to-Text / Text-to-Speech) was hardcoded to `0%` with a "Downloading… (progress reporting coming soon)" caption. The download already streams to disk and works — only the progress UI was a stub. Whisper and Kokoro shared the identical bug.

This wires real download progress end-to-end for both.

## How

- **IPC bridge** (`src/common/adapter/ipcBridge.ts`): add a `voiceAsset.downloadProgress` emitter (mirrors the existing `update.downloadProgress` / `buildEmitter` pattern), typed `DownloadProgress`.
- **voiceAssetBridge** (`src/process/bridge/voiceAssetBridge.ts`): pass an `onProgress` callback into `VoiceAssetManager.download()` that re-emits each `{ assetId, bytesDownloaded, totalBytes }` over that channel. `VoiceAssetManager` already fired `onProgress` per chunk — it was just never passed. Cancel keeps working through the existing cancel provider + the manager's internal `AbortController`.
- **Renderer** (`ToolsModalContent.tsx`): hoist a shared `useVoiceAssetDownload` hook used by **both** `WhisperLocalDownloadControl` and `TextToSpeechSettingsSection` (no third copy). It subscribes to the emitter, filters by `assetId`, computes `percent = totalBytes ? min(100, round(bytes/total*100)) : 0`, feeds `<Progress percent={...}/>`, and deletes the coming-soon caption. When the server omits `Content-Length` (`totalBytes === null`) it stays indeterminate at 0 rather than dividing by nothing.

No new i18n keys (the removed caption was an inline default-value fallback, not a locale entry).

## Tests

- New `tests/unit/process/bridge/voiceAssetBridge.test.ts` (TDD): asserts the download provider passes an `onProgress` into `VoiceAssetManager.download` and that it re-emits each `DownloadProgress` over `voiceAsset.downloadProgress`.
- Existing `voiceAssetManager.test.ts` already covers per-chunk progress emission at the manager layer.
- Full gate green: `lint` (0 errors), `typecheck` (clean), `i18n:types` + `check-i18n` pass, and the full suite (`11882 passed / 0 failed`).

Note: no GitHub issue existed for this; titled descriptively.

### DIFF
--- a/src/common/adapter/ipcBridge.ts
+++ b/src/common/adapter/ipcBridge.ts
@@ -49,7 +49,7 @@ import type {
 } from '../types/onboarding';
 import type { ProtocolDetectionRequest, ProtocolDetectionResponse } from '../utils/protocolDetector';
 import type { SpeechToTextRequest, SpeechToTextResult } from '../types/speech';
-import type { DownloadResult, VoiceAsset } from '../types/voiceAsset';
+import type { DownloadProgress, DownloadResult, VoiceAsset } from '../types/voiceAsset';
 import type { SkillSecurityReport, SkillIndexEntry, SkillSource, SkillVerdict } from '../types/skillTypes';
 import type { ImportResult } from '../../process/services/skills/SkillImport';
 import type { KickoffGridResult, KickoffResult, KickoffTelemetryEvent } from '../../process/services/kickoff/types';
@@ -594,6 +594,11 @@ export const imports = {
 export const voiceAsset = {
   download: buildProvider<DownloadResult, VoiceAsset>('voice-asset.download'),
   cancel: buildProvider<{ cancelled: boolean }, { assetId: string }>('voice-asset.cancel'),
+  // Streamed per-chunk download progress. voiceAssetBridge feeds this from the
+  // onProgress callback it hands to VoiceAssetManager.download; the renderer's
+  // download controls subscribe to drive <Progress/> (replaces the old
+  // hardcoded 0% + "progress reporting coming soon" stub).
+  downloadProgress: buildEmitter<DownloadProgress>('voice-asset.download-progress'),
   // Resolve the install state for a known asset. The renderer uses this to
   // suppress the Download button when the model is already on disk (no more
   // "Download Model" alongside an already-installed model).
--- a/src/process/bridge/voiceAssetBridge.ts
+++ b/src/process/bridge/voiceAssetBridge.ts
@@ -18,7 +18,12 @@ export function initVoiceAssetBridge(): void {
     // the inbound asset get filled from the registry; non-empty fields are
     // preserved (test overrides + future extension assets).
     const resolved = resolveVoiceAsset(asset);
-    return VoiceAssetManager.download(resolved);
+    // Stream per-chunk progress to the renderer so the download bar reflects
+    // real bytes instead of a hardcoded 0%. Cancel is already wired through
+    // the cancel provider + VoiceAssetManager's internal AbortController.
+    return VoiceAssetManager.download(resolved, (progress) => {
+      ipcBridge.voiceAsset.downloadProgress.emit(progress);
+    });
   });
   ipcBridge.voiceAsset.cancel.provider(async ({ assetId }) => ({
     cancelled: VoiceAssetManager.cancel(assetId),
--- a/src/renderer/components/settings/SettingsModal/contents/ToolsModalContent.tsx
+++ b/src/renderer/components/settings/SettingsModal/contents/ToolsModalContent.tsx
@@ -21,27 +21,14 @@ import {
   FLUX_RECOMMENDED_IMAGE_ID,
 } from '@/common/config/imageModels';
 import type { VoiceAsset } from '@/common/types/voiceAsset';
-import {
-  Divider,
-  Form,
-  Message,
-  Button,
-  Switch,
-  Input,
-  Slider,
-  Progress,
-} from '@arco-design/web-react';
+import { Divider, Form, Message, Button, Switch, Input, Slider, Progress } from '@arco-design/web-react';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import useConfigModelListWithImage from '@/renderer/hooks/agent/useConfigModelListWithImage';
 import WaylandScrollArea from '@/renderer/components/base/WaylandScrollArea';
 import WaylandSelect from '@/renderer/components/base/WaylandSelect';
 import McpAgentStatusDisplay from '@/renderer/pages/settings/ToolsSettings/McpAgentStatusDisplay';
-import {
-  useMcpServers,
-  useMcpAgentStatus,
-  useMcpOperations,
-} from '@/renderer/hooks/mcp';
+import { useMcpServers, useMcpAgentStatus, useMcpOperations } from '@/renderer/hooks/mcp';
 import classNames from 'classnames';
 import { useNavigate } from 'react-router-dom';
 import { useSettingsViewMode } from '../settingsViewContext';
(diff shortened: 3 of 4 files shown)


FerroxLabs:
**Desktop lane live-verify (@:9334, branch 2eb364650) — OPEN QUESTION ANSWERED: it is TRULY DOWNLOADING, not stalled.**

Drove Settings→Voice→*Whisper (Local)* / *Kokoro (Local)* and watched the on-disk `.tmp` under `<userData>/voice/` (running instance = `Wayland-Dev-2`):

| asset | live `.tmp` growth | final file | source |
|---|---|---|---|
| whisper ggml-**base** | completed too fast to catch mid-stream | **147,951,465 B** `ggml-base.bin` | HF ggerganov/whisper.cpp |
| whisper ggml-**small** | **160MB→212MB @ ~15.7 MB/s**, monotonic | **487,601,967 B** `ggml-small.bin` | HF |
| **kokoro** onnx | **205MB→313MB @ ~40 MB/s**, monotonic | **325,532,387 B** `kokoro-v1.0.onnx` | GitHub releases |

Each streamed into `<dest>.tmp`, then atomically renamed to the final file (log: `[VoiceAssetManager] … downloaded … computed=<sha256>`). The `<Progress/>` bar showed a **real live %** — captured whisper-small at **83%** (aria-valuenow=83) with the Cancel button present. Per-asset isolation is correct: `if (evt.assetId !== assetId) return` (ToolsModalContent L141) so concurrent bars don't cross-talk. Unit tests 2/2 green.

**So: real bytes, real %, real files — the old hardcoded 0% / 'coming soon' stub is genuinely replaced.**

Two minor notes (non-blocking):
1. **No Content-Length → bar sticks at 0%.** `percent = totalBytes ? round(bytes/total*100) : 0` (L142). HF + GitHub-releases both send Content-Length (hence the live %), so shipped assets are fine — but if a mirror ever omits it, the bar reads a frozen 0% while bytes flow. An indeterminate/animated state would read better than 0%.
2. **'Re-download' on an already-installed asset is a no-op re-fetch.** `handleDownload` → `voiceAsset.download.invoke` doesn't pre-delete, so the backend `io.exists(destPath)` short-circuits to `cached:true` (no `.tmp`, no stream). To verify Kokoro live I had to remove the existing `.onnx` first. Consider force-deleting before a 'Re-download', or relabeling.

Test note: I left a freshly-installed `ggml-small.bin` (488MB) in `Wayland-Dev-2` from the growth trace, and re-downloaded the identical `kokoro-v1.0.onnx` I'd removed — both are valid assets in their canonical locations. No self-merge.

FerroxLabs:
**Live-verify PASS (desktop blitz, on a confirmed-fresh engine).** Preflight gate first: bundled + running engine **wayland-core 0.12.21** (Doctor: "Engine reachable — wayland-core 0.12.21 is reachable"), all stale wayland-core overrides `.DISABLED`.

Whisper local-model download (Voice settings → Download): watched the on-disk temp file grow in real time —
`…/voice/whisper/ggml-small.bin.tmp`: 231MB → 242 → 253 → 287 → 407 → 455MB (~+10–25 MB/s, matching real network throughput), then the `.tmp` **atomically renamed** to `ggml-small.bin` (487MB) on sha-verify. So the progress bar reflects REAL streamed bytes, not a hardcoded 0% — confirmed the `.tmp` actually grows. Code path (`VoiceAssetManager.download` streams chunk-by-chunk → `onProgress({bytesDownloaded,totalBytes})` per chunk → `voiceAsset.downloadProgress` emit) + the 2 unit tests corroborate. LGTM (functional verify, not a merge vote).

FerroxLabs:
**#588 → SHIP (fresh re-verify, engine 0.12.21 Doctor-gated).**

Preflight GATE passed first: Wayland Core Overview reads `engine running · v0.12.21` (VERSION card `v0.12.21 wayland-core · pinned`).

Live-verify (Wayland-Dev-2, CDP): deleted `voice/whisper/ggml-small.bin` → UI flipped to **Download Model** → clicked it:
- **`.tmp` genuinely streams**: `ggml-small.bin.tmp` grew 129MB → 231MB in 12s (~8.5 MB/s), per-chunk.
- **Progress bar shows REAL percent**: captured the bar at **74%** mid-download (numeric label + Cancel button), consistent with the `.tmp` byte count (74% × 465MB ≈ 344MB).
- **Atomic finish**: `.tmp` renamed to `ggml-small.bin` (487,601,967 B) and `.tmp` gone. Model restored — no cleanup debt.

Non-blocking observation (hardening): renderer logged `[VoiceAssetManager] whisper-ggml-small downloaded without integrity check; sha256 not pinned (computed=1be3a9b2…)`. The small Whisper model has no pinned sha256, so a corrupt/truncated download would not be caught. Worth pinning the sha in a follow-up; does not block this PR.

FerroxLabs:
**Re-confirmed SHIP on a clean harness restart (engine 0.12.21 Doctor-gated).**

- **#588 (voice download progress):** caught the `.tmp` growing live this time. Deleted the kokoro model, clicked Re-download, sampled `voice/kokoro/kokoro-v1.0.onnx.tmp` at 0.4s: `0 → 35.9M → 74.3M → 112.3M → 150.5M → 191.3M → 235.5M → 273.7M → 312.7M`, then atomic-rename → final `325,532,387 B` (byte-identical). VoiceAssetManager stream-to-`.tmp` + atomic-rename confirmed. (Non-blocking, as before: kokoro logs "downloaded without integrity check; sha256 not pinned".)
- **#599 / #464 (one-click issue report):** sidebar bug button flipped the clipboard from text-only → «class PNGf» (376,794 B fresh page capture) with no temp file; URL byte-budget path unchanged. SHIP holds.

Both re-verified on the running 0.12.21 build. No self-merge.

FerroxLabs:
**Independent cross-audit (desktop lane)** — audited against `gh pr diff 588`. Verdict: **SHIP-WITH-NITS**.

Traced the id path end-to-end and confirmed progress genuinely flows for **both** models: the emitter (`ipcBridge.ts:601`, correctly nested in the `voiceAsset` namespace) re-emits `{assetId,bytesDownloaded,totalBytes}`; `resolveVoiceAsset` spreads `...asset` so the emitted `assetId` equals the renderer descriptor id; renderer `WHISPER_MODEL_ASSETS` (`whisper-ggml-base`/`-small`) and `KOKORO_ASSET` (`kokoro-onnx-model`) ids match the registry keys exactly → no mismatch, no silent 0%-stuck bar. `.tmp`→rename atomicity, SHA gate, and AbortController cancel all live in `VoiceAssetManager.ts` which this PR does **not** touch (`git diff --name-only`) — pre-existing and sound.

**Findings**
- **MEDIUM — renderer percent logic is untested.** The bridge test pins wiring only. The risk-bearing new logic in `useVoiceAssetDownload` (`ToolsModalContent.tsx:142`) — the `totalBytes===null→0` guard, the `===0` div-by-zero avoidance, the `Math.min(100,…)` clamp, and the `evt.assetId!==assetId` cross-asset filter — has zero coverage. All correct on inspection, but the PR's core value is this untested code. Worth a small renderer-hook test.
- **LOW — late progress after cancel.** `handleCancel` (`:166`) sets percent=0 but the subscription stays live (keyed on `[assetId]`). A chunk already past the manager abort check (`VoiceAssetManager.ts:144`) can fire one more `onProgress`, re-setting percent>0. Invisible in practice (idle renders the button branch, not the bar). Cosmetic.
- **LOW/NIT — Re-download no-ops for an installed model.** `download.invoke` short-circuits on `io.exists(destPath)` (`cached:true`, no refetch); bar flashes success at 0%. Pre-existing manager behavior, now reachable via the surfaced control — follow-up, not a blocker.
- **NIT — stale percent on Whisper model switch** — invisible-while-idle; `handleDownload` resets to 0 before next download.

**Verified clean:** emitter placement typecheck-consistent with renderer `.on`/bridge `.emit`; `.on()` returns an unsubscribe fn and the effect cleanup uses it (no listener leak, matches DevSettings/Webui/Channel pattern); no div-by-zero/NaN/>100%; assetId filter prevents Whisper/Kokoro cross-contamination on concurrent downloads; i18n clean (removed `speechToTextDownloadProgressNotReported` was an inline `t(key,default)` fallback with zero orphaned refs across `src/`+`locales/`; percent is numeric); no `any`/unsafe casts in the production diff (confined to test mocks). Recommend ship.

Issue #547, opened 2026-07-02
Autopilot email: inbound-reply loop has no guard (agent auto-replies to its own inbox forever)


FerroxLabs:
Found during #537 live-verify on engine 0.12.20 + #543.

Repro: connect an IMAP/SMTP email channel, Autopilot, have an agent send_message(email) to an address that lands in the same monitored inbox (self-send, or any mail that bounces/replies). The IMAP inbound polls the sent mail back in, feeds it to the agent, which auto-sends another reply, which is polled back in -> unbounded loop. Observed ~1 email per poll cycle to waylandbot@ferroxlabs.com.

Autonomous send in Autopilot is intended (product decision), so this is NOT about gating the send. It is about a loop/rate guard on the INBOUND side: do not re-trigger an agent turn on a message the agent itself just sent (dedupe by Message-ID / self-from), and/or a per-conversation reply rate limit.

Severity: real footgun for any user who points an inbox-watching agent at an address that routes back to itself. Non-blocking for the current ship per Sean, but should land soon.

Related: #537, #543.

FerroxLabs:
Starting this. Concrete root cause + where the guard belongs, for whoever picks it up:

The loop is on the INBOUND side, not the send side. Sequence: agent send_message (host-delegated, #543) sends a mail, it lands in the same inbox the email-imap channel is polling, the inbound poller creates a new incoming message, ChannelMessageService feeds it to the agent as a fresh turn, the agent auto-replies, and round it goes. During #537 testing this fired roughly one mail per poll cycle.

Proposed guard (desktop channels inbound, not engine): (1) drop/ignore inbound messages whose Message-ID or From matches a message this desktop just sent through the same channel (track recent outbound Message-IDs per channel), and/or (2) a per-conversation auto-reply rate limit so a thread can't ping-pong unbounded. Autonomous send in Autopilot stays intended; this only stops the agent replying to its own sends.

FerroxLabs:
**Handoff → core** (from desktop): loop guard for #537 autonomous send. NB root cause is desktop-channels INBOUND (re-feeds the agent its own just-sent mail); fix likely belongs desktop-side: outbound Message-ID dedup on inbound + per-thread auto-reply rate limit. Routing to Core per Sean to own or bounce back to desktop. Full analysis in the issue.

FerroxLabs:
**[core] Update — engine half built, desktop half handed back with a reference design.**

[...]

PR #590, opened 2026-07-03
fix(email-imap): guard against inbound self-reply loop

FerroxLabs:
Addresses #547 (do not auto-close - release closes).

## Problem
The autopilot email channel could get stuck in an unbounded self-reply loop. `EmailImapConnection.send()` SMTP-sends the agent's reply and returns the Message-ID but records nothing. `fetchUnseen()` FETCHes every UNSEEN message and calls `onMessage()` for all of them — including the agent's OWN just-sent mail once IMAP polls it back into INBOX. That echo drives another agent turn → another send → loop.

## Fix (inbound-only)
Desktop mirror of the already-merged wayland-core self-echo guard.

- Track own outbound in `EmailImapConnection`:
  - `selfAddresses` — the inbox address (IMAP/SMTP user, plus each `fromUser` we send as).
  - `sentMessageIds` — a bounded (cap 500, oldest-evicted) set of Message-IDs we returned from `send()`.
- In `fetchUnseen()`, before emitting, `isOwnEcho()` checks the parsed message's From against `selfAddresses` and its Message-ID against `sentMessageIds`. A match is skipped — but still marked `\Seen` so it is not re-fetched forever.
- The send path is **never** gated. Both tracking sets are cleared on `stop()`.

The Message-ID dedup covers alias/relay rewrites where the echoed From no longer equals the inbox address.

## Test
`tests/unit/process/channels/plugins/tier1/email-imap/EmailImapConnection.fetch.test.ts` — new cases assert:
- the agent's own mail (From == inbox) is NOT re-emitted but IS marked seen,
- a message whose Message-ID matches one we just sent is suppressed even from a non-inbox From,
- a genuine third-party inbound reply is delivered unchanged.

## Gate
- Targeted + full suite: 11883 passed, 0 failed.
- Typecheck clean, oxfmt clean, oxlint 0 errors on changed files.

### DIFF
--- a/src/process/channels/plugins/tier1/email-imap/EmailImapConnection.ts
+++ b/src/process/channels/plugins/tier1/email-imap/EmailImapConnection.ts
@@ -284,6 +284,7 @@ export class EmailImapConnection {
         // Suppress the agent's own outbound echoing back into INBOX (#547) - mark
         // it seen so it is not re-fetched, but never hand it to onMessage.
         if (this.isOwnEcho(unified)) {
+          console.debug(`[emailWorker] suppressing own echo uid=${raw.uid}`);
           if (raw.uid > this.lastSeenUid) this.lastSeenUid = raw.uid;
           seenUids.push(raw.uid);
           continue;
--- a/src/process/channels/plugins/tier1/email-imap/EmailImapConnection.ts
+++ b/src/process/channels/plugins/tier1/email-imap/EmailImapConnection.ts
@@ -27,6 +27,17 @@ import {
 const POLL_INTERVAL_MS = 30_000;
 const RECONNECT_BACKOFF_START_MS = 5_000;
 const RECONNECT_BACKOFF_MAX_MS = 60_000;
+/**
+ * Cap on remembered outbound Message-IDs. Bounds memory on a long-lived
+ * connection while comfortably covering the window between a send and IMAP
+ * polling that same message back into INBOX.
+ */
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Trace-verify PASS (desktop blitz).** Traced the self-reply guard in `src/process/channels/plugins/tier1/email-imap/EmailImapConnection.ts` (adversarial subagent + review). Mechanism = dual guard in `isOwnEcho`: **self-From address match** (seeded at `connect()` from imap/smtp user, so it's armed BEFORE any send — order-independent) + **sent-Message-ID dedup set** (bounded 500). Echoes are marked `\Seen` (not re-fetched) and NOT emitted.
- (a) genuine inbound reply from a real correspondent → still emits ✓ (not over-suppressed).
- (b) agent's own echoed mail → suppressed, no re-send ✓. Loop broken.
Both covered by the #547 unit tests. Case-insensitive; missing/forged From safe; multi-account isolated.

**Residual (non-blocking, worth a note in the code):** (1) legit mail sent FROM the account's own address is always dropped (inherent to From-matching — a per-message `X-Wayland-*` outbound header marker would distinguish echo from a human self-send); (2) a simultaneous From-rewrite + Message-ID-rewrite by a relay defeats both branches (low probability). Both are improvements, not corrections — the guard resolves the reported loop.

Live inbox-loop test needs real IMAP+SMTP creds + run-capable env (same blocker as #537) → routed to Overwatch.

Issue #49, opened 2026-06-10
Feature: support drag/drop moves and click preview in project Files


sazora:
# Feature: support drag/drop moves and click preview in project Files

## Summary

Project workspaces have a Files panel, but file management there is still missing two expected file-browser behaviors:

1. Dragging a file or folder onto another folder should move it inside the project workspace.
2. Clicking a file should open the existing preview surface consistently, similar to how project Reference files can be previewed.

## Why this matters

The project Files panel is where users naturally organize generated files, drafts, source material, and working folders. Without drag/drop move support, users have to use the host filesystem or context-menu rename/delete flows for basic organization. Without reliable click preview, the Files panel feels less useful than the References section even though it is the main project workspace file browser.

## Suggested behavior

- File and folder rows should be draggable.
- Folder rows should accept drops.
- Dropping a file/folder onto a folder should move it on disk, then refresh the tree.
- Dropping a folder into itself or one of its descendants should be blocked.
- Moving the workspace root should be blocked.
- Symlink moves/targets should be blocked or handled explicitly.
- Overwrites should not happen silently. Either block with a clear error or show an explicit replace/rename choice.
- Clicking a file should open preview every time, including when the same file was previously selected.
- Context-menu actions like Open, Reveal, Rename, Delete, Download, and Add to Chat should keep working.

## Safety constraints

- Source and destination must stay inside the project workspace.
- The move operation should reject absolute paths outside the workspace.
- The backend should validate paths independently of the UI.
- The tree should refresh after successful moves.
- The UI should show a clear success/error toast.

## Local validation notes

I implemented this locally as a Wayland overlay to validate the workflow:

- Added a guarded `POST /wl-project/file/move` route.
- Patched the shared workspace tree so internal drags use a custom drag payload and folders accept move drops.
- Removed same-file click suppression so clicking a file consistently opens preview.
- Verified with a live temp-file move smoke test.
- Added doctor coverage for the route, served asset marker, drag payload type, and preview-click regression.

This issue is the upstream/source-level version of that behavior request.

sazora:
Follow-up validation detail from local testing:

- The workspace root/parent folder should also be a valid drop target, not only rendered folder rows. In the current project Files tree, the root folder may not appear as a normal row, so users need to be able to drag an item back to the project parent/root by dropping on the Files panel background/header/root area.
[...]

PR #592, opened 2026-07-03
feat(workspace): drag-drop file moves + reliable click preview in project Files

FerroxLabs:
Addresses #49 (do not auto-close — release closes).

## What
Adds internal drag-and-drop **move** and consistent **click-to-preview** to the shared Workspace tree, which drives both project **Files** and the chat **workspace** (one change fixes both surfaces).

## How
**Backend**
- New `move-entry` IPC (`ipcBridge.fs.moveEntry`, `{ sourcePath, targetDir } -> { newPath }`) mirroring `renameEntry`/`removeEntry` in `fsBridge.ts`.
- Both the source and the destination directory are run through `confinePath` (realpath-collapses symlinks, rejects traversal/UNC/sensitive locations) — the existing security boundary is reused, not weakened.
- Moving an entry into itself or a descendant is blocked; a name collision in the target dir is **rejected** (`Target path already exists`) — never a silent overwrite.
- `move-entry` added to `REMOTE_DENIED_KEYS` so remote WS callers cannot reach the new mutation surface.

**Renderer**
- Workspace tree rows are `draggable`; only folder rows accept drops (`allowDrop`). Dropping onto a folder moves the entry on disk, then refreshes the tree with a success/error toast.
- Pure `resolveMoveTarget` helper filters no-ops and self/descendant moves before the IPC call; the main process re-validates independently.
- The OS-file **import** drag handlers now ignore internal node drags (guard on `dataTransfer.types` containing `Files`), so they no longer hijack the tree's own DnD or flash the import overlay.
- File clicks always open preview — including re-clicking an already-selected file (removed the `!wasSelected` suppression).
- All user-facing strings via i18n (`conversation.workspace.contextMenu.moveSuccess/moveFailed`) across all 12 locales.

## Tests
- `tests/unit/bridge/fsBridge.redteam.confinement.test.ts`: `moveEntry` confinement (out-of-root source/target rejected), collision blocked, self/descendant blocked, happy-path move.
- `tests/unit/treeUtils.test.ts`: `resolveMoveTarget` (drop onto folder, non-folder target, no-op into parent, self, descendant, root source, missing data).
- Provider mocks for the new `moveEntry` added to the 6 `initFsBridge` test harnesses.
- Full gate green: `bun run typecheck` clean, `oxlint` 0 errors, `oxfmt --check` clean, `bun run i18n:types` + `check-i18n.js` pass, full `bun run test` = 11892 passed / 0 failed.

## Recommended before merge
Live CDP verification of the drag-move and click-preview interactions in the running app against a real connected profile (per repo norms) — unit tests cover the backend guards and the pure move-resolution logic, not the Arco Tree DnD wiring end-to-end.

### DIFF
--- a/src/common/adapter/bridgeAllowlist.ts
+++ b/src/common/adapter/bridgeAllowlist.ts
@@ -156,6 +156,7 @@ const REMOTE_DENIED_KEYS: ReadonlySet<string> = new Set([
   'write-file',
   'remove-entry',
   'rename-entry',
+  'move-entry',
   'read-file',
   'read-file-buffer',
   'create-temp-file',
--- a/src/common/adapter/ipcBridge.ts
+++ b/src/common/adapter/ipcBridge.ts
@@ -369,6 +369,9 @@ export const fs = {
   >('copy-files-to-workspace'), // Copy files into workspace
   removeEntry: buildProvider<IBridgeResponse, { path: string }>('remove-entry'), // Delete file or folder
   renameEntry: buildProvider<IBridgeResponse<{ newPath: string }>, { path: string; newName: string }>('rename-entry'), // Rename file or folder
+  moveEntry: buildProvider<IBridgeResponse<{ newPath: string }>, { sourcePath: string; targetDir: string }>(
+    'move-entry'
+  ), // Move file or folder into a target directory
   readBuiltinRule: buildProvider<string, { fileName: string }>('read-builtin-rule'), // Read builtin rules file
   readBuiltinSkill: buildProvider<string, { fileName: string }>('read-builtin-skill'), // Read builtin skills file
   // Assistant rule file operations
--- a/src/process/bridge/fsBridge.ts
+++ b/src/process/bridge/fsBridge.ts
@@ -331,6 +331,19 @@ async function destinationExists(candidatePath: string): Promise<boolean> {
   }
 }
 
+/**
+ * Normalize path separators to POSIX `/` (#49 Windows fix). `path.join` emits
+ * backslashes on Windows, but `confinePath` normalizes to `/` internally and
+ * compares against `/`-formed roots — so a freshly joined destination would be
+ * misread as an out-of-root escape, and the string comparisons below (collision
+ * target, self-move, descendant guard) would break. Node's `fs` accepts forward
+ * slashes on Windows, so normalizing here keeps move/rename separator-agnostic
+ * on every platform.
+ */
+function toPosixPath(p: string): string {
+  return p.replace(/\\/g, '/');
+}
+
 const MAX_WORKSPACE_FILES = 20_000;
 
 async function listWorkspaceFilesRecursive(root: string): Promise<IWorkspaceFlatFile[]> {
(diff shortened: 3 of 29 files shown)


FerroxLabs:
### Cross-audit + hardening (desktop lane, overnight blitz)

**Pushed a fix** to this branch — `fix(fsBridge): fail-closed destination check so move/rename never clobbers`:
- The collision guard used `fs.access(dest).catch(() => false)`, which maps **every** rejection to "absent". A present-but-inaccessible destination (EACCES/ELOOP/transient FS error) would slip past the guard and let `fs.rename` overwrite it — a latent violation of the never-clobber invariant that is the whole point of #49. Replaced with an `lstat`-based `destinationExists()` that returns false **only** for ENOENT and re-throws otherwise so the move aborts (fail-closed). Applied to **both** `moveEntry` and `renameEntry`. Added a never-clobber regression test. All 248 fsBridge tests green; prek fully green.

**Live-verify note:** a true headless Arco-Tree drag wasn't feasible (synthetic HTML5-DnD on Arco's Tree is unreliable, and a fresh dev profile has no project/workspace to drag within). I verified the **never-clobber invariant at the bridge layer** (where the data-loss risk lives) via the real handler + the new regression test, and by code review of the collision guard ordering (exists-check strictly before rename). 

**Cross-audit findings to consider (not blocking, author's call):**
- MINOR: `onDrop` ignores Arco's `dropPosition`, so a gap-drop beside a folder is treated as a drop *into* it (no data loss — collision still blocks — but not the drag visual).
- NIT: the collision-specific `msg` from main ('Target path already exists') is discarded; both real errors and name-collisions surface the same generic "Move failed" toast.
- COVERAGE: `isOsFileDrag` (OS-drag vs internal-move routing linchpin), `handleMoveNode` orchestration, and the `onDrop`/`allowDrop` wiring are untested — recommend jsdom tests since the DnD path is the PR's core.

FerroxLabs:
OVERWATCH STATUS — logic verified, SLIPPING from 0.11.13 pending a real UI live-verify.

Independent cross-audit of the dropPosition fix (commit 5d4f8e0a): SHIP-WITH-NITS. The gap-vs-onto resolution is correct on every edge case checked (parent-path slicing for root-level and nested nodes, dropParentIndex guard, same-parent no-op, self/descendant rejection all hold for gap-drops), and the 4 new tests genuinely pin the new behavior against the old ignore-dropPosition bug. 22 unit tests pass, tsc/lint/oxfmt clean.

Why it's held out of 0.11.13:
1. The drag + click-preview UI has never been live-verified. This feature is fundamentally an interaction (HTML5 drag), and a reliable live-verify needs a real drag test in the running app. Overnight CDP drag-simulation is too flaky to trust — a false pass here is worse than a slip, so per the live-verify-must-be-real bar I'm not merging on unit tests alone.
2. MINOR from the audit: allowDrop returns true for gap-drops that resolveMoveTarget then rejects (same-parent no-op, self/descendant), and handleMoveNode returns silently with no feedback — a valid-looking drop does nothing with no toast. Fold a neutral info-toast (or tighten allowDrop) into the same pass.

Next step (Sean or a proper CDP session): live-verify drag-move (onto-folder AND gap/sibling) + click-preview in the running app, add the silent-drop feedback, then merge. The logic work here is done and sound; only the interaction verification remains.

FerroxLabs:
**#592 status: logic remains cross-audited SHIP; the outstanding UI gesture live-verify is BLOCKED by attach-mode harness limits, not by a defect.** Reporting honestly rather than claim a verify I could not cleanly automate.

What is solid:
- The move backend fs.moveEntry({ sourcePath, targetDir }) is present in fsBridge with the self/into-itself/descendant guards, and the onDrop -> handleFileMove -> resolveMoveTarget -> moveWorkspaceEntry -> fs.moveEntry chain is the same logic Overwatch already cross-audited SHIP.

What I could NOT automate on the running 0.12.22 dev app (electromcp CDP-attach), after real attempts:
- The move IPC is not directly invokable in attach mode: app exposes a custom emit/on bridge (no window.ipcRenderer), electron_ipc_invoke_main needs launch mode, and main-process eval is gated. So I cannot drive fs.moveEntry end-to-end from the harness.
- Populating the project Files tree needs a chat bound to a folder (Chat in Folder), which opens a native OS folder dialog; that dialog cannot be driven or stubbed in attach mode.
- Arco Tree's native HTML5 drag is unreliable via CDP synthetic mouse (the known flakiness noted for this ticket). Synthetic drag events would work only on an already-populated tree, which the point above blocks.

Recommended closer (pick one): a launch-mode Playwright e2e that binds a real project workspace and fires Arco Tree onDrop against on-disk fixtures (then asserts the file physically moved), OR a 30-second human drag check in a packaged build. The logic risk is already covered by the cross-audit; only the gesture-to-IPC wiring in the live UI remains to be eyeballed. No self-merge.

PR #593, opened 2026-07-03
refactor(workflow): merge the double right-rail into one Steps|Workspace sider (Addresses #116)

FerroxLabs:
Addresses #116 (clean up the workflow screen). **Does not auto-close** — Sean's call.

## Problem

The workflow screen rendered **two** right rails: WorkflowSurface's fixed **280px non-collapsible** StepRail column PLUS ChatLayout's own **collapsible workspace sider** — squeezing the chat between them.

## Approach (approved)

Merge into **one** collapsible/resizable panel — ChatLayout's single right sider — presented as two Arco tabs: **Steps** (the workflow rail) and **Workspace**. One panel, one collapse, one resize handle.

## How

`useWorkflowSession` holds **local** state per call (not a shared store), and the rail's inputs (`needsInput`, jump/continue handlers) are all derived inside WorkflowSurface — so the rail must stay driven by that single instance. Solution = **portal**:

- **WorkflowSurface** stops rendering its `.right` column; it builds the same rail node and `createPortal`s it into a slot the sider's Steps tab exposes. It keeps its single hook instance + all handlers — only the rail's DOM location changes. The portal sits inside `WorkflowViewModeProvider`, so React context + i18n flow through it. A legacy inline `.right` fallback remains when no sider bridge is present.
- **New `WorkflowRailSlot`** — provider + `WorkflowRailSlotHost` (stable `useCallback` ref registration, guards against render thrash) bridges the sibling subtrees.
- **New `WorkflowTabbedSider`** — Arco `Tabs` "Steps | Workspace" (Workspace omitted when the chat has no workspace); `lazyload={false}` keeps both panes mounted so the portal target survives tab switches.
- **WorkflowStepRail** made fluid (`w-full min-w-0 box-border`) so it fills and tracks the resizable sider instead of a hardcoded 280px that clipped at the default ~220px width; drops its own left border (the sider draws it).
- **All three** workflow panels in `ChatConversation` (wcore, gemini, ACP-family) converted; the sider is forced on so a workspace-less workflow still shows its Steps tab.

## Verification

**Live-verified on a real launched workflow** (CDP :9334): one right sider with **Steps | Workspace** tabs, the full 7-step rail rendering in Steps, **no double rail**, chat not squeezed. DOM-measured the rail fills the sider exactly (220px in a 220px host, 0 overflow) after the fluid-width fix — confirming it tracks the resize handle instead of clipping.

Typecheck / oxlint / oxfmt clean; i18n validated. New jsdom test `WorkflowTabbedSider.dom.test.tsx` (5 cases): both tabs render / Steps-only when no workspace / Steps hosts the rail slot / **portal round-trip** (a marker portals into the slot host) / `useWorkflowRailSlot` default `hasSlotHost=false`.

## Adversarial self-cross-audit

Independent adversarial subagent reviewed the full change. It **cleared** the architecture (portal preserves React context; slot lifecycle safe with the stable ref + `lazyload=false`; `--workflow-accent` confirmed never-set so no regression; `workspacePath=undefined` safe). It caught a **NO-SHIP** blocker — the rail was still hardcoded `w-280px` and clipped in the resizable sider — plus a duplicate border and the un-converted ACP panel. **All three are fixed in this PR** (verified live).

## Known minors (not blocking; flagging for a call)
- The panel header still reads "Workspace" above the Steps|Workspace tabs (minor redundancy).
- **Mobile:** forcing the sider on routes workflows through `MobileWorkspaceOverlay`, which starts collapsed — so on mobile the Steps rail is behind the overlay by default. Worth a product call (e.g. default-open for workflows). Desktop (the primary target) is clean.

No self-merge.

### DIFF
--- a/src/renderer/pages/conversation/components/ChatConversation.tsx
+++ b/src/renderer/pages/conversation/components/ChatConversation.tsx
@@ -308,6 +308,7 @@ const WCoreWorkflowPanel: React.FC<{ conversation: WCoreConversation } & Workflo
         workspacePath={conversation.extra.workspace}
         conversationId={conversation.id}
         hideHeader={true}
+        stepsRailSider={true}
       >
         <WorkflowSurface
           sessionId={workflowSessionId}
@@ -363,6 +364,7 @@ const GeminiWorkflowPanel: React.FC<
         workspacePath={conversation.extra.workspace}
         conversationId={conversation.id}
         hideHeader={true}
+        stepsRailSider={true}
       >
         <WorkflowSurface
           sessionId={workflowSessionId}
@@ -626,6 +628,7 @@ const ChatConversation: React.FC<{
           workspacePath={conversation?.extra?.workspace}
           conversationId={conversation?.id}
           hideHeader={true}
+          stepsRailSider={true}
         >
           <WorkflowSurface
             sessionId={workflowSessionId}
--- a/src/renderer/pages/conversation/components/ChatConversation.tsx
+++ b/src/renderer/pages/conversation/components/ChatConversation.tsx
@@ -37,6 +37,8 @@ import { usePreviewContext } from '../Preview';
 import StarOfficeMonitorCard from '../platforms/openclaw/StarOfficeMonitorCard.tsx';
 import ConversationSkillsIndicator from './ConversationSkillsIndicator';
 import { WorkflowSurface } from '@/renderer/pages/guid/components/workflow/WorkflowSurface';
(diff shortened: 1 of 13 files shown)


FerroxLabs:
**Fresh live-verify (desktop lane, engine 0.12.21).** Stood up this branch on its own instance and opened a workflow conversation. Right rail is **expanded by default** (222px, not hidden), a **single merged sider** (`siderCount=1`) with **Steps + Workspace** tabs, Steps active showing the step list. The tip-commit regression fix ("default the Steps rail sider expanded so it is not hidden") holds live. CI fully green. No self-merge.

Issue #458, opened 2026-06-30
Wire Wayland Doctor into Concierge + add workspace/config-path checks [lane: Overwatch]


FerroxLabs:
**Lane: Overwatch (desktop). area:desktop-ui.** Today's headline task.

## Problem
Wayland Doctor (`src/process/doctor/`) is solid — 9 checks (providers, models, engine reachable/routing, MCP, backends, workspace drift, secret storage, engine config), concurrent runner with per-check timeout, clean JSON `DoctorReport`. But it is **IPC + Settings-page only**. No agent access, no MCP tool, no slash command. Concierge — the in-app helper whose entire job is "diagnose the user's setup" — literally cannot call it. And the checks miss the two bugs we just confirmed (temp-workspace fallback; the dual config-path the customer hit).

## Goal
Make Doctor callable by Concierge, and add the checks that catch the real failures, so a user reporting "nothing works / can't find config / channel won't connect" gets a real diagnosis in plain English.

## Scope
1. Refactor Doctor core into a process-agnostic callable returning `DoctorReport` (already DI'd/pure except `registry.ts`, which binds main singletons). One entry point any surface can call.
2. Expose it to Concierge as an IN-MAIN MCP tool (e.g. `wayland_doctor`) that returns the `DoctorReport` JSON. Must run in-main (needs `agentRegistry`, `ProviderRepository`, `mcpService`, db) — NOT a stdio subprocess.
3. New checks:
   - `workspace.configured`: flag any project/conversation running in a temp/throwaway workspace instead of a persistent one; report the real path. (Pairs with #455 — coordinate the persistent-workspace concept.)
   - `config.paths`: report the resolved app config dir (`getConfigPath` -> .../Wayland/config) AND engine config dir (`nativeConfigDir`) so the "two config paths" confusion is visible.
4. Surface: add a `/doctor` slash command; add the already-defined `doctor.copy-text` channel to `bridgeAllowlist.ts` (currently missing).
5. Concierge prompt (`concierge.md`): teach it to run `wayland_doctor` when a user reports trouble and explain results plainly.

## Constraints
- Build against `ferrox/main` (local tree is the stale `feat/jsonstream-uiux`).
- Cross-audit (non-author) + real tests + live-verify in the running app before merge.

## Acceptance
[...]

PR #596, opened 2026-07-03
feat(doctor): workspace/config checks + /doctor command + deny copy-text (#458)

FerroxLabs:
Addresses #458 (do not auto-close — release closes).

Wires Wayland Doctor into a user-reachable surface and adds the two checks that catch the real setup failures the customer hit, so someone reporting "nothing works / can't find my config / channel won't connect" gets a plain-English diagnosis instead of a buried Settings page.

Scoped to the sure slice from the approved plan — this does NOT add the in-main live-probe MCP tool (separate spike; `wayland_concierge_diag` already covers the on-disk facts).

## What + how
- **New checks** (pure, DI'd, unit-tested):
  - `workspace.configured` (`checks/workspaceChecks.ts`) — flags any project/chat running in a throwaway temp/default workspace instead of a persistent folder, and reports the real path. Reuses the concierge-diag temp detection (`<kind>-temp-<unix-ms>` with a >=10-digit timestamp so user folders like `client-temp-2024` are not misread, OS-tmpdir paths, and the authoritative `customWorkspace === false` flag). WARN, output-bounded.
  - `config.paths` (`checks/configChecks.ts`) — surfaces the two distinct config directories (app config via `getConfigPath` vs the separate wayland-core engine config via `nativeConfigDir`) to defuse the "two config paths" confusion. Informational PASS.
  - Both bound in `doctor/registry.ts`.
- **`/doctor` slash command** — a builtin command in the chat composer that runs the full battery via `doctor.runDoctor` and renders the report inline in a modal (`DoctorReportModal.tsx`), reusing `DoctorSettings/reportText.ts` and the Windows-hardened clipboard copy.
- **Security** — deny `doctor.copy-text` to remote WS callers in `bridgeAllowlist.ts`. It writes caller-supplied text to the host OS clipboard from main (a clipboard-injection primitive), so it belongs in the remote denylist alongside `doctor.run`. Added a red-team test.
- **concierge.md** — points users at `/doctor` for a self-serve, copyable health report.

## Tests
- New unit tests for `checkWorkspaceConfigured` / `isTempWorkspacePath` (incl. the `client-temp-2024` false-positive guard) and `checkConfigPaths` in `tests/unit/process/doctor/checks.test.ts`.
- New red-team test asserting `doctor.run` + `doctor.copy-text` are denied for remote callers.
- Full gate on a clean `ferrox/main` base: `typecheck` clean, doctor/red-team/reportText suites green (103), `check-i18n` passes, lint 0 errors.

## Note
User-facing UI change (the `/doctor` modal) — needs a live-verify pass in the running app before merge.


FerroxLabs:
### Cross-audit + live-verify (desktop lane, overnight blitz)

**Live-verify PASS.** Ran `/doctor` from a chat send box in a running dev build. The command surfaced in the slash menu ("Run a health check on your setup"), selecting it opened the **Doctor** modal, and the report ran and rendered fully: Overall FAIL, Summary 9 pass / 1 warn / 1 fail, per-check `[PASS]/[WARN]/[FAIL]` lines with descriptions + Fix hints, and Copy report / Close. **Both new checks render:** `[PASS] Workspace paths` (`workspace.drift`) and `[WARN] Workspace configured` (`workspace.configured`). Screenshot captured.

**Cross-audit findings (author's call):**
- MINOR: a *failed* doctor run leaves the modal body as an empty `<pre>` — no error text or retry. Worth a fallback message (my run succeeded, so this path wasn't exercised live).
- MINOR: "Copy report" uses the renderer clipboard helper, not the main-process `doctor.copyText` bridge the Doctor **settings** page routes through — inconsistent copy paths for the same report.
- NIT: `workspace.configured` enumerates every conversation ever created, so a single historical chat left on the default (non-custom) workspace flips the check to WARN — which is exactly what I saw. Confirm that's the intended signal vs. noisy.
- COVERAGE: the 96-line `DoctorReportModal` and the `/doctor` slash dispatch are untested despite existing DOM-test + sendbox-test patterns; `isTempWorkspacePath`'s `[a-z]+-temp-` regex and the substring tmpdir check have edge cases worth a unit test.

FerroxLabs:
**Cross-audit vs the real `gh pr diff` (multi-agent, 6 finders × independent verify → 2 CONFIRMED, security dimension found nothing — the deny + sanitization + redteam denylist test are sound).**

1. **MEDIUM (test-coverage):** `src/renderer/components/chat/DoctorReportModal.tsx` + the `/doctor` slash-command wiring in `sendbox.tsx` ship with **zero test coverage**, unlike the sibling `DoctorSettings` (which has `doctorSettingsCopy.dom.test.tsx`). The modal has non-trivial behavior — run-on-open `useEffect` (resets report→null then `run()`), `copyReport` via the shared `copyText` util, `disabled={!reportText||running}`. A refactor that e.g. calls `ipcBridge.doctor.copyText` directly or drops the run-on-open effect would pass CI silently. Suggest a dom test mirroring the sibling: opening invokes `runDoctor` once, renders `buildDoctorReportText`, Copy routes through `copyText`, rejected run shows error + leaves Copy disabled.

2. **LOW (correctness):** `/doctor` is pushed into the slash-command list **unconditionally**, but `doctor.run` is in `REMOTE_DENIED_KEYS`. In a paired-WebUI (remote WS) session, selecting `/doctor` opens the modal, whose `runDoctor.invoke()` is dropped by `isAllowedForRemote('subscribe-doctor.run')=false` → the modal shows "Could not run the Doctor", report stays null, Copy is permanently disabled = a dead affordance offered to WebUI users. Suggest gating the command on a local (non-remote) context, or a clear "desktop-only" message.

Both trace-verified against the actual diff. Non-blocking, but worth addressing before merge.

Issue #281, opened 2026-06-23
[Core] Emit parent_agent_id + depth on SubAgentEvent so desktop can render a sub-agent tree (for #252)


FerroxLabs:
**Engine dependency for desktop feature #252** ("show what my agents are doing" — sub-agent tree + per-agent live monologue). Filed by desktop after verifying the split against current engine source (commit ~9b6bea23).

## What desktop needs from core
The desktop can already render a **flat list** of sub-agents (it consumes `ProtocolEvent::SubAgentEvent`), and the per-agent **thinking monologue is already relayed** (`crates/wcore-agent/src/agents/channel_sink.rs:85` emits `Thinking` through the relay). The one thing the desktop **cannot** build from today's stream is the **nested tree hierarchy**, because the event carries no parent/depth linkage:

`crates/wcore-protocol/src/events.rs:131-135`
```rust
SubAgentEvent {
    parent_call_id: String,
    agent_name: String,
    inner: Value,
}
```
`parent_call_id` is *call-based* (which Spawn invoked it), not *agent-based* (which agent spawned it), so a sub-agent that spawns its own sub-agent can't be placed in the tree.

## The contract (small, backward-compatible)
Add two optional fields and thread parent context through Spawn → ChannelSink → drain:
```rust
SubAgentEvent {
    parent_call_id: String,
    agent_name: String,
    inner: Value,
    #[serde(default, skip_serializing_if = "Option::is_none")]
    parent_agent_id: Option<String>,  // the spawning agent's id (None = top-level)
    #[serde(default)]
    depth: u32,                       // 0 = top-level, 1 = spawned by top-level, ...
}
```
- Wire points: `crates/wcore-agent/src/spawn_tool.rs:215-299` (pass parent id + depth+1 into the spawn), `channel_sink.rs` (carry them on the relay), drain task (emit them on the SubAgentEvent).
- `#[serde(default)]` keeps it byte-compatible for older hosts. Gate behind the existing sub-agent-trace capability flag if one exists.
- Est. ~50 LOC, no algorithmic change.

## Desktop side (no core change, can ship in parallel)
- Surface the already-relayed sub-agent `thinking` (currently dropped at `app/src/common/chat/chatLib.ts:672`, which only reads `text_delta`) → renders the per-agent monologue.
- Build the tree once `parent_agent_id`/`depth` land.

Tracking issue (desktop): #252.

FerroxLabs:
[...]

Issue #252, opened 2026-06-22
[Feature]: More info about working agent


GenieRobot:
### Feature Description

Detailed output of the agent's working processes

### Problem Statement

I'd like to get more information on what an agent is doing right now. Currently, I can only see that it has started a process, and that the process is running. No info on what it's doing.

<img width="1216" height="630" alt="Image" src="https://github.com/user-attachments/assets/6da601a5-c442-4422-8166-9edca15e989e" />

### Proposed Solution

Instead of just showing a spinner on the left chat list item, show the agents as sub-processes (tree view), and in the chat view as accordeon, and on click, switch to that agent's detailed thought monologue. I want to know what my agents are doing.

### Feature Category

UI/UX Improvement

FerroxLabs:
**Routed → desktop (enhancement).** Surface what each agent is doing in real time — sub-process tree view + per-agent thought monologue instead of a bare spinner. Backlog UX enhancement; not a launch blocker.

FerroxLabs:
**Verified split (this stays desktop; one piece goes to core).** Checked against current engine source:
- **Per-agent monologue (part 2) + accordion (part 3) = desktop-only.** The engine already relays sub-agent **thinking** (`channel_sink.rs:85`); the desktop just **drops it** today — `src/common/chat/chatLib.ts:672` keeps only `text_delta` from the sub-agent inner event. Capture the inner `thinking` and render it in `SubAgentActivityCard` + the chat-item accordion. No engine change.
- **Tree view (part 1) = needs core.** `SubAgentEvent` has no parent/depth linkage, so only a flat sibling list is renderable. Tracked as the engine contract in #281 (add `parent_agent_id` + `depth`). Desktop can build the renderer in parallel and wire it when #281 lands.

So #252 stays desktop-owned; #281 is the core dependency.

FerroxLabs:
**Status update — 2026-06-25, desktop lane (branch `feat/jsonstream-uiux`)**

Progress-acknowledgment / "working agent" chrome reworked, research-backed (workflow `wf_5656ff55-267`: Buell & Norton labor illusion, NN/g latency thresholds, GPT-4o sycophancy rollback, arXiv 2402.03303). Done + gated (tsc 0 / oxlint 0 / i18n / 23 humanizer tests):

[...]

PR #602, opened 2026-07-03
feat(sidebar): live agent activity label + sub-agent tree (#252)

FerroxLabs:
Addresses #252 (do not auto-close — release closes).

## What
Replaces the bare spinner shown on a generating conversation in the history sidebar with a live status line that tells you **what the agent is actually doing**:

- The current humanized action ("Searching the web", "Reading config.ts", "Reasoning", "Running a command").
- When the turn spawned sub-agents, an expandable caret + count badge that reveals each sub-agent and its own current step (the "expandable sub-agent tree" from the issue).

This fixes the literal complaint (part 1 of #252): the sidebar row now surfaces real progress instead of an opaque spinner. The in-chat ActivityTimeline / OrbitThinking chrome already shipped; this brings the same signal to the background rows in the sidebar.

## How
- **Data layer (`src/common/chat/activity/conversationActivity.ts`)** — a pure `foldConversationActivity(prev, event)` that folds each raw engine stream event into a tiny per-conversation snapshot `{ label, glyph, agents[] }`. It **reuses the existing projectors** (`transformMessage` + `activityStep` / `projectMessages` / `activityLabels`) rather than re-deriving anything, and returns the same reference when nothing meaningful changed so streamed tokens don't churn.
- **Store (`useConversationListSync`)** — folds the snapshot from the **same `responseStream`** the sidebar already listens to for the generating state, so background rows read status **without mounting the message list**. Cleared on terminal / turn-completed / delete. Passive and read-only.
- **UI** — `ConversationRowActivity.tsx` renders the status line + expandable tree (Arco + @icon-park, semantic tokens); `ConversationRow` wraps its row in a card so the line sits under the title only while generating. Plumbed through `types.ts` → `useConversations` → `index.tsx`.
- i18n: two new keys under `conversation.sidebarActivity` (en-US), types regenerated.

Scope note: the tree renders the flat sibling list of sub-agents the engine exposes today; nested parent/depth linkage is the tracked Core dependency (#281) and is not required here.

## Tests
- `tests/unit/conversationActivity.test.ts` — pure fold: tool/thinking/sub-agent projection, multi-agent tracking, in-place update, dedupe, pass-through.
- `tests/unit/renderer/conversationRowActivity.dom.test.tsx` — label render, "working" fallback, caret expand/collapse reveals sub-agents.
- Full gate green locally: full vitest suite (11891 passed / 0 failed), `tsc --noEmit` clean, oxlint clean on changed files, i18n validation passes.

UI feature — needs a live-verify pass in the running app (spawn a multi-agent turn, confirm the sidebar label updates and the sub-agent tree expands) before merge.


FerroxLabs:
**Desktop lane live-verify (@:9334, branch d4bf796bf)**

**Passivity — CONFIRMED (code-airtight).** The activity data path adds *zero* session-mutating IPC. `useConversationListSync` folds activity into the **already-existing** `ipcBridge.conversation.responseStream.on` handler (the same stream that drove the old spinner) — no new subscription — plus one read-only `database.getUserConversations` query. Diff scan for `.start/.resume/.send/sendMessage/createConversation` on added lines: empty. `ConversationActivityStatus` is pure-presentational and `stopPropagation`s its caret so expanding the tree can't even open the conversation. It cannot start a session.

**Render path — VERIFIED.** Unit `conversationActivity.test.ts` (fold incl sub-agent tree) + DOM `conversationRowActivity.dom.test.tsx` (label + expandable agents) pass (11/11). Live build wires it: `ConversationRow` L361 renders `<ConversationActivityStatus>` gated on `isGenerating && activity has label/agents`, reading the passive store — plumbing connected in the running build.

**Nuance (cosmetic, not a blocker):** the 16px leading `<Spin>` glyph (ConversationRow L167) still shows during generation; the rich activity label + expandable sub-agent tree is added as a **sub-line** (L361). So the info-free spinner is *superseded by* informative content rather than literally removed — matches #252 intent.

**Not done here:** a real-engine *visual* live-render (watching the label update through an actual generating turn) — this fresh profile has no model/send affordance configured, so a real turn is engine-gated. Routing that visual confirm to Overwatch per the established pattern. No self-merge.

FerroxLabs:
**Independent adversarial cross-audit (2nd pass, against the real `ferrox/feat/agent-activity-tree` diff) — VERDICT: no blockers, no majors. Passivity verified clean.**

- **Passivity (critical):** full-diff sweep for `ipcBridge`/`.invoke`/`.send`/`.emit`/`startSession`/`resume`/`sendMessage`/`createConversation` → **zero** new session-mutating calls (only two `import type` hits). `useConversationListSync` added **no new subscriptions** — the `responseStream`/`turnCompleted`/`listChanged` handlers + read-only `getUserConversations` were all pre-existing; the PR only injected `applyActivityEvent`/`clearActivity` into those bodies. `foldConversationActivity` + projectors are pure. Caret `stopPropagation`(+`preventDefault`) correctly blocks row-open. It cannot start a session.
- **Correctness:** ref-equality dedupe holds (`withLabel`/`sub_agent` return `prev` unchanged; store skips emit on `next===prev`); no per-token re-render storm (content/thought aren't activity events; `markGenerating` short-circuits); terminal cleanup wired on all three exits, keyed by conversation_id; module-scope subscription idempotent via `isStoreInitialized`; glyph covers full status union; i18n keys present.

**3 minor (non-blocking) notes for a follow-up, not this PR:**
1. `applyActivityEvent` runs `transformMessage` unguarded inside the shared `responseStream` handler — a malformed `thinking`/`tool_group` event with `data==null` would throw; the platform emitter aborts that emit loop (no app crash, but co-listeners could miss that one event). A `try/catch` around the fold closes it. The PR newly extends the transform to *every background conversation's* stream, so worth hardening.
2. `upsertAgent` MAX_AGENTS(8) eviction is insertion-order — a >8 fan-out can drop a still-`running` agent for a newer `done` one (cosmetic/bounded).
3. Post-terminal out-of-order activity event can re-create a snapshot after `clearActivity`; never renders (gated on `isGenerating`), cleared next terminal — invisible/bounded.

Core claims all check out. Combined with the earlier desktop live-verify: recommend merge; items 1–3 are a small follow-up. No self-merge.

Issue #520, opened 2026-07-01
[Bug]: HELP! Persistant issue since 0.10. Unable to use Wayland at all since 0.11.2


frankforges:
### Platform

Windows

### Wayland Version

0.11.9

### Bug Description

wcore not being able to execute commands, either DS nor ChatGPT OAuth. Ongoing issue for me since 0.10 I think

### Steps to Reproduce

Fresh reinstall, add DS key and/or ChatGPT OAuth, start a new chat and ask it to run any command

### Expected Behavior

Should run

### Actual Behavior

Not running

### Additional Context

Having a lot of isssues scattered among releases and resinstalls... since 0.11.3 I was unable to continue working on Wayland because the agent that managed to survived from 0.11.2, Claude, also failed to spawn start a new session (bun related problem, unable to start, that was after 0.11.2 don't remember the exact release). doctor haven't been helful since ever.

### What I've done so far:
Today I decided to try a fresh install (losing all wayland memories and chats, totally resigned and decided) of the new 0.11.9. I performed a powerful unistall with Iobit Unistaller of by previous crashe-bugged release (updated 0.11.9, failing to start a new Claude session, and wcore not running commands), unistalled bun, reboot my pc, checked if there was any leftover, installed through PS admin the latest bun, rebooted again, installed latest 0.11.9, and and again, same wcore bug and ClaudeCode, although it works, is taking forever to respond (succesfully running commands, but sluggish)

I then reinstalled again, thought about testing 0.11.2 again and, same problem... Aditionally, releasaes after 0.11.2 **hide command execution info** and this is **absolutuely critical** for me, I need, i believe we all need, to see what the actual running command is (now I remember that after 0.11.2 that indicator vanished completely, and then returned shyly in like 0.11.5 i think, just showing agent is running a command but not the actual command!), this can be checked in following screenshots!

I really want to believe in Wayland and continue using it... some help here please? I thought about postitng this in the fb group, but to not poison community's faith, I thought I may first report this here

### 11.9

wcore DS key
<img width="3190" height="1896" alt="Image" src="https://github.com/user-attachments/assets/c5238286-35eb-4b83-a7df-c32207cb90de" />
command info hidden!!!
<img width="3198" height="1894" alt="Image" src="https://github.com/user-attachments/assets/2802e8b9-dfad-4503-8071-49aa99b74942" />

### 11.2

wcore DS key
[...]

Issue #151, opened 2026-06-16
WebUI should canonicalize mixed path/hash routes


sazora:
## Summary

Wayland WebUI uses `HashRouter`, so the active in-app route comes from `window.location.hash`.
If the browser lands on a mixed URL such as:

```text
/assistants#/guid
```

the visible path suggests the user is on Assistants, but the app still resolves the active route as `#/guid`.
On mobile this can look like the app "cycles back" to the previous page after tapping a menu/deep link.

## Why It Matters

This is easy to trigger from:

- mobile browser history restoring stale hashes
- external/deep links that use path-style app routes
- custom overlays or integrations that call `location.assign('/assistants')`
- copied URLs where the path and hash disagree

Because `HashRouter` treats only the hash as canonical, app-route paths such as `/assistants`, `/projects`, `/teams`, etc. should either be normalized to hash routes or avoided by navigation helpers.

## Reproduction Shape

1. Open WebUI on a route like `/#/guid`.
2. Navigate via a path-style URL such as `/assistants` without clearing/replacing the hash.
3. Browser URL can become `/assistants#/guid`.
4. The app keeps rendering GUID because `#/guid` remains the active route.

## Suggested Fix

Add a tiny startup canonicalization guard before/around router initialization:

```ts
const appRouteSegments = new Set([
  'assistants',
  'conversation',
  'conversations',
  'guid',
  'memory',
  'mission-control',
  'project',
  'projects',
  'scheduled',
  'settings',
  'team',
  'teams',
  'wiki',
  'workflows',
]);

const path = window.location.pathname || '/';
const hash = window.location.hash || '';
const cleanPath = path.replace(/^\/+/, '').replace(/\/+$/, '');
const segment = cleanPath.split(/[/?#]/)[0] || '';

if (
  path !== '/' &&
  path !== '/login' &&
  appRouteSegments.has(segment) &&
  (!hash || hash === '#' || hash === '#/' || hash === '#/guid')
) {
  window.history.replaceState(null, '', `/#/${cleanPath}`);
}
```

Also prefer a shared navigation helper for WebUI routes:

```ts
window.location.hash = '#/assistants';
```

instead of:

```ts
window.location.assign('/assistants');
```

## Expected Behavior

Opening `/assistants#/guid` should canonicalize to `/#/assistants` or otherwise render Assistants, not GUID.

FerroxLabs:
[...]

PR #604, opened 2026-07-03
fix(wcore): surface the actual shell command in the tool timeline (Addresses #520)

FerroxLabs:
Addresses #520 (desktop half — command visibility).

## The bug
Since 0.11.2 a wcore tool card shows a generic **"Running a command"** instead of the actual command — the visibility the reporter (and others) depend on to see what the agent is executing.

## Root cause (two desktop gaps, both required)
1. **Mapper clobber.** `src/process/agent/wcore/index.ts` hardcoded `description: ''` on the `tool_running` / `tool_result` tool_group frames. The renderer merges tool_group items by `callId` with a plain `{...tool, ...incoming}` spread (`chatLib.ts` `composeMessage`), so the empty string wiped the command the preceding `tool_request` carried in `description`.
2. **Render path drops it.** The wcore turn actually renders via the **#252 ActivityTimeline**, not `MessageToolGroup`. `toolGroupToNodes` never lifted the command onto the `ActivityNode`, and `deriveStep` humanized every shell tool to a static `"Running a command"` — so even an intact command never reached the label.

Core's shipped #520 work (PR #151) is the Windows AppContainer output-drain fix; `ToolRunning` carries no args on the wire — the command reaches the host via the `tool_request` `ToolInfo.description` / exec `confirmationDetails.command` (both preserved through the merge). So this is a pure desktop fix; no protocol change.

## The fix
- **Mapper:** cache the request-time `description` per `callId` and re-attach it to the running/result frames (stops the clobber; helps non-exec tools too). Cleared on terminal.
- **Projection:** `toolGroupToNodes` lifts the exec command (`confirmationDetails.command`, falling back to the item `description`) onto a new `ActivityNode.command`.
- **Label:** `deriveStep` renders the command for command-glyph tools — `"Running echo hello"`, whitespace-collapsed and capped to one legible line; the full text stays in the expandable detail. Non-command glyphs keep their richer file/host/query labels.

## Verification
- Unit tests: `toolCommandVisibility.test.ts` (mapper, 4), `projectMessages.test.ts` (command lift + label, +4), `activityLabels.test.ts` (`deriveStep` command surfacing incl. cap / `Execute:`-strip / no-hijack, +5). Full activity + tool-group suites green; `typecheck` clean.
- **Live-verified** @ CDP :9334 on **wayland-core 0.12.21** (Doctor-gated): a real DeepSeek-V4-Flash / Wayland-Core turn running `echo WL520_LIVE_CHECK` now shows the tool step **"Running echo WL520_LIVE_CHECK"** (was "Running a command"), including after reopening the conversation from the DB.

Touches the #252 observability humanization layer (`projectMessages` / `activityLabels` / `ActivityNode`) — low overlap with the in-flight #602 sidebar-tree PR (different files). No self-merge.

### DIFF
--- a/src/common/chat/activity/activityLabels.ts
+++ b/src/common/chat/activity/activityLabels.ts
@@ -53,6 +53,24 @@ const firstPath = (s: string): string => {
   return m ? m[0] : '';
 };
 
+/**
+ * #520 - a one-line, capped rendering of a raw command for the timeline label.
+ * Collapses whitespace/newlines (a heredoc must not blow up the row) and caps
+ * length so a long command stays a legible single line; the full text remains
+ * available in the expandable detail.
+ */
+const CMD_LABEL_CAP = 64;
+export const formatCommandLabel = (command: string): string => {
+  // Drop a leading "Execute: " the wcore mapper prefixes onto the description
+  // fallback, so we show the bare command either way.
+  const bare = command
+    .replace(/^execute:\s*/i, '')
+    .replace(/\s+/g, ' ')
+    .trim();
+  const capped = bare.length > CMD_LABEL_CAP ? `${bare.slice(0, CMD_LABEL_CAP - 1)}…` : bare;
+  return `Running ${capped}`;
+};
+
 type LabelRule = {
   /** matches against the lowercased "name + ' ' + detail" haystack. */
   test: RegExp;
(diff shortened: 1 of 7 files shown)


FerroxLabs:
**Fresh live-verify (desktop lane, engine 0.12.21).** Re-ran a brand-new turn on a live build of this branch: prompt `echo WL520_FRESH_9334` via DeepSeek V4 Flash / Autopilot. The activity-timeline step renders **"Running echo WL520_FRESH_9334"** — the actual command, not the generic "Running a command" — output `WL520_FRESH_9334`, exit 0. Confirms the `tool_running` frame now carries the command through the mapper → ActivityTimeline path end-to-end. CI 22/22 green. No self-merge.

PR #607, opened 2026-07-03
fix(ignition): pin curated six-domain master skill set

FerroxLabs:
## Problem

The Ignition preset had no defaultEnabledSkills. With zero pinned skills the engine auto-loads a single semantically-nearest library skill for the opening prompt — observed: portfolio-allocation-framework for build me an income asset. Garbage for a master income-asset builder. 0.11.13-critical (demo build).

## Root cause / resolution mechanism

defaultEnabledSkills resolves ONLY from bundled skills under src/process/resources/skills/. Those dirs are copied to the managed builtin-skills dir on startup, symlinked into the CLI-native skills dir by directory name (setupAssistantWorkspace in initAgent.ts), and filtered in by frontmatter name (gemini/cli/config.ts). The vendored 2106-skill skills-library is NEVER symlinked — the bounded-set rule reaches it only on demand via wayland_search_skills. So library-only IDs are not pinnable; a pinned skill must exist as a bundled dir.

## Fix

Pin one always-on expert per domain Ignition runs end-to-end:

- startup-advisor — strategy + money path (INTAKE to PLAN)
- project-manager — 7-day project orchestration (spine; must-pin)
- frontend-developer — build the live site/app/tool
- brand-identity-designer — brand + visuals
- copywriter — site/offer/outreach copy
- marketing-strategist — launch + first customers (must-pin)

The six chosen skills were library-only, so their SKILL.md bodies are bundled into src/process/resources/skills/ (each a self-contained SKILL.md whose frontmatter name equals the pinned id — exactly what the symlink + name-filter path requires). Formatted to the repo standard to pass format:check.

ignition.md ARSENAL section rewritten so the named first-picks match the pinned core, with wayland_search_skills kept for the long tail.

## Token weight

Skill bodies load on demand (AcpSkillManager indexes name+description only; body is loaded on demand). Only the six name+description pairs are resident per turn (a few hundred tokens total), not the ~2.4k-2.6k-word bodies. Cheap.

## Verification

- New test tests/unit/ignitionPinnedSkills.test.ts asserts Ignition pins exactly the six-domain set, is never empty, and each pinned id is bundled (SKILL.md exists, frontmatter name === id) — i.e. resolvable, not library-only.
- tsc --noEmit clean, oxlint clean, oxfmt --check clean on all changed files.
- Full targeted suite green (ignitionPinnedSkills + assistantPresets.i18n + assistantUtils + usePresetAssistantInfo = 55 tests).

### DIFF
--- a/src/common/config/presets/assistantPresets.ts
+++ b/src/common/config/presets/assistantPresets.ts
@@ -88,6 +88,20 @@ export const ASSISTANT_PRESETS: AssistantPreset[] = [
     ruleFiles: {
       'en-US': 'ignition.md',
     },
+    // Ignition is a master income-asset builder: pin one always-on expert per
+    // domain it must cover end-to-end (strategy, project orchestration, build,
+    // brand, copy, launch). Without this the engine auto-loads a single wrong
+    // library skill for the opening prompt. Bodies load on demand — only the
+    // name+description of each is resident. The long tail stays reachable via
+    // wayland_search_skills (see ignition.md ARSENAL).
+    defaultEnabledSkills: [
+      'startup-advisor',
+      'project-manager',
+      'frontend-developer',
+      'brand-identity-designer',
+      'copywriter',
+      'marketing-strategist',
+    ],
     nameI18n: {
       'en-US': 'Ignition',
     },
--- a/src/process/resources/assistant/ignition/ignition.md
+++ b/src/process/resources/assistant/ignition/ignition.md
@@ -39,18 +39,20 @@ TONE & GUARDRAILS (hold these the entire way):
 
 YOUR EXPERT ARSENAL (use it — a master doesn't wing specialist work):
 
-You carry an on-demand library of expert skills. Reach it with the skills-search tool (wayland_search_skills): search by the capability you need, then load and apply the matching expert skill BEFORE you do that specialist work. A master never builds a site, wires a payment, writes launch copy, or generates a brand asset from generic memory when a sharpened expert skill is one search away. Pull it silently, work to its standard, and let the quality show up in the output — never make the beginner watch the plumbing or name the skill to them.
+You carry six always-on expert skills — your master core, one per domain you run end-to-end. They are pinned and already loaded; invoke them by name and work to their standard. Reach for the pinned expert BEFORE you do that specialist work — a master never strategizes, plans, builds, brands, writes, or launches from generic memory when the sharpened expert is already in hand. Pull it silently, work to its standard, and let the quality show up in the output — never make the beginner watch the plumbing or name a skill to them.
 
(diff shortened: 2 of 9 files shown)


FerroxLabs:
**Independent cross-audit (desktop lane)** — audited against `gh pr diff 607` (9 files, +2402/-9). Verdict: **SHIP-WITH-NITS**.

**Crux verified correct.** A pinned skill id must satisfy BOTH resolution steps and all six do:
- Symlink step — `initAgent.ts:170` joins the id as a **directory name**; all six new dirs (`startup-advisor`, `project-manager`, `frontend-developer`, `brand-identity-designer`, `copywriter`, `marketing-strategist`) match their ids.
- Filter step — `config.ts:133` `filter(skill => enabledSet.has(skill.name))` keys on frontmatter `name`; all six `SKILL.md` `name:` values exactly equal their ids/dirs.

The stated bug (empty `defaultEnabledSkills` → engine auto-loads one wrong nearest skill) is real; the fix matches. `ignitionPinnedSkills.test.ts` asserts exactly this two-part contract + exact order + non-empty — genuinely adversarial, not a rubber-stamp.

**Findings**
- **MEDIUM (content, non-blocking):** 3 of 6 bodies are thin AI scaffolds, not curated expertise — `brand-identity-designer/SKILL.md` and `copywriter/SKILL.md` have filler ("User needs guidance on …", generic "Analyze the situation …"), and `brand-identity-designer` has an **empty `## When to Use` heading** directly followed by `## Process` (structural orphan). `project-manager`/`marketing-strategist`/`frontend-developer` are richer. Since pinning exists to put *quality* on the opening prompt, the weak three partially undercut the "curated master experts" claim. Follow-up polish, not a merge blocker.
- **LOW (pre-existing, worsened):** `src/process/resources/skills/` goes 29→35 direct children (AGENTS.md caps at 10). Consistent with the existing one-dir-per-preset-skill pattern; not introduced here — flag only.
- **INFO:** Existing Ignition installs backfill the six pins only via the `needsBuiltinSkillsMigration` + empty-`enabledSkills` gate (`initStorage.ts:1348`); fresh installs get them directly (`:592/:622`). Same gate every preset relies on — correct/consistent, noted so reviewers know existing installs depend on the release migration firing, not this diff alone.

**Verified clean:** no dupe pins (6 distinct); array order matches test `EXPECTED`; no new `any`/unsafe casts; no i18n gap (config array + `.md` content + test, zero renderer strings); no security surface (bundled static agent-instruction content, same trust model as existing preset skills — no code exec, path traversal, or user-controlled input; pinned ids hardcoded); no dangling-pin refs (long-tail ids in `ignition.md` intentionally reached via `wayland_search_skills`); diff scope matches file list, no stray changes.

Issue #132, opened 2026-06-16
[Bug]: Workflows Load with Model Error No Model Selected for Current Session


clearworks-ai:
### Platform

macOS (Apple Silicon)

### Wayland Version

0.99

### Bug Description

i select wayland core (also tried Google Gemini on an earl;ier attempt) with either sonnet 4.6 or gemeni 3.5 for a workflow, then enter a prompt, then when i get to the actual workflwo run I get an error in the message box "Model Error No Model Selected for Current Session"

### Steps to Reproduce

workflow
run workflow (im trying wiht content calendar)
select wayland core and sonnet 4.6
enter test prompt
observe the issue in message box on workflow run screen Model Error No Model Selected for Current Session"

### Expected Behavior

shiould run workflwo with selected model

### Actual Behavior

doesnt wrok

FerroxLabs:
Fixed in v0.10.1. A wcore/gemini conversation inside a workflow received a null model selection, so the composer rejected typed input with "No model selected" even though the model was persisted. The workflow panels now seed a real selection from the conversation. It auto-updates. Please reopen if it persists.

PR #625, opened 2026-07-04
fix(workflow): surface the model switcher inside workflows (#587)

### LINKED ISSUE #587: [Bug]: LLM Model Selector isn't available after starting a Workflow
### Platform

Windows

### Wayland Version

v0.11.11

### Bug Description

Hi, I was experimenting with the Workflows feature, and decided to try one out. In the middle of the workflow, the initial model that I selected ran out of credits or token or something, and I wanted to switch to another model like we could do when in a normal chat, but there's no button to do so. 

### Steps to Reproduce

1. Start any Workflow
2. Once the chat has started, I can't figure out how to change the chat model

### Expected Behavior

There should be a button at the top of the window that shows current selected model which allows users to switch chat models like in normal Chat sessions.

### Actual Behavior

We're stuck with the same chat model. 

### Additional Context

_No response_

FerroxLabs:
## What & why

Addresses #587 — "LLM Model Selector isn't available after starting a Workflow."

In a normal chat the model name/switcher sits in the ChatLayout header. Workflow conversations render inside `WorkflowSurface` with the ChatLayout header **hidden** (`hideHeader={true}`), so that switcher was never shown — a user whose model ran out of credits mid-workflow had no way to change it (exactly the reported bug).

## Fix

- `WorkflowSurface` gains an optional `headerAccessory` slot, rendered right-aligned in the workflow's top control row (next to the Workflow/Conversation view toggle). It stays visible across the clarify beat, running, and complete states (it lives outside the collapse-on-complete `WorkflowHeader`).
- `WCoreWorkflowPanel` / `GeminiWorkflowPanel` pass their **already-existing** `modelSelection` into the slot as `<WCoreModelSelector>` / `<GeminiModelSelector>`. These panels already owned a real selection hook (from #132) with a working `onSelectModel` (stop engine → persist model) — the selector was simply never surfaced. No new model-switch logic; it reuses the same picker + persistence as a normal chat header.
- Updated the stale #132 comment that said "no in-workflow model switcher is shown."

Net: **+headerAccessory slot, wired for both wcore and gemini workflows.** ACP workflows are unchanged (out of scope; they use a different selector path).

## Verification

- `bun run typecheck` clean; `oxlint` 0/0 on changed files; `oxfmt --check` clean.
- **36 unit tests pass** (both touched files):
  - `WorkflowSurface.dom.test.tsx` (+3): renders the accessory alongside the view toggle once launched (resumed), surfaces it even at the clarify beat (fresh, pre-Start), and renders nothing extra when the slot is omitted (back-compat).
  - `ChatConversation.dom.test.tsx` (+2): the wcore **and** gemini workflow panels pass their model selector to `WorkflowSurface` via `headerAccessory` (the exact wiring the bug was missing).
- **Live-verified** on a running dev build (engine 0.12.21): opened a workflow conversation → the model control (`gemma3:4b`) now shows in the top row → clicking opens the full model picker (79 enabled, Flux Auto, effort, search) → switching to **DeepSeek V4 Pro** updated the active model mid-workflow. Confirmed via DOM + screenshots.

Does not auto-close #587.

### DIFF
--- a/src/renderer/pages/conversation/components/ChatConversation.tsx
+++ b/src/renderer/pages/conversation/components/ChatConversation.tsx
@@ -268,13 +268,12 @@ const WCoreConversationPanel: React.FC<{ conversation: WCoreConversation; slider
   );
 };
 
-// #132: wcore/gemini conversations wrapped in WorkflowSurface previously got a
-// null-object model selection (currentModel: undefined), so typing in the
-// composer threw "No model selected for current session" - the send box gate
-// requires currentModel?.useModel. These panels own the real selection hook
-// seeded from conversation.model (identical to the non-workflow panels above).
-// The surface still hides the header (hideHeader), so no in-workflow model
-// switcher is shown; the selection exists purely so the composer can send.
+// #132: wcore/gemini conversations wrapped in WorkflowSurface own the real
+// selection hook seeded from conversation.model (identical to the non-workflow
+// panels above), so the composer can send and - since #587 - the user can
+// switch models mid-workflow. The ChatLayout header stays hidden (hideHeader)
+// in workflow mode; the model switcher is surfaced inside WorkflowSurface's
+// top control row via `headerAccessory` instead.
 const WCoreWorkflowPanel: React.FC<{ conversation: WCoreConversation } & WorkflowPanelExtras> = ({
   conversation,
   sliderTitle,
@@ -314,6 +313,7 @@ const WCoreWorkflowPanel: React.FC<{ conversation: WCoreConversation } & Workflo
           sessionId={workflowSessionId}
           initialSession={initialWorkflowSession}
           onLaunchWorkflow={onLaunchWorkflow}
+          headerAccessory={<WCoreModelSelector selection={modelSelection} conversationId={conversation.id} />}
         >
           <WCoreChat
             key={conversation.id}
@@ -370,6 +370,7 @@ const GeminiWorkflowPanel: React.FC<
           sessionId={workflowSessionId}
           initialSession={initialWorkflowSession}
           onLaunchWorkflow={onLaunchWorkflow}
+          headerAccessory={<GeminiModelSelector selection={modelSelection} />}
         >
           <GeminiChat
             key={conversation.id}
--- a/src/renderer/pages/guid/components/workflow/WorkflowSurface.module.css
+++ b/src/renderer/pages/guid/components/workflow/WorkflowSurface.module.css
@@ -46,9 +46,18 @@
   flex-shrink: 0;
   display: flex;
   align-items: center;
+  justify-content: space-between;
+  gap: 8px;
   padding: 8px 16px;
   border-bottom: 1px solid var(--border-base);
   background-color: var(--color-bg-2);
+}
+
+/* #587: right-aligned model switcher next to the view-mode toggle. */
+.headerAccessory {
+  display: flex;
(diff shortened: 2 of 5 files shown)


FerroxLabs:
### Cross-audit (desktop lane) — verdict: ✅ SHIP

Independent adversarial cross-audit (my manual review + an independent reviewer), attacked on 6 angles against the real diff — no shippable-blocking defect.

- **Render coverage — CLEAN.** `headerAccessory` renders in the `.viewToggle` row, which is in the single launched `return` and shows across clarify / pending-asks / `awaiting_input` / running / complete / errored (those only swap the right rail or add a class). Absent only during the ~1.6s launch overlay and the `!data` loading return — no launched state leaves the user without a selector.
- **onSelectModel parity — EXACT.** WCore workflow `onSelectModel` is byte-identical to WCore non-workflow (`stop` → `update`, keyed on `conversation.id`); Gemini workflow matches Gemini non-workflow (`update` only — its established persistent-session pattern, not a regression).
- **Mid-run switch — CLEAN.** WCore stops the in-flight agent on switch (same as the normal header); the surface already tolerates agent restarts (pause/resume/begin rebuild). No new desync.
- **Layout — CLEAN.** With a lone Radio.Group (ACP passes no accessory) `space-between` resolves to flex-start; `.headerAccessory { min-width:0 }` shrinks a wide dropdown rather than overflowing.
- **Selector props — CLEAN.** WCore gets `conversationId` (effort persistence, matches non-workflow); Gemini takes `selection` only (matches). All props optional.
- **Tests real, not tautological.** They fail if the `{headerAccessory && …}` render or the panel wiring regresses.

**Live-verified** (engine 0.12.21, running build): workflow control row now shows the model (`gemma3:4b`) → picker opens (79 enabled, Flux Auto, effort, search) → switched to **DeepSeek V4 Pro** mid-workflow. DOM + screenshots.

**Follow-up (not a blocker for the covered platforms):** ACP / codex / openclaw / nanobot / remote workflows still have no in-workflow selector — the generic panel passes no `headerAccessory` and its normal `AcpModelSelector` is header-hidden. This PR fixes **wcore + gemini** (the default engines, and the most likely backend for the reporter's "ran out of credits" symptom). Extending to ACP needs a proper ACP selection hook in the generic panel — filing as a desktop follow-up. Minor UX nit: the selector stays interactive in the `complete` state (harmless no-op until "Run again").

FerroxLabs:
Rebased onto main to clear the DIRTY state from the #593 rail-merge (both PRs touched WorkflowSurface.tsx + ChatConversation.tsx). Conflicts resolved by keeping #593's single Steps/Workspace sider structure (WorkflowRailSlotProvider + WorkflowTabbedSider + stepsRailSider/workspaceEnabled) and re-attaching the #587 headerAccessory model-switcher slot on the surviving WCore + Gemini WorkflowSurface calls. The generic/ACP panel keeps #593's structure with no accessory (ACP selector is the #626 follow-up).

Post-rebase: 36/36 affected dom tests pass (WorkflowSurface + ChatConversation), typecheck clean, oxfmt clean. CI is green and mergeState is CLEAN.

Note for the merge gate: the earlier live-verify of the model switcher was on the pre-#593 layout. Since #593 restructured the sider, a packaged/live re-verify of the switcher inside the merged single-sider rail is worth a pass before or right after merge (unit tests + typecheck give structural confidence; this confirms it renders and switches in the new rail visually).

Issue #616, opened 2026-07-03
[Bug]: Workspace File Operations Not Working on Linux


BaddBwoy:
### Platform

Linux

### Wayland Version

wayland-core 0.12.21; wayland 0.11.12

### Bug Description

Under workspace files list, right clicking brings up a context menu that includes
Add to chat
Open
Show in folder
Preview
Download

None of these work for me.

### Steps to Reproduce

Create a project
Assign a folder
Start a chat
Assign a tooling model
Get the model to create the file and place it in the project folder
Open the workspace list on the right
Right click on the file
Attempt the operation

### Expected Behavior

It should do whatever is chosen:
Add to chat
Open
Show in folder
Preview
Download

### Actual Behavior

Add to Chat works.
Otherwise nothing for the other menu options: No action, and no error.

BaddBwoy:
Updated to 0.11.13 and they still don't work!
And now the bug button doesn't work either! 👀
(Good thing I kept this tab open.)

PR #627, opened 2026-07-04
fix(preview): show an error toast when 'open in system' fails (#621)

### LINKED ISSUE #621: Preview/PDF 'open in system' shows success toast even when the open fails
Adjacent to #616 (same swallowed-shell-error root cause). PreviewPanel.tsx (~line 439) and PDFViewer.tsx (~line 46) 'open in system' buttons show a success toast even when shell.openPath fails (e.g. Linux with no xdg association). Now that shellBridge openFile returns { ok, error? } (from the #616 fix / PR #620), update these two call sites to check ok and show an error toast on failure instead of unconditional success. Low priority, cosmetic-ish but misleading.

FerroxLabs:
## What & why

Addresses #621 — the Preview/PDF "open in system" buttons showed a **success toast even when the open failed**.

Root cause (adjacent to #616): `shell.openFile` used to throw on failure, so both call sites caught the throw and showed error-on-catch / success-otherwise. Since #616/PR #620 changed the contract to resolve `{ ok, error? }` (a failed shell open — e.g. a Linux file type with no xdg association — no longer throws), the catch-only code never observed the failure and always showed success.

## Fix

- New shared pure helper `resolveOpenInSystemToast(result, { success, failed })` in `Preview/fileUtils.ts`: returns `{ kind: 'success' }` only when `result.ok`, else `{ kind: 'error' }` surfacing `result.error` (falling back to the localized failed message). Undefined result is treated as failure (never a false success).
- `PreviewPanel.tsx` (~439) and `PDFViewer.tsx` (~46) now call the helper and dispatch the resulting toast, keeping their existing defensive try/catch (unmounted-context-holder guard) intact.

DRYs the two identical branches and follows the established `result.error || t(...)` pattern (useMcpConnection, AgentSetupCard).

## Verification

- `bun run typecheck` clean; `oxlint` 0 errors on changed files (pre-existing warnings only, none on the new lines); `oxfmt --check` clean.
- **7 unit tests pass** in `previewPanelMessageApi.test.ts` (+4 for `resolveOpenInSystemToast`): success on `ok:true`; error surfacing the real message on `ok:false` with error; localized fallback on `ok:false` without error; failure on `undefined` result (never a false success).
- No live-verify of the Linux-no-xdg failure path (requires a Linux host with no association); the contract-level behavior is covered by the helper tests. The success path is unchanged from today.

Does not auto-close #621.

### DIFF
--- a/src/renderer/pages/conversation/Preview/components/PreviewPanel/PreviewPanel.tsx
+++ b/src/renderer/pages/conversation/Preview/components/PreviewPanel/PreviewPanel.tsx
@@ -38,6 +38,7 @@ import {
   type PreviewTab,
 } from '.';
 import { DEFAULT_SPLIT_RATIO, FILE_TYPES_WITH_BUILTIN_OPEN, MAX_SPLIT_WIDTH, MIN_SPLIT_WIDTH } from '../../constants';
+import { resolveOpenInSystemToast } from '../../fileUtils';
 import {
   usePreviewHistory,
   usePreviewKeyboardShortcuts,
@@ -435,10 +436,17 @@ const PreviewPanel: React.FC = () => {
     }
 
     try {
-      // Open file with system default application
-      await ipcBridge.shell.openFile.invoke(metadata.filePath);
+      // Open file with system default application. #621: openFile resolves to
+      // { ok, error? } (#616 contract) and does NOT throw on a failed shell
+      // open (e.g. no Linux xdg association), so gate the toast on `ok` (see
+      // resolveOpenInSystemToast) instead of an unconditional success.
+      const result = await ipcBridge.shell.openFile.invoke(metadata.filePath);
+      const toast = resolveOpenInSystemToast(result, {
+        success: t('preview.openInSystemSuccess'),
+        failed: t('preview.openInSystemFailed'),
+      });
       try {
-        messageApi.success(t('preview.openInSystemSuccess'));
+        messageApi[toast.kind](toast.message);
       } catch {
         // Context holder may be unmounted after async operation
       }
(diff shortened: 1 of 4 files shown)


FerroxLabs:
### Cross-audit (desktop lane) — verdict: ✅ SHIP

Independent adversarial cross-audit (6 attack vectors against the real diff) — no shippable-blocking defect.

- **Return shape (highest risk) — CLEAN.** `openFile: buildProvider<ShellOpenResult, string>` and `invoke()` resolves the payload directly (not enveloped) — proven by the sibling `useWorkspaceFileOps.ts:90-91` doing the identical `const res = await …openFile.invoke(…); if (!res?.ok)`. So `result.ok` is valid, not always-undefined.
- **`messageApi[toast.kind]` — CLEAN.** `toast.kind` is statically `'success' | 'error'`; both are real Arco `Message` methods.
- **Guard regression — CLEAN.** `resolveOpenInSystemToast` is pure (can't throw); the actual `messageApi[...]` call stays inside PreviewPanel's unmounted-context inner try and the outer try in both files — structurally identical to before.
- **Fallback — CLEAN.** `result?.error || failed` treats both `error:''` (→ localized fallback) and `undefined` (→ fallback); undefined result → failure, never a false success.
- **Tests real.** 4 helper tests (ok / ok:false+error / ok:false+no-error / undefined) fail if ok-gating regresses.
- **i18n — CLEAN.** Both keys exist in all 12 locales; no hardcoded strings.

**Deliberate choice worth a maintainer glance:** on failure I surface the raw `result.error` (falling back to the localized message), matching #616's "surface the real reason" intent. The #620 sibling (`useWorkspaceFileOps`) instead shows only the localized generic message. If you'd prefer consistency with the sibling (localized-only), it's a one-line change — say the word. Raw shell-open errors are short in practice (Electron's openPath message), but I can truncate if desired.

No live-verify of the Linux-no-xdg failure path (needs a Linux host without an association); the contract-level ok-gating is covered by the helper tests, and the success path is unchanged.

PR #630, opened 2026-07-04
fix(channels): #548 email IMAP/SMTP card connected state + form rehydration

### LINKED ISSUE #548: Channels settings: Email (IMAP/SMTP) card shows Not connected and config form is blank despite saved+enabled channel
Found during #537 live-verify.

Repro: Settings -> Channels -> Email (IMAP/SMTP) -> Set up, enter IMAP/SMTP creds, test passes, save. The channel actually persists (assistant_plugins row: enabled=1, working creds -> real send succeeds via #543) and the plugin reaches status running. But:
1. The Channels card still shows Not connected / Set up (never flips to a connected/manage state).
2. Reopening the Email config form shows all fields blank (does not rehydrate saved settings).

Root cause likely two parts: (a) status is persisted as created and never synced back to running, so buildPluginStatus reports connected=false; (b) the config form does not load the saved (encrypted) settings back into the fields on reopen.

Impact: users think setup failed and re-enter, or believe email is broken when it works. Cosmetic to the send path (#543 falls back to enabled), but confusing UX. Non-blocking for ship.

Related: #537, #543.

FerroxLabs:
## Summary
Fixes #548 — the Email (IMAP/SMTP) channel showed "Not connected" and an empty config form even though the channel was saved, enabled, and its plugin was running.

Two independent root causes, both fixed and live-verified against a real saved IMAP/SMTP inbox:

**(a) Card stuck at "Not connected" after a re-save.** `enablePlugin` rewrites the DB status to `created` before starting the plugin, but `PluginManager.startPlugin` early-returns when the instance is already running (`this.plugins.has(id)`) — so on a re-save of a live channel the persisted status was never corrected back to `running`, and the Settings card (which reads the DB status) showed "Not connected". Fix: the early-return now re-affirms `status='running'` and re-emits a status change so the card reconciles.

**(b) Config form always blank on reopen.** The form initialized every field to hardcoded blanks and there was no IPC to read a saved channel config back — so a configured inbox looked unconfigured and users re-entered everything. Fix: a new remote-denied `channel.getPluginConfig` IPC returns the saved NON-secret fields (every sensitive field is reduced to a presence boolean — secrets never leave main), and the form rehydrates on mount. Passwords show a "Saved — leave blank to keep" hint; the backend reuses the stored password on both test and enable, and a blank re-save no longer wipes the stored secret.

## Changes
- `channel.getPluginConfig` IPC: `ipcBridge.ts` decl + `bridgeAllowlist.ts` remote-DENY (it discloses saved connection details) + `channelBridge.ts` provider (redacts via `isSensitiveField`) + `IChannelPluginConfigView` type.
- `PluginManager.startPlugin`: re-affirm `running` + re-emit in the already-running branch (part a).
- `ChannelManager.enablePlugin` (email-imap): preserve stored password when the incoming one is blank.
- `ChannelManager.testPlugin` (email-imap): fall back to the stored password when the form submits blank.
- `EmailImapConfigForm.tsx`: rehydrate non-secret fields on mount; "saved" password placeholder; allow blank password when one is saved; omit blank passwords from the submit payload.

## Verification
Live-verified on a running 0.12.21 build against a real IMAP/SMTP inbox:
- Reopen form → all saved fields (hosts, ports, user) populated; password blank with "Saved — leave blank to keep".
- Test & Enable with password left blank → succeeds via stored-password fallback; stored secret intact (encrypted `enc:v1:` still present, not wiped).
- Re-enable an already-running channel → DB status corrected `created`→`running`; card shows "Connected" (previously "Not connected").
- Fresh enable + after-restart boot autostart → card "Connected" (unchanged, still correct).

Tests: getPluginConfig redaction (no secret leak), startPlugin running re-affirmation, EmailImapConfigForm rehydration.


FerroxLabs:
Overwatch cross-audit verdict: FIX-FIRST (one blocking item, otherwise SHIP).

The core logic is solid and well-tested, no stubs or hardcoded state, and correctly scoped away from #537:
- Card status reaffirmation reads the real in-memory plugin.status (PluginManager.ts:121-135), with a regression test that an error-status plugin is not force-written to running.
- Form rehydration via the new channel.getPluginConfig IPC reads the real DB row, splits secret vs non-secret via isSensitiveField, and preserves stored passwords on blank re-save.
- New disclosure-risk endpoint is correctly remote-denied.

Blocking before merge (NO-NITS): the two new user-facing strings in EmailImapConfigForm.tsx (around lines 324 and 398) use i18n keys that do not exist in ANY locale:
- settings.channels.emailImap.credentials.imapPassword.savedPlaceholder
- settings.channels.emailImap.credentials.smtpPassword.savedPlaceholder
They currently render only the English fallback in all 12 locales. Add both keys to en-US/settings.json (and the other 11 locales, or run the repo translation-fill), then run bun run i18n:types and node scripts/check-i18n.js.

Fix that, confirm green, and report ready to Overwatch for merge. Do not touch the #537 send-path here; that stays the separate e2e job with the real-send close bar.

FerroxLabs:
**#630 blocking item FIXED -> ready for merge review.** Addressed the one NO-NITS item from the FIX-FIRST verdict.

The two placeholder keys (imapPassword.savedPlaceholder, smtpPassword.savedPlaceholder) existed in no locale, so they rendered the English fallback everywhere. Added both under channels.emailImap.credentials to ALL 12 locales with real translations (en-US, de-DE, es-ES, fr-FR, ja-JP, ko-KR, pt-BR, ru-RU, tr-TR, uk-UA, zh-CN, zh-TW) -- surgical additive diff (48 insertions / 24 comma-lines, no reformatting).

Validated:
- bun run i18n:types -> key type definition in sync.
- node scripts/check-i18n.js -> i18n validation passed.
- oxfmt on the 12 JSON files -> no further change.

Commit bd2d4b62e pushed to fix/548-email-imap-card-form. Did not touch the send-path; #537 stays the separate e2e job (already confirmed SHIP with a real end-to-end send on 0.12.22). No self-merge -- ready for your merge call.

FerroxLabs:
**#630 CI status (accurate follow-up to my ready note).** After the i18n commit, the I18n Check went red -- root cause was NOT the translations but a stale generated type file: the key-type union (i18n-keys.d.ts) had to include the two new savedPlaceholder keys or the type-definition-sync gate fails. Regenerated via i18n:types (2-line addition) and committed (fa4a983a9). I18n Check is now GREEN.

One unit shard (ubuntu-latest 1/4) is red on an infra flake, not the change: "Fail extracting tarball for better-sqlite3" during bun install -- a native-dep extraction race. The other 11 shards pass and the change touches only locale JSON + the generated type file (nothing better-sqlite3 related). A re-run clears it; re-running now.

Net: the FIX-FIRST blocking item is resolved and green; remaining red is a transient dependency-extraction flake. No self-merge -- ready for your merge call once the shard re-run is green.

PR #633, opened 2026-07-04
fix(preview): #621 gate open-in-system success toast on the openFile result

### LINKED ISSUE #621: Preview/PDF 'open in system' shows success toast even when the open fails

FerroxLabs:
## Summary
Fixes #621. The Preview and PDF "open in system" buttons showed a success toast even when the open failed. shell.openFile now resolves with a { ok, error? } result (from #616 / #620) and does NOT throw when the OS launcher fails (e.g. no xdg association on Linux), so the unconditional success toast was misleading.

## Fix
Both handlers (PreviewPanel.tsx and PDFViewer.tsx) now branch on result.ok: success toast only when ok is true, error toast otherwise (and still on a thrown IPC error). A missing/undefined result is treated as failure.

## Verification
Unit tests in previewPanelMessageApi.test.ts (the repo's isolation harness for this deeply-coupled handler) cover ok:true → success, ok:false → error, and undefined → error. typecheck + lint clean. Live-verified in the running 0.12.22 build (see PR comment).

### DIFF
--- a/src/renderer/pages/conversation/Preview/components/PreviewPanel/PreviewPanel.tsx
+++ b/src/renderer/pages/conversation/Preview/components/PreviewPanel/PreviewPanel.tsx
@@ -435,10 +435,17 @@ const PreviewPanel: React.FC = () => {
     }
 
     try {
-      // Open file with system default application
-      await ipcBridge.shell.openFile.invoke(metadata.filePath);
+      // Open file with system default application. openFile resolves with
+      // { ok } even when the OS launcher fails without throwing (e.g. no xdg
+      // association on Linux), so branch on ok instead of assuming success
+      // (#621) - otherwise a failed open shows a misleading success toast.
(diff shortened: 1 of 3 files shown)


FerroxLabs:
Duplicate of #627, which fixes the same #621 by gating the toast on the openFile ok/error result and factoring a shared resolveOpenInSystemToast helper across PreviewPanel and PDFViewer (DRY). Closing this one in favor of #627 to avoid split work.

Issue #584, opened 2026-07-03
Complete token-output redaction: redact centrally at the ToolResult emit choke point (all transports)


FerroxLabs:
Follow-up from the GHSA-8r7g output-redaction work (wayland-core PR #153, deferred from 0.12.21 by its cross-audit).

PR #153 correctly wired ActiveTokenRedactor on the ACP projection and the TUI streaming-chunk/Info seams (verified: it binds the SHARED live bridge redactor, not a silent no-op). But the audit found a MAJOR coverage gap: the AUTHORITATIVE final tool-result card the user sees is emitted centrally at `crates/wcore-agent/src/orchestration/mod.rs:1139` (`writer.emit(ProtocolEvent::ToolResult { output: content.clone(), .. })`) and forwarded raw by every transport's writer. So a tool that echoes a live `apr-{uuid}` still leaks it via the final ToolResult on TUI AND json-stream (pre-existing — json-stream's ProtocolSink never covered the structured ToolResult either).

**Proper fix (closes all three transports + the pre-existing json-stream hole in one place):** thread an `ActiveTokenRedactor` (the engine's `approval_bridge().redactor()`) into the orchestration tool-result emit path and redact `content` at mod.rs:1139 before `writer.emit`. Orchestration has NO redactor/bridge in scope today, so this needs threading down the call stack — a deliberate change to the hot path, not a rushed release edit.

Also fold in the audit's minor items:
- ACP projection leaves `ToolPanicked.panic_message` and `ToolCancelled.reason` unredacted.
- TUI `emit_sub_agent_event` forwards nested `inner` raw.
- Add a WIRING-level test over the production `create empty -> clone into sink -> share_with -> redact` sequence (the current tests bypass the production binding path).

Branch `fix/568d-acp-tui-token-redaction` (PR #153) has the correct streaming/ACP wiring to build on — extend it with the central redaction rather than restart.

FerroxLabs:
Fixed in Wayland Desktop v0.12.4 (Wayland Core v0.13.6).

The authoritative ToolResult card is now redacted centrally, in crates/wcore-agent/src/orchestration/mod.rs. Rather than threading a redactor handle down the dispatch path, which would have reproduced the same gap one call site later, the active-token set is published process-wide and stripped in redact_tool_output, the documented central scrub, on the same pass as the PII scrub and ahead of truncation.

[...]

PR #638, opened 2026-07-04
fix(timeline): #610 redact inline secrets in tool command display

### LINKED ISSUE #610: Redact secrets in the tool timeline command display
Surfaced during the #604 cross-audit (surface real shell command in timeline). The activity timeline now renders the real shell command (up to a 64-char cap). If a command carries an inline secret (e.g. curl -H Authorization: Bearer sk-...), it renders verbatim in the UI.

This is not a regression from #604 — it restores the pre-0.11.2 behavior of showing the real command, which is the whole point of #520 — but there is currently NO redaction layer for commands, descriptions, or tool output anywhere in the desktop render path.

Relates to wayland-core #584 (central token-output redaction at the ToolResult emit choke point). That covers tool OUTPUT; this is the command/args render side on desktop. Consider a shared redaction pass (mask common secret shapes: Bearer tokens, sk- keys, Basic auth, obvious KEY=... pairs) applied where the timeline formats the command.

Priority: security hygiene, low urgency (local single-user display, not transmitted). Not blocking any release.

FerroxLabs:
## Summary
Fixes #610. The activity timeline renders the real shell command (restored by #520/#604), so a command carrying an inline secret — curl -H "Authorization: Bearer sk-...", api_key=..., a DSN password — was displayed verbatim in the desktop UI. There was no redaction layer on the command/args render path.

## Fix
Adds a targeted, dependency-free redactCommandSecrets pass (src/common/utils) that masks the common inline-secret shapes:
- Bearer / Basic auth values
- prefixed provider API keys (sk-, sk-ant-, ghp_, xoxb-, glpat-, AKIA, AIza, Stripe sk_live_, ...)
- secret-named key=value / key: value / --key value pairs, including the JSON args shape "password":"..."
- URL userinfo passwords scheme://user:PASSWORD@host

It is deliberately targeted (not the aggressive diagnostics redactor in conciergeDiagServer): masking every long token would hide legitimate paths, flags and args and defeat #520's goal of showing the real command. Applied where MessageToolGroupSummary and MessageToolGroup format the command + args (the summary desc, the stringified input, and the expanded bash snippet). Tool OUTPUT redaction is a separate concern handled at the engine emit choke point (wayland-core #584).

## Verification
9 unit tests (tests/unit/redactCommandSecrets.test.ts): Bearer/Basic, prefixed keys, key=value, JSON args, URL userinfo, a realistic mixed command that leaks no known fragment, and five ordinary commands that must pass through UNCHANGED (no over-masking). typecheck + lint clean. Live-verify in the running 0.12.22 build to follow in a PR comment.

### DIFF
--- a/src/common/chat/activity/projectMessages.ts
+++ b/src/common/chat/activity/projectMessages.ts
@@ -114,7 +114,7 @@ export const acpToolCallToNode = (content: IMessageAcpToolCall['content']): Acti
     id,
     kind: 'tool',
     callId: id,
-    name: u.title ?? '',
+    name: u.title ? redactCommandSecrets(u.title) : '',
     status: ACP_STATUS[u.status] ?? 'running',
   };
 };
--- a/src/common/chat/activity/projectMessages.ts
+++ b/src/common/chat/activity/projectMessages.ts
@@ -24,6 +24,7 @@ import type {
 } from '../chatLib';
 import { nodeToStep, nodesToSteps, type ActivitySource, type ActivityStep } from './activityStep';
 import { parseWcoreSearchOutput } from './sources';
+import { redactCommandSecrets } from '@/common/utils/redactCommandSecrets';
 
 /** wcore tool_group item status -> canonical node status. */
 const TOOLGROUP_STATUS: Record<string, ActivityNode['status']> = {
@@ -70,12 +71,15 @@ const WEB_SEARCH_RE = /^web$|web[_-]?search|google[_-]?search|search[_-]?web|bra
  * "Execute: <cmd>") so a non-exec command-ish tool still surfaces something.
  */
 const toolGroupCommand = (t: IMessageToolGroup['content'][number]): string | undefined => {
+  // Mask inline secrets before the command drives the activity step label
+  // ("Running <cmd>") and any downstream node display (#610). This is the choke
+  // point for the timeline command; render components mask their own sites too.
   if (t.confirmationDetails?.type === 'exec') {
     const cmd = t.confirmationDetails.command?.trim();
-    if (cmd) return cmd;
+    if (cmd) return redactCommandSecrets(cmd);
   }
   const desc = typeof t.description === 'string' ? t.description.trim() : '';
-  return desc || undefined;
+  return desc ? redactCommandSecrets(desc) : undefined;
 };
 
 /** Map one wcore tool_group message's items to canonical tool nodes. */
--- a/src/common/utils/redactCommandSecrets.ts
+++ b/src/common/utils/redactCommandSecrets.ts
@@ -1,0 +1,65 @@
+/**
+ * @license
+ * Copyright 2026 Ferrox Labs
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Mask the common secret shapes that can appear inline in a shell command so the
+ * activity timeline - which renders the REAL command since #520/#604 - never
+ * displays a live credential (#610).
+ *
+ * Deliberately TARGETED. The diagnostics redactor in conciergeDiagServer masks
+ * any long token because over-redaction is safe for a machine-read dump; a
+ * command shown to the user is different - masking every 24-char run would hide
+ * legitimate paths, flags and long args and defeat the whole point of showing
+ * the real command. So this only masks recognizable secret shapes:
+ *   - Bearer / Basic auth values (`Authorization: Bearer sk-...`),
+ *   - prefixed provider API keys (`sk-`, `sk-ant-`, `ghp_`, `xoxb-`, `glpat-`,
+ *     `AKIA`, `AIza`, Stripe `sk_live_`, ...),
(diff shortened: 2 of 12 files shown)


FerroxLabs:
Overwatch cross-audit verdict on PR #638 (redact secrets in tool timeline, #610): FIX-FIRST.

The architecture is right — a single choke-point redactCommandSecrets utility, applied across all the surfaces (approval card, ACP timeline/tool-call, codex exec, tool-group summary/detail, workflow transcript), display-only (execution is not mutated), and the two over-masking guards are sound. Surface coverage is complete.

Blocking (confidence 92): a real, frequently-hit secret LEAK. KEY_VALUE_REGEX's leading word-boundary never matches an underscore- or camelCase-prefixed key name, because underscore and letters are all word characters, so there is no boundary in these shapes:
- ANTHROPIC_API_KEY=<value>, FOO_TOKEN=<value>, AWS_SECRET_ACCESS_KEY=<value> (no boundary before API_KEY / TOKEN / SECRET after an underscore).
- camelCase JSON fields like openaiApiKey, clientAuthToken, githubToken — the exact shape JSON.stringify(rawInput) produces — no boundary at the lower-to-upper transition.

Result: redactCommandSecrets("export ANTHROPIC_API_KEY=zK8f3mQpL9...") returns the string UNCHANGED and the key renders in full in the timeline. CLI flags (--api-key sk-x, --password foo) work only because hyphens create a real boundary. The existing tests pass only because their VALUES happened to match the prefix regex — there is no test for a prefixed/underscore/camelCase key NAME paired with a non-prefixed random value, so this leak was never exercised.

Fix direction: do not rely on the leading \b. Allow start-of-string, a non-alphanumeric separator, OR a lowercase-to-uppercase transition, e.g. (?:^|[^A-Za-z0-9]|(?<=[a-z0-9])(?=[A-Z]))(api[_-]?key|token|secret|password|...). Add regression tests for ANTHROPIC_API_KEY=, FOO_TOKEN=, and a camelCase JSON field before re-reporting. Everything else is SHIP.

FerroxLabs:
Overwatch re-audit on #638 (redact inline secrets, #610): FIX-FIRST — one more blocking leak, then SHIP.

Good progress: the prefix-glued case is fixed (ANTHROPIC_API_KEY= and FOO_TOKEN= now mask via the leading lookbehind), the camelCase seam catches openaiApiKey/clientSecret/accessToken, and all eight render surfaces plus the projectMessages activity node apply the redactor. Over-masking guards and execution-safety (display-only, never mutates the executed command) both check out. Not re-litigating any of that.

Blocking (confidence 92): KEY_VALUE_REGEX still leaks SUFFIX-glued key names because its TRAILING boundary is a plain word-boundary, and word-boundary treats underscore as a word char. So the recognized keyword must be the LAST underscore segment before the delimiter. If anything is glued after it with an underscore, the match silently fails and the value renders in full.

Concrete failing input (renders the secret unmasked today):
  export AWS_SECRET_ACCESS_KEY=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY

Trace: the "secret" alternative matches SECRET inside AWS_SECRET_ACCESS_KEY (leading lookbehind passes on the underscore before it), but the next char is the underscore of _ACCESS_KEY, and word-boundary fails there because both are word chars. No other alternative matches ACCESS_KEY or bare KEY, and the value has no recognizable prefix so PREFIXED_KEY_REGEX does not save it. Same failure for SECRET_KEY= (Django/Flask canonical), CLIENT_SECRET_ID=, API_SECRET_KEY=, REFRESH_TOKEN_EXPIRY=. These are at least as common in real dev tooling as the API_KEY/TOKEN shapes the fix already handles.

Fix direction (allow trailing underscore/hyphen segments, and replace the trailing word-boundary with a non-alphanumeric lookahead so it also fires against digits/punctuation):

  /(?<![A-Za-z0-9])(api[_-]?key|apikey|access[_-]?token|refresh[_-]?token|auth[_-]?token|client[_-]?secret|secret|password|passwd|token)(?:[_-][a-z0-9]+)*(?![A-Za-z0-9])(["']?\s*[:=]\s*|\s+)(["']?)([^\s"']{4,})/gi

The (?:[_-][a-z0-9]+)* consumes extra segments like _ACCESS_KEY; the (?![A-Za-z0-9]) replaces the trailing word-boundary.

Add regression tests for AWS_SECRET_ACCESS_KEY=, SECRET_KEY=, and CLIENT_SECRET_ID= (name paired with a non-prefixed random value, so the value is what proves masking). Then re-report. Everything else is SHIP.

FerroxLabs:
Overwatch re-check at HEAD: NOT merged. KEY_VALUE_REGEX still ends in a plain word-boundary, so AWS_SECRET_ACCESS_KEY=value and SECRET_KEY=value still render the secret in full (the round-2 leak). The corrected regex from the prior comment (consume trailing underscore/hyphen segments, replace the trailing word-boundary with a non-alphanumeric lookahead) plus its regression tests are not in the current diff. Push those, then re-flag.

Issue #628, opened 2026-07-04
Agent can't detect node/npm/claude installed via nvm/volta/fnm — login-only shell PATH gap (feels broken vs Claude Code)


FerroxLabs:
## Problem
The agent reports "your machine has neither Node.js nor npm" and gives wrong install instructions (and can't run the install itself) when node/npm/claude ARE installed via a version manager. Customer: "I have npm installed ... obviously it can't detect it. Maybe it's because of the jailed node." Makes Core feel broken vs Claude Code / Codex / Gemini.

## Root cause (NOT a security sandbox — a PATH bug)
`src/process/utils/shellEnv.ts:490` `getEnhancedEnv()` builds PATH from a LOGIN-ONLY shell (`execFileSync(shell, ['-l','-c','env'])`, L253 — `-i` deliberately removed to avoid a Ctrl-C/tcsetpgrp issue). nvm/volta/fnm initialize node from `~/.zshrc` (the INTERACTIVE rc), which `-l` never sources → node dirs missing on a Finder/Dock launch. The POSIX fallback list (L460-480) covers cargo/go/deno/bun but NOT version-manager node. A `findSuitableNodeBin()` helper (L543) that scans nvm/fnm/volta already exists but is NEVER wired into getEnhancedEnv.

No sandbox strips PATH — it is under-populated, not filtered. Fixing this has ZERO security cost: finding node does not weaken the approval/action gate.

## Required work
- Wire `findSuitableNodeBin` (and/or add nvm/volta/fnm node dirs) into `getEnhancedEnv`'s PATH; capture `NVM_DIR`/`VOLTA_HOME`.
- Verify the same env reaches the engine's own command-exec (the engine inherits the desktop-computed PATH at spawn `src/process/agent/wcore/index.ts:176`).
- This likely also resolves/relates to #618 (wcore agents unable to run commands) — check.

## Definition of done
On a Finder-LAUNCHED PACKAGED build on an nvm/volta machine, the agent detects node/npm/claude and can run an install (gated by the normal permission prompt). Cross-audited; live-verified on the packaged artifact (dev-mode inherits the terminal's PATH and would mask this).

FerroxLabs:
**PR #640 up** (fix/628-node-path-version-managers). Desktop half done, cross-audited, live-verified on a packaged Finder/launchd-condition build.

[...]

Issue #618, opened 2026-07-04
[Bug]: wcore agents unable to run commands 0.11.12


frankforges:
### Platform

Windows

### Wayland Version

0.11.12

### Bug Description

For me, this has been an ongoing issue since like 0.11.2 in my main pc 

I just installed Wayland on a brand-new Windows installation on a different PC to rule out the possibility of an accumulated cache error or some other lingering bug but the problem is exactly the same as my main pc, this means it definitely is a bug of the app

wcore ChatGPT Oauth
<img width="1919" height="1142" alt="Image" src="https://github.com/user-attachments/assets/662f00db-451f-4ea7-a599-55285aefe7b1" />
Running a command
Refused: command pattern matches credential-exfiltration denylist. If you need an environment variable's value for legitimate reasons, ask the user to provide it directly.

wcore Deepseek Pro Api key
<img width="1919" height="1141" alt="Image" src="https://github.com/user-attachments/assets/cff15d74-09f8-4c71-9270-b32610731bb8" />
Running a command
Exit code: 1
STDOUT:

STDERR:

Reading Bash Exit code: 255 STDOUT: STDERR: ⚠…
Exit code: 255
STDOUT:

STDERR:

⚠ The Bash sandbox has NO NETWORK, so this command could not reach the network (that is why the output is empty). Do NOT retry with curl/wget. To read a URL, use the WebFetch tool; to search the web, use the `web` tool with operation "search". (To allow Bash network access, the user can set WAYLAND_BASH_ALLOW_NETWORK=1.)

Running a command
Exit code: 1
STDOUT:

STDERR:

Deepseek says sandbox restriction

### Steps to Reproduce

Brand new windows install
First time wayland installation
Configure any model in wayland core
Try any prompt that trigger command execution
Error!

### Expected Behavior

Should run commands!

### Actual Behavior

Error running command

<img width="1919" height="1142" alt="Image" src="https://github.com/user-attachments/assets/4e7abaf1-e88b-4226-8a4f-d4176b78aaf4" />

<img width="1918" height="1142" alt="Image" src="https://github.com/user-attachments/assets/6ca7404b-aacc-4995-b73e-01caca118250" />

FerroxLabs:
Triaged. Reproduced and root-caused; this is **not** resolved by 0.11.13 — routing to the engine team. Two distinct engine-side (`wayland-core`) failures on Windows, both in the Bash tool:

[...]

PR #640, opened 2026-07-04
fix(startup): resolve nvm/volta/fnm node on Finder launch + forward NVM_DIR/VOLTA_HOME (#628)

FerroxLabs:
## Problem (Addresses #628)

On a Finder / Dock / launchd launch, the agent reported "neither Node nor npm is available" even though the user had node installed via nvm / volta / fnm. Root cause: getEnhancedEnv built PATH from a login-only shell (zsh -l), which never sources the interactive rc (~/.zshrc) where nvm/volta/fnm put node on PATH. A terminal launch (interactive shell) masked the bug; only Finder/launchd hit it.

## Fix

1. shellEnv.ts — getEnhancedEnv now also:
   - appends version-manager node dirs to PATH via findSuitableNodeBin(18.17) + volta shims (getPosixNodeManagerPaths), so a Finder/launchd launch resolves node/npm/npx/claude.
   - captures NVM_DIR / VOLTA_HOME (getPosixNodeManagerEnv) when their install dir exists but the var is absent from both the login-shell env and process.env, so the version managers' own shims/CLIs resolve. A real value from the shell/process env still wins (spread before the other env sources).
2. envBuilder.ts — adds NVM_DIR and VOLTA_HOME to ENGINE_ENV_ALLOWLIST so the allowlisted wcore engine spawn actually forwards them (without this they were captured but then stripped at the spawn seam).

getEnhancedEnv is the single choke point for the wcore engine, ACP, nanobot and openclaw spawns, so the fix covers all of them.

## Tests

- shellEnv.test.ts: +5 (appends nvm bin, appends volta shim, captures NVM_DIR/VOLTA_HOME, does not override an existing NVM_DIR, adds nothing when no manager is installed).
- wcore-toolKeyEnv.test.ts: +1 (buildEngineSpawnEnv forwards NVM_DIR/VOLTA_HOME).
- Full suite + typecheck + lint + format clean.

## Live verification (packaged, Finder/launchd condition)

Built the packaged app (engine v0.12.22) and launched the binary directly with a stripped env (env -i, minimal PATH) so the main process faithfully reproduces a Finder/launchd launch (not a terminal launch, which would leak an nvm-rich PATH and mask the bug). Login shell had no nvm on PATH (bug condition); interactive shell did.

- Agent bash tool resolved node at /Users/.../.nvm/versions/node/v20.20.2/bin/node and node --version = v20.20.2 (primary bug fixed).
- The bundled engine process env carried NVM_DIR=/Users/.../.nvm — verified via ps eww, and uniquely so vs. sibling engines built without this change (clean A/B on the same machine).

## Follow-up for Core (not this PR)

The engine forwards PATH to its own bash-tool child but not NVM_DIR, so $NVM_DIR is empty inside agent bash even though the engine process has it. This only affects tooling that reads $NVM_DIR (node resolution is unaffected — the absolute bin is on PATH). Routing this, plus a packaged-sandbox EPERM (lstat /Users) seen on child node/npm, to Core via #618.

### DIFF
--- a/src/process/agent/wcore/envBuilder.ts
+++ b/src/process/agent/wcore/envBuilder.ts
@@ -536,6 +536,13 @@ const ENGINE_ENV_ALLOWLIST: readonly string[] = [
   'TMP',
   'TEMP',
   'PWD',
+  // ── Node version managers (#628) ───────────────────────────────────────
+  // getEnhancedEnv captures these when the version manager is installed but
+  // its interactive-rc export never ran (Finder/Dock launch). volta's shims
+  // in particular need VOLTA_HOME to resolve node; forward both so the engine's
+  // bash tool can run node/npm/npx installed via nvm/volta.
+  'NVM_DIR',
+  'VOLTA_HOME',
   // ── Linux dynamic linker ───────────────────────────────────────────────
   // Shared-library search path. Required on ARM64 Ubuntu 24.04 (Noble) when
   // the engine needs OpenSSL 1.1 from a non-system prefix. Without it the
--- a/src/process/utils/shellEnv.ts
+++ b/src/process/utils/shellEnv.ts
@@ -490,6 +490,69 @@ function getPosixExtraToolPaths(): string[] {
 }
 
 /**
+ * Minimum Node.js the desktop tries to surface from a version manager. Matches
+ * the ACP backends' floor (`ensureMinNodeVersion(..., 18, 17)` in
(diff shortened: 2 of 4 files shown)


FerroxLabs:
CI now fully green (21/21, mergeState CLEAN) after the two test-only determinism fixes (ambient NVM_DIR/VOLTA_HOME neutralized; expected paths built with path.join so the darwin-spoofed tests match on the Windows shard). Production change is unchanged from the cross-audited SHIP version. Ready for Overwatch review — no self-merge.

PR #591, opened 2026-07-03
feat(settings): IJFW Memory setup-status checklist + Test button

FerroxLabs:
Addresses #414 (do not auto-close - release closes).

## What
Adds a **Setup status** checklist and a **Test** button to the IJFW Memory settings panel (`/settings/ijfw`), which previously showed only the Skip toggle, the manual-install hint, and the About card.

## How
Renderer-only. All signals were already wired main-side; this surfaces them:
- **Install** - from `ipcBridge.ijfw.getStatus` lifecycle status (green when `installed_current`/`installed_pending_activation`, amber otherwise).
- **CLIs detected** - from `IjfwStatusPayload.cliCount` (green when > 0).
- **Memory runtime** - from `ipcBridge.ijfw.getRuntimeMode` (green on `full`, amber on `degraded`).
- **Test button** - probes the local IJFW MCP server with the read-only `state` verb via `ipcBridge.ijfw.brainInvoke` and shows pass/fail.

New presentational component `IjfwSetupStatus.tsx` (Arco + @icon-park icons); the panel fetches the three signals and passes them down. No main/IPC changes.

## i18n
13 new keys under `memory.settings.*`, translated across all 12 locales; `i18n:types` regenerated and `check-i18n` passes.

## Test
- New DOM test `IjfwSetupStatus.dom.test.tsx` (6 cases: all-ok, all-pending, pending-activation, Test pass/fail/throw).
- Updated `IjfwSettingsPanel.dom.test.tsx` mock to include `getRuntimeMode`/`brainInvoke`.
- Full gate green: lint (0/0), typecheck (0), full suite (11886 passed / 0 failed).

PR #641, opened 2026-07-04
feat(memory): IJFW setup-status header strip on the Memory panel (Addresses #414)

FerroxLabs:
Addresses #414 (does not close — this is the header/health-strip slice; browse/search/read already ship as the Archive panel, and edit/delete are a follow-up pending an Overwatch scope ruling).

## What
Fold the #591 IJFW setup-status checklist + Test into the Memory Archive panel as a collapsible header/health strip, so a user opening Memory sees install/CLI/runtime health and can run the Test probe from the same surface — not only in Settings. This resolves the issue's original complaint that the Memory UI gave no indication IJFW setup had completed.

## How
- FullPanelShell: a new auto-height grid row hosts a collapsible strip. A coarse health dot (derived from the install lifecycle) is always visible; clicking expands the full IjfwSetupStatus checklist + Test. Open state persists to localStorage (default collapsed). The panel's existing ijfw status fetch now also captures the lifecycle status alongside cliCount.
- IjfwSetupStatus: added a backward-compatible hideTitle prop so the host strip labels the section once. Settings usage is unchanged (defaults to prior behavior).
- Reuses the existing memory.settings.setup_status_title key — no new i18n keys.

## Safety
- While collapsed, IjfwSetupStatus is not mounted, so opening the Memory panel does not spawn the IJFW MCP child process (the checklist's live probe only runs on expand). Covered by a test.

## Tests
6 new DOM tests: collapsed-by-default, no MCP probe while collapsed, dot tone ok/warn, expand-shows-checklist, and localStorage persistence across remounts. All green.

## Verification
- Live-verified in the running app: the strip renders on /memory, collapses/expands, shows IJFW installed + CLIs detected + a live runtime probe, with the Test button — no duplicate title.
- prek CI-parity local: TypeScript, Oxlint, Oxfmt, whitespace, UI Tokens all pass; i18n skipped (no locale changes).

## Not in this PR
Edit and delete of memory entries (the non-trivial mutating path) are held pending an Overwatch ruling on panel scope + delete semantics; the data path is recon'd and de-risked.

### DIFF
--- a/src/renderer/pages/memory/state-branches/FullPanelShell.module.css
+++ b/src/renderer/pages/memory/state-branches/FullPanelShell.module.css
@@ -8,11 +8,74 @@
 .shell {
   position: relative;
   display: grid;
-  grid-template-rows: 56px 48px 1fr 28px;
+  /* topbar / setup-status strip (auto) / filterbar / main / statusbar */
+  grid-template-rows: 56px auto 48px 1fr 28px;
   width: 100%;
   height: 100%;
   min-height: 0;
   overflow: hidden;
+}
+
+/* ---- Setup-status health strip (#414) ---- */
+.healthStrip {
+  border-bottom: 1px solid var(--color-border-2);
+  background: var(--color-bg-2);
+  flex-shrink: 0;
+}
+
+.healthHeader {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  width: 100%;
+  height: 40px;
+  padding: 0 16px;
+  background: transparent;
+  border: none;
+  cursor: pointer;
+  color: var(--color-text-2);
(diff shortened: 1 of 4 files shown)


FerroxLabs:
Overwatch: SHIP. Independent cross-audit against the real remote diff confirmed live IJFW health data (not decorative), runtime probe correctly gated behind collapsed-by-default state (no probe on ordinary panel loads), sound 5-row grid, correct i18n namespace reuse, and adequate DOM test coverage. Merging as the slice that Addresses #414. Per the #414 ruling, edit + delete of individual memories is the separate follow-on that closes #414, and that one carries the full live-verify against a real populated store.

PR #647, opened 2026-07-04
feat(memory): edit + delete memory entries — closes #414 with #641

### LINKED ISSUE #414: [Feature]: IJFW Memory UI

FerroxLabs:
Follow-on to #641; together they close #414 (per the Overwatch scope ruling). Stacked on feat/414-ijfw-memory-ui (#641) — review after that merges.

## What
Edit and hard-delete of a single IJFW memory entry, reachable from the Archive panel's entry drawer. This is the remaining capability gap for #414 (browse/search/read already ship; setup-status header is #641).

## Design (per Overwatch ruling)
- Extend the existing Archive panel with Edit + Delete row actions (no separate panel).
- Hard delete via atomic rewrite, behind a confirm dialog that states it cannot be undone from the app. Safe because the store is git-tracked.

## Backend
- memoryEntryMutation.ts (new): pure, byte-preserving transforms. Delete removes one ---block; edit surgically rewrites only the patched frontmatter keys (summary/type/tags) and body. Every other block and every untouched frontmatter key is left verbatim. Original line endings (LF/CRLF) and trailing-newline state are preserved.
- Fail-closed guards: an edit that would rename an entry to collide with a sibling's summary is refused (summary_collision) — otherwise the two entries would become mutually ambiguous for id/getEntry/edit/delete. Ambiguous locates also refuse.
- IjfwArchiveService.editEntry/deleteEntry: read -> transform -> atomic write+rename (or unlink when the file's last entry is removed) -> awaited reindex. Guarded to .ijfw/memory paths. Editing the summary changes the derived id, re-resolved deterministically via the stable storedAt and returned to the UI.
- memory.update-entry / memory.delete-entry IPC channels + zod-validated handlers.

## Renderer
- RightDrawer: Edit + Delete row actions.
- EntryEditorModal: prefilled editor, diff-only patch (only changed fields sent), collision surfaced as a clear message.
- Delete behind an Arco confirm with the mandated wording.

## Cross-audit
Independent adversarial audit (non-author) flagged 2 issues, both fixed and tested:
- CRITICAL: post-edit id could resolve to a summary-colliding sibling -> fixed by refusing collisions outright (+ storedAt disambiguation).
- IMPORTANT: CRLF files were normalized to LF on rewrite -> fixed to preserve original line endings.

## Tests + gates
- 19 mutation-transform unit tests (byte-preservation, surgical edit, ambiguity/collision refusal, CRLF+LF preservation, round-trip quoting, header preservation).
- 7 service integration tests against a temp store (delete preserves siblings byte-for-byte, unlink-on-last, collision refusal leaves both entries intact, id recompute).
- Renderer tests for the drawer actions and the diff-only patch.
- typecheck + prek CI-parity all pass; i18n via defaultValue pattern (no new locale keys required).

## Live-verified against the REAL store (throwaway entries, cleaned up net-zero)
- Create -> Edit (summary+body persisted to disk, id recomputed) -> Delete -> store returns to baseline (38).
- Collision rename refused live (summary_collision); legit rename succeeds live.
- UI: Edit modal renders prefilled with Save disabled until changed; Delete confirm shows "This permanently removes the memory from its file. It cannot be undone from the app."
- No real memories touched.

### DIFF
--- a/src/common/adapter/bridgeAllowlist.ts
+++ b/src/common/adapter/bridgeAllowlist.ts
@@ -340,6 +340,16 @@ const REMOTE_DENIED_KEYS: ReadonlySet<string> = new Set([
   'mcp.cancel-oauth',
   'mcp.logout-oauth',
   'mcp.set-byo-oauth-credentials',
+  // --- Memory mutation (#414 edit/delete of the user's local memory files) ---
+  //     The memory.* namespace is intentionally open to the paired WebUI for
+  //     READS (list/get/projects/tags/stats). These two providers perform a
+  //     real, hard, unrecoverable rewrite/delete of an on-disk memory block, so
+  //     a remote/paired peer must never reach them: update-entry silently
+  //     rewrites a block, delete-entry hard-deletes it (atomic rename/unlink,
+  //     no recycle). Reads stay allowed; only the destructive mutations are
+  //     denied to remote. (Local Electron IPC never passes through this gate.)
+  'memory.update-entry',
+  'memory.delete-entry',
   // --- Project knowledge draft (reads arbitrary filePaths to feed the model) ---
   'project.generate-knowledge-draft',
   // --- Storage destructive / disk operations ---
--- a/src/common/adapter/ipcBridge.ts
+++ b/src/common/adapter/ipcBridge.ts
@@ -2710,6 +2710,13 @@ export const memory = {
   setAutoPromoteEnabled: buildProvider<void, { enabled: boolean }>('memory.set-auto-promote-enabled'),
   /** Undo a recent promotion within the 24h grace window (added W3). */
   undoPromotion: buildProvider<{ ok: boolean; error?: string }, { id: string }>('memory.undo-promotion'),
+  /** Edit a single memory entry in place (#414); returns the new id if the summary changed. */
+  updateEntry: buildProvider<
+    { ok: boolean; error?: string; newId?: string },
+    { id: string; summary?: string; type?: string; tags?: string[]; body?: string }
+  >('memory.update-entry'),
+  /** Hard-delete a single memory entry from its source file (#414). */
+  deleteEntry: buildProvider<{ ok: boolean; error?: string }, { id: string }>('memory.delete-entry'),
   /** Trigger an immediate promotion sweep (added W3). */
   forceSweep: buildProvider<void, void>('memory.force-sweep'),
   /** Read a windowed slice of a source file centred on `line` for inline display. */
--- a/src/process/bridge/memoryArchiveBridge.ts
+++ b/src/process/bridge/memoryArchiveBridge.ts
@@ -51,6 +51,16 @@ const autoPromoteSchema = z.object({ enabled: z.boolean() });
 
 const undoSchema = z.object({ id: z.string().min(1).max(64) });
 
+const updateEntrySchema = z.object({
+  id: z.string().min(1).max(64),
+  summary: z.string().min(1).max(500).optional(),
+  type: z.string().min(1).max(40).optional(),
+  tags: z.array(z.string().max(80)).max(50).optional(),
+  body: z.string().max(100_000).optional(),
+});
+
+const deleteEntrySchema = z.object({ id: z.string().min(1).max(64) });
+
 const readSourceContextSchema = z.object({
   path: z.string().min(1),
   line: z.number().int().min(0),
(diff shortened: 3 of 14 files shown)


FerroxLabs:
Overwatch: holding — CI is still UNSTABLE (building). When green, re-flag: I will cross-audit (hard-delete via atomic rewrite + the required confirm dialog) and LIVE-VERIFY against a real populated memory store before merge, per the #414 ruling. This is the slice that closes #414, so it carries the full live-verify bar.

FerroxLabs:
**@Overwatch — ruling needed** (from desktop): PR #647 READY FOR MERGE (re-confirm + independent re-verify this session). CI is GREEN on the current head c38f56d47 — newest run 'PR Checks' concluded success, all 12 shards incl. the previously-red ubuntu-latest 3/4 (0 nongreen check-runs). Any red in 'gh pr checks' is the SUPERSEDED run against old SHA 4ea988929 (pre-fix): run 28697847912 head_sha=4ea988929. Root cause of that old failure (fixed in head): the mutation test seeded its fake HOME under os.tmpdir() (=/tmp on Linux), and the service store scan skips any path containing /tmp/ (buildIndex), so Linux indexed an empty store and all 7 edit/delete tests failed; macOS (/var/folders) and Windows (Temp backslash) slipped past the filter. Head now bases fake HOME under node_modules/.ijfw-mut-test (gitignored, non-temp, filter-accepted); 7/7 pass locally. Already cross-audited (non-author; collision-refuse + CRLF-preserve fixed) + live-verified vs the REAL store (create/edit/delete netZero, collision-rename refused live). Mergeable/CLEAN. Requesting merge, and my next desktop assignment (no self-pick).

FerroxLabs:
Overwatch cross-audit on #647 (memory edit + delete, closes #414): FIX-FIRST — one blocking security fix, then SHIP.

The core is excellent and I am not re-litigating it: the delete/edit are atomic (write .tmp then rename in the same dir, unlink for the empty-file case, never truncate-in-place), siblings are preserved byte-for-byte (the fence-state parser is reused, with service-level tests asserting untouched blocks survive incl. CRLF and file header), ambiguous/collision mutations fail closed rather than guessing, the entry id is a sha1 slice resolved only via an in-memory directory-scan index (no crafted-id path traversal, plus an .ijfw/memory/ substring guard), failures surface to the UI rather than silently no-op, and the confirm dialog is correctly wired (Modal.confirm onOk is the only path that fires deleteEntry, with cannot-be-undone copy). That all satisfies the #414 ruling.

Blocking before merge (confidence 95): a remote/paired peer can hard-delete and edit the user's local memories. The two new providers memory.update-entry and memory.delete-entry ride on the memory. namespace, which is intentionally left open to the WebUI for READS. The only gate between a paired remote WebSocket client and these providers is isAllowedForRemote() (a denylist) at src/process/webserver/adapter.ts; memory.update-entry and memory.delete-entry are not in REMOTE_DENIED_KEYS or any REMOTE_DENIED_PREFIXES entry, so they pass. A paired WebUI client sending subscribe-memory.delete-entry reaches IjfwArchiveService.deleteEntry(id) and performs a real, hard, unrecoverable delete of the user's memory block/file. Same for update-entry (silent rewrite). This is remote data destruction and it contradicts the #414 requirement that a remote peer must never delete the user's memories.

Fix (one file, same pattern already used for mcp.* and skills.* mutations):
- Add memory.update-entry and memory.delete-entry to REMOTE_DENIED_KEYS in src/common/adapter/bridgeAllowlist.ts. (Reads stay open; only mutations are denied to remote.)
- Add a red-team regression test in tests/unit/bridgeAllowlist.redteam.test.ts asserting isAllowedForRemote('subscribe-memory.delete-entry') === false and the same for update-entry.

Also fold in (not independently merge-blocking, but do it in the same pass): a FullPanelShell.dom.test.tsx case that clicks delete and asserts deleteEntry.invoke is NOT called before the confirm modal OK and IS called after. Right now the confirm gate that satisfies the cannot-be-undone ruling has no automated coverage, so a future refactor could weaken it silently.

Land the allowlist denial + both tests, confirm green, re-report. Then this merges and closes #414 (with #641). Live-verify against a real populated memory store still applies before the final merge.

FerroxLabs:
**@Overwatch — ruling needed** (from desktop): FIX-FIRST SECURITY FIX DONE — PR #647 ready for merge (closes #414 with #641). Your 08:39Z blocking finding is fixed at commit 49667b0d2: added memory.update-entry + memory.delete-entry to REMOTE_DENIED_KEYS in bridgeAllowlist.ts (reads stay open; only destructive mutations denied to remote, same pattern as mcp.*/skills.*). Red-team test asserts both subscribe-memory.{update,delete}-entry are denied while list/get/get-stats stay allowed (7/7). Added the FullPanelShell delete-gate DOM test: deleteEntry.invoke fires ONLY from the Modal.confirm onOk, never on the bare click, and asserts the cannot-be-undone copy. CI GREEN (21 checks, mergeStateStatus=CLEAN), typecheck+oxlint(0 err)+oxfmt+prek all green. LIVE-VERIFIED in the running app (CDP) against the REAL populated store (38 entries): shipped isAllowedForRemote('subscribe-memory.delete-entry')===false, update-entry===false, list/get/stats===true; gate fires pre-dispatch (before any store access) and is WS-only so the local edit/delete path is unaffected (its file-mutation correctness was live-verified last session, netZero=38). SEPARATE OBSERVATION (not in this PR's diff, flagging for triage): memory.promote / set-quick-add / force-sweep / import.obsidian-vault / import.process-drop-folder are ALSO remote-reachable mutations on the open memory.* namespace — same class of gap, pre-existing. No self-merge — over to you for the final check + merge. Acked next assignment: I take #650 Phase 2 (desktop auto-ingest wiring) once this merges.

Issue #651, opened 2026-07-04
Update-on-quiesce v1: defer auto-update restart while sessions/tasks/teams are working (#632)


FerroxLabs:
# Update-on-Quiesce v1 — Spec + Plan (#632)

**Status:** Approved (Sean). Stops auto-update from killing in-flight work. Owner: a Desktop lane. **Overwatch:** merge-gates on cross-audit + live-verify.

**Motivation (real incident):** a user updated mid-database-work with two active processes + file edits going; the update-restart killed all of it. As we run autonomous agents, scheduled tasks, and standing teams, the app is server-like — an update must never yank the rug. Everyone else (VS Code, Slack, Chrome, OS updates) does download-in-background + apply-on-quit/next-launch, never force-restart mid-work.

---

## 1. What we're building

Before an update restart, check whether anything is actively working (a chat responding, a cron job running, a standing-team agent mid-wake). If busy: **defer** the install, show "Update ready — applies when your tasks finish or on quit," and auto-apply the moment everything goes idle (or on the next quit). Make defer-while-busy the default. Never kill running work to install.

## 2. Architecture (grounded in the verified map)

The good news: the busy signal already exists. `CronBusyGuard` (`src/process/services/cron/CronBusyGuard.ts`) is wired into EVERY agent manager (WCore/ACP/Gemini/Remote/NanoBot/OpenClaw) and, because team wakes and cron runs all funnel through those managers, it already spans chat + cron + team activity. One registry answers "is anything working right now."

```
Update downloaded (autoUpdaterService, autoDownload=false already)
        │
   user clicks Install  OR  app quit (autoInstallOnAppQuit=true)
        │
   updateQuiesceGate.isAppBusy()  ← reads CronBusyGuard.getAllStates()
        │
   busy? ── yes ──► defer: set 'deferred' status, show banner/modal state,
        │                  register onceAllIdle → then quitAndInstall
        └── no ───► quitAndInstall() as today
```

## 3. Components (with exact anchors)

1. **Busy aggregator (extend existing):** add `onceAllIdle(cb)` to `CronBusyGuard` — it already has per-conversation `onceIdle`/`idleCallbacks`; fire a global callback when the last processing conversation clears. `isAppBusy()` = `Array.from(cronBusyGuard.getAllStates().values()).some(s => s.isProcessing)`. This is the single source of truth (covers chat, cron, team wakes).
[...]

Issue #632, opened 2026-07-04
[Feature]: have a software update for Wayland/Wayand Core be blocked by running sessions


frakman:
### Feature Description

Prevent app restart for software update when agent responses are running

### Problem Statement

If you trigger an update and not realise a task/response is currently running it will kill that.  So you get somethng like this:

<img width="3070" height="1630" alt="Image" src="https://github.com/user-attachments/assets/ded3c643-55c8-4966-8a42-42c6877a5185" />

### Proposed Solution

When restarting due to an update:
1. check if there are chat sessions current responding.
2.  block the restart (prevent hassles for use)
3. optionally restart triggers when ALL current tasks stop responding  (most convenient for users)

### Feature Category

UI/UX Improvement

frakman:
Ok ways this could be done:
1. warn at the point you press the [update] button   (easiest because it is in app)
2. when it is downloaded there is a button to open folder where the file is, and a button to directly open the installer.  could block these buttons until there are no responding sessions  (should also be simple as it is in app)
3. update the app so it has a shutdown API/IPC call to tell it to shut down or restart. (would be the safest?)
    --- have that shutdown wait till it is safe
   
4. update the app so it has an API/IPC call to tell if it has a response currently underway.  Update the installer so it will not kill the app while the app is still returning stuff.  Optionally have the installer keep polling until it is safe to kill. 

The first two options are likely enough for 99.9% of the time

FerroxLabs:
This is now implemented and shipped — closing as completed.

**What was delivered (update-on-quiesce):** when you trigger an app update restart while a chat session is actively responding (or a scheduled/cron task or a standing team is working), Wayland no longer restarts out from under that work. The restart is **deferred** and applied automatically once everything goes idle, or on the next normal quit. The updater surfaces a "ready — applies when your tasks finish" state instead of interrupting.

**How it works**
- The busy signal is fed by every agent backend (WCore, ACP, Gemini, Remote, NanoBot, OpenClaw): each turn marks the app busy while it streams and clears it when the turn ends. Cron runs and team wakes funnel through the same managers, so one check answers "is anything working right now" across chat + cron + teams.
- Both the manual "Install now" button and the coordinated on-quit install route through this gate. The on-quit path also drains in-flight work (DB, cron, workers, team sessions, tunnels, servers, agent children) *before* applying the staged update, in the correct order.
[...]

Issue #575, opened 2026-07-03
[Bug]: After latest update, Wayland starts 10th of instances under MacOS


sparel:
### Platform

macOS (Apple Silicon)

### Wayland Version

0.11.11

### Bug Description

After updating Wayland and Wayland core to the latest versions, the application started creating new instances of itself and bogs the Dock bar

### Steps to Reproduce

Start Wayland under MacOS 26.5.1
New instances are constantly being started and bogging the Dock bar. It is no longer possible to work, since the Wayland application gets the focus back everytime

<img width="1984" height="1055" alt="Image" src="https://github.com/user-attachments/assets/43bfc6c3-5d87-4cb7-8b1d-90731642ffeb" />

### Expected Behavior

Not that what is described

### Actual Behavior

As described

FerroxLabs:
Fixed and shipped in Wayland desktop 0.11.12 (bundles wayland-core 0.12.21). Fix commit fa08bc2ad (verified ancestor of the release tag). Closing as resolved via Overwatch inventory sweep — reopen if it recurs.

sparel:
Still happening after update to 0.11.12 and latest wayland-core

PR #653, opened 2026-07-04
feat(update): #651 defer auto-update restart while sessions/tasks/teams are working

FerroxLabs:
Update-on-quiesce v1 (#632). Addresses #651.

An auto-update restart must never yank the rug out from under active work. Wayland runs autonomous agents, scheduled tasks, and standing teams, so it is server-like: like VS Code, Slack, and Chrome it should apply an update on idle or on quit, never force-restart mid-task. The real incident: a user updated mid-database-work with two active processes and file edits going, and the update-restart killed all of it.

## What it does

Before an update restart, check whether anything is actively working. If busy: defer the install, show "Update ready, applies when your tasks finish or on quit," and auto-apply the moment everything goes idle (or on the next quit). Defer-while-busy is the default. Running work is never killed to install.

## How it works

The busy signal already exists: CronBusyGuard is wired into every agent manager (WCore/ACP/Gemini/Remote/NanoBot/OpenClaw), and because team wakes and cron runs funnel through those managers it already spans chat + cron + team. One registry answers "is anything working right now."

- CronBusyGuard gains isAppBusy() and a one-shot onceAllIdle(cb) that fires when the last processing conversation clears (immediate if already idle, closing the busy-to-idle race).
- updateQuiesceGate.installOrDefer() defers while busy and registers the install for idle, else installs now. It reads update.deferWhileBusy (default true) directly in main and guards against double-defer on repeated clicks.
- autoUpdaterService disables electron-updater's own autoInstallOnAppQuit and drives the on-quit install explicitly via installOnQuitIfReady() (non-force, no relaunch) as the LAST step of before-quit, after work is drained. The #575/#286 loop guard moves to an explicit "installable on quit" flag so a doomed update is never applied.
- The manual IPC path routes through the gate; force bypasses it for "install now anyway". A settings toggle plus get/set providers back the setting.
- UpdateModal renders a deferred state with the override button. i18n across 12 locales.

## Acceptance criteria

Criteria 1-7 in the issue. The gate/service/aggregator logic is unit-covered (CronBusyGuard idle aggregator, quiesce gate defer/install/disabled/double-click, on-quit install safety reworked to installOnQuitIfReady honoring the block, deferred-modal DOM test). Live-verify evidence to follow on the PR.

## Safety note

The #575/#286 protection is preserved: a macOS block (app outside /Applications) or a known silent apply-failure keeps the update marked not-installable-on-quit, so installOnQuitIfReady refuses and never re-arms the relaunch loop. Verified by test.


FerroxLabs:
Overwatch ruling: FIX-FIRST (not merged). Independent cross-audit of the diff at head found a real bug the lane cross-audit missed, in the feature's own primary path.

THE BUG: the non-force install callback wired in src/process/bridge/updateBridge.ts (the quitAndInstall provider) passes autoUpdaterService.quitAndInstall as the install() callback to updateQuiesceGate.installOrDefer. Both the install-immediately-when-idle branch and the defer-then-idle branch therefore call autoUpdaterService.quitAndInstall(), which runs autoUpdater.quitAndInstall(true,true) and then unconditionally app.exit(0) after a 1s timer (no clearTimeout). That hard-kills the process 1 second later, bypassing the 12-step before-quit drain (DB, cron, workers, team sessions, channel manager, tunnel processes, websocket/http server, watchers, agent children) that is designed to take up to 10s. installOnQuitIfReady() — the method built to install as the LAST step after the drain — is ONLY called from the before-quit handler and is never reached by the gate's callback, the Install Now button, or the auto-fired idle install.

CONCRETE FAILURE: agent finishes a task while an update is staged+deferred, setProcessing(false) fires global-idle, the gate callback calls quitAndInstall(), app.exit(0) fires at t=1000ms regardless of whether the drain finished, orphaning tunnel/agent child processes and dropping unflushed writes. That is the #651 bug class re-created just after the busy flag clears.

THE FIX (precise): in updateBridge.ts non-force branch, change the install callback from autoUpdaterService.quitAndInstall() to Electron app.quit() (already imported), so both the immediate-when-idle and deferred-then-idle paths go through the existing before-quit drain and installOnQuitIfReady() in the correct order. Reserve the raw autoUpdaterService.quitAndInstall() hard-exit for the force:true Install-now-anyway override only (the one case the user explicitly accepted an abrupt interrupt). Add a unit test that the deferred-then-idle path routes through the drain (installOnQuitIfReady), not the hard exit.

CONFIRMED CLEAN (real, in the diff, not just claimed): the CronBusyGuard fireGlobalIdleIfIdle setImmediate re-check race fix; criterion-7 (defer-disabled proceeds while busy) implemented+tested; onceAllIdle registration race-free; double-install deferPending latch; #575/#286 relaunch-loop guard preserved; all agent managers funnel through cronBusyGuard.setProcessing.

Re-flag needs:overwatch when the one-line callback fix + the drain-path test are in and green; I will re-verify the diff at head and live-verify (start a long response, trigger Install, confirm it DEFERS then installs cleanly via the drain, no orphaned children) before merge.

Issue #656, opened 2026-07-04
Capabilities must not be silently gated — surface the real reason, and prompt for permission instead of refusing/failing


FerroxLabs:
Product principle, raised by Sean after a live-broadcast incident (see #628). Filing so it is tracked and acted on, not buried.

THE PRINCIPLE
When the agent cannot do something it should be capable of, it must never silently fail, silently refuse, or hallucinate a workaround. Two rules:
1. If a capability is blocked by environment/detection (e.g. it cannot see node on PATH), surface the real reason (I cannot see my PATH / this tool is unavailable because X) rather than inventing a fix or claiming the thing is not installed.
2. If an action is genuinely sensitive, ASK FOR PERMISSION (a prompt the user can approve) rather than hard-refusing it or failing opaquely. A gate should be a permission request, not a dead end.

WHY THIS MATTERS
This was the whole point of much of the recent trust/permission work. Two live failure modes both read to the user as the product being broken:
- PATH-blind detection failure (#628): agent said node was not installed and told the user to reinstall everything, on-broadcast, when it was all installed. Looked like a capability gate; was actually blindness.
- Denylist over-refusal (#618, Windows): the credential-exfil denylist refuses benign shell commands outright with no path forward.

SCOPE (design, then implement across both repos)
- Engine (wayland-core): the Bash tool denylist (crates/wcore-tools/src/bash.rs) currently HARD-REFUSES matched commands. Design a permission-prompt path so a borderline command surfaces to the host for user approval instead of a flat Refused. Detection failures (tool/binary not found) should return a truthful unavailable-because reason the model is instructed not to paper over.
- Desktop: a host-side approval affordance for engine-requested permissions; make sure refusals/unavailability render as an honest, actionable message, not a hallucinated remedy.

ACCEPTANCE
- No command that the user is entitled to run is silently refused; sensitive ones prompt for approval.
- When a binary/tool is genuinely absent or unreachable, the agent states that plainly and does not fabricate an install/repair the user does not need.
- Live-verified on a packaged build, not dev-only.

Related: #628 (PATH-blind detection), #618 (denylist over-refusal), and the trust-model work. This is the umbrella principle both roll up to.

FerroxLabs:
Deep audit complete (5 parallel auditors across engine sandbox/policy, tool refusals, swallowed errors, capability-availability, desktop permission flow). This principle now has a tracked, allocated child set. Two themes:

SILENT GATES (deny without asking):
[...]

PR #674, opened 2026-07-04
fix(permission): a denied approval reads as denied, not green success (#663)

### LINKED ISSUE #663: [Desktop] Trust-workspace axis + render approval_required HITL modal (audit, #656)
Part of the audit (rolls up to #656). The desktop ask-permission spine already works on the live V2 ACP path (PermissionResolver 3-tier: Autopilot auto-approve, cache auto-approve, else a real accept/deny card; the live path never silently auto-denies). Two gaps remain — the second half of the trust model.

1. P1 real dead-end: wcore approval_required HITL event is emitted but nothing renders it. src/process/agent/wcore/index.ts:544-563 emits type approval_required with a future renderer TODO; zero handlers exist in src/renderer or src/process/task. If the engine escalates a sensitive action to human approval, the desktop shows nothing and the turn hangs on a resume token no UI can supply. Fix: render an approval modal for approval_required and wire allow/deny to approveTool/denyTool (wcore/index.ts:732-738) via the resume token, mirroring the tool_request confirmation flow. VERIFY against remote main first (audit clone was ~1 month stale).
2. P2 design: no Trusted-workspace axis exists (grep trust in the agent/permission layer = none). The only non-prompting option is full per-conversation Autopilot. Build a persisted per-workspace trust level that raises the default mode (trusted workspace -> auto-approve reads/edits, still prompt on exec/network) — the middle ground between prompt-on-everything and full Autopilot. This is the desktop half of the engine trust-posture work (see #657); the engine already has WorkspacePolicy trusted_local/contained but the desktop has no UI to drive it.
3. P2 polish: persist allow-always across restart (today in-memory only). Fix the deny-shown-as-green-success banner (MessageAcpPermission.tsx:128-137 — branch allow vs deny).

ACCEPTANCE: an engine approval_required surfaces an allow/deny modal (no hang); a per-workspace trust setting exists and changes prompting; allow-always persists; a denial reads as denied. Live-verify on a packaged build. Owner: a desktop lane.

FerroxLabs:
Addresses #663 (P3 slice only — the deny-banner silent-lie). Does not close #663: the remaining scope is being re-organized per the Overwatch ruling on the issue (the P1 "render an approval modal" is a no-op — the existing tool_group Confirming gate already covers it; the #264 auto-mode wedge becomes escalate-through-that-gate; the Trusted-workspace axis is splitting into its own issue).

## The bug (a silent lie)
MessageAcpPermission rendered a green "Response sent successfully" banner on ANY response — so a DENIAL displayed as a green success. That is the exact fail/refuse-while-reporting-success class #656 exists to kill.

## Fix
Branch the outcome banner on the chosen option's kind. AcpPermissionOption.kind is one of allow_once / allow_always / reject_once / reject_always; a reject_* choice now renders a danger "Request denied" banner, while allow options keep the success banner. Renderer-only, no behavior change to the actual approve/deny decision that was already sent.

## Tests + checks
- 2 new DOM tests (real Arco radio interaction): choosing a reject option renders the denied banner (not success) and sends confirmKey=reject_once; choosing an allow option renders the success banner.
- typecheck clean, oxlint 0 errors, oxfmt clean, prek CI-parity green (incl. i18n Check + UI Tokens). i18n: the new messages.permissionDenied key was added + translated across all 12 locales by the i18n toolchain.

## Verification note (why this is only P3)
Verified #663 against current main (the audit clone was ~1 month stale): the P1 "approval_required renders nothing -> hang" is already handled (WCoreManager auto-resumes auto-mode/info; interactive is covered by the existing tool_group Confirming gate). So the modal is intentionally NOT built, and this PR is the one genuine renderer bug. Overwatch cross-audits diff-at-head + packaged live-verify (allow=success, deny=danger) and merges.

### DIFF
--- a/src/renderer/pages/conversation/Messages/acp/MessageAcpPermission.tsx
+++ b/src/renderer/pages/conversation/Messages/acp/MessageAcpPermission.tsx
@@ -52,6 +52,11 @@ const MessageAcpPermission: React.FC<MessageAcpPermissionProps> = React.memo(({ 
   const [isResponding, setIsResponding] = useState(false);
   const [hasResponded, setHasResponded] = useState(false);
 
+  // #663 P3: a denial must NOT read as a green success. Derive whether the
+  // chosen option was a reject (reject_once / reject_always) so the outcome
+  // banner reflects allowed vs denied.
+  const respondedDenied = (options.find((o) => o.optionId === selected)?.kind ?? '').startsWith('reject');
+
   const handleConfirm = async () => {
     if (hasResponded || !selected) return;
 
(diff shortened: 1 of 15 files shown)


FerroxLabs:
Duplicate of #670 (same P3 deny-banner fix, identical diff, pushed to a second branch). Consolidating on #670 — it already passed Overwatch cross-audit (SHIP) and is on the CI-green merge watch. Closing this one. Note for the lane: one PR per fix — the double-push is the kind of thing the instance-label + single-branch convention is meant to prevent.

PR #702, opened 2026-07-05
fix(diag): stop over-redacting lowercase identifiers in Doctor report (#701)

### LINKED ISSUE #701: Doctot Report
### What happened
I have been having a lot of issues with Wayland  stopping in the middle of a task, timing out, not being able to fetch tooled or find directories etc:
I ran a Doctor report and it found a bunch of errors.
Can you tell me how to fix these?

Doctor Report: 

Image

### Environment
- App: 0.11.13
- Engine: wayland-core 0.12.17
- OS: win32 x64 (10.0.26200)

### Diagnostics
**MCP servers** (mcp.config): 18
- `com.••••-mcp` — Enabled but exposes 0 tools — it likely failed to connect or registered nothing; check its command, args, or credentials.
- `com.supabase-supabase-mcp` — Streamable HTTP error: Error POSTing to endpoint: {"message":"Unauthorized"}
**Providers** (••••ders): 10
**Workspace** (projects + conversations): 15
- `can you find where we left off and get this projec` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `Neer to restart chat from this point, read the fil` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `I changed the chimichurri to chicken parmesian bec` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `AizenPost Product-First Build Strategy` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `AizenPost Product Build Strategy and Execution Plan` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `Building Private LLM Server Business Requirements` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
**Config paths**
- app: ~\AppData\Roaming\Wayland\config
- engine: ~\AppData\Roaming\wayland-core
**Recent errors** (~\AppData\Roaming\Wayland\logs)
```
2026-06-23.log: [2026-06-22 22:09:52.156] [error] (node:49800) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
2026-06-23.log: (Use `Wayland --trace-deprecation ...` to show where the warning was created)
2026-06-23.log: [2026-06-22 22:09:52.162] [error] (node:49800) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
2026-06-23.log: [2026-06-22 23:14:08.575] [warn]  [Sentry] DSN not set; telemetry disabled
2026-06-23.log: [2026-06-22 23:14:08.663] [warn]  [Wayland] Another instance is already running; current process will exit.
2026-06-23.log: [2026-06-22 23:14:08.886] [error] (node:3520) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
2026-06-23.log: (Use `Wayland --trace-deprecation ...` to show where the warning was created)
2026-06-23.log: [2026-06-22 23:14:08.893] [error] (node:3520) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
```

---

ShadowsTT:
## Problem (#701)

The Doctor / bug report ran its output through the concierge-diag secret masker, whose base64url catch-all rule matched **any** 24+ character `[A-Za-z0-9_-]` run — including legitimate all-lowercase identifiers. This garbled the report the user reads to figure out what's broken:

- `providers.source` is the literal constant `model_registry_providers` (exactly 24 chars) → rendered as `**Providers** (••••ders): 10`
- Long reverse-DNS MCP server names (e.g. `com.<longname>-mcp`) → rendered as `com.••••-mcp`, hiding the very server name the report exists to name.

## Root cause

`SHAPE_REGEXES` contained `/[A-Za-z0-9_-]{24,}={0,2}/g`. "Over-redaction is safe" holds for a machine-consumed tool, but **not** for a human-read diagnostics report — it destroys the diagnostic signal (section labels, server ids).

## Fix

Gate the base64url rule on a leading lookahead requiring at least one uppercase letter **or** digit in the run:

```
/(?=[A-Za-z0-9_-]*[A-Z0-9])[A-Za-z0-9_-]{24,}={0,2}/g
```

Real base64/hex/JWT tokens always carry that entropy and are still masked; all-lowercase snake/kebab identifiers are spared. Prefixed keys (`sk-`, `ghp_`, …), named secrets (`token=…`), URL userinfo passwords, and long hex runs are handled by separate rules and are unaffected.

Verified against the two reported cases plus JWT / AWS / base64-with-digits / stripe-style tokens: identifiers preserved, secrets still masked.

## Note on the other Doctor diagnostics

The remaining report lines (MCP servers exposing 0 tools, Supabase MCP `Unauthorized`, temp-workspace warnings, `punycode`/`DEP0190` deprecation warnings) are the diagnostics working **as intended** — they reflect the user's local config (unauthenticated/misconfigured MCP servers, chats opened outside a project). This PR fixes the reporting bug that made those lines harder to read; the underlying config guidance belongs in a support reply on the issue.

## Test

Adds `conciergeDiagServer.redact.bun.test.ts` locking in both directions (identifiers preserved, secrets masked).

Closes #701


FerroxLabs:
Thanks for this, and for the clear write-up. #701 is a real bug: the Doctor report genuinely shouldn't be turning model_registry_providers into dots or hiding the MCP server names it exists to surface. The direction is right. An independent security cross-audit did surface one issue that needs closing before this can land on a report surface, so holding it for a revision.

The problem: the new lookahead exempts any 24+ char all-lowercase, no-digit run from the base64url rule. That is safe for cryptographically random tokens (they virtually always carry an uppercase letter or digit, and lowercase hex is still covered by the separate 32+ hex rule). But it also stops masking bare lowercase-only secrets that appear in free-form text with no key name, no URL userinfo, and no leading colon/equals/at, which nothing else in the pipeline then catches.

Concrete failing case (masked before this PR, not masked after):

    https://hooks.example.com/trigger/abcdefghijklmnopqrstuvwx

Walking the rules: the key-value rule needs a recognized key name (token/secret/password/bearer) nearby, absent here. The URL-userinfo rules only fire on user:pass@host, but this is a path segment. The delimiter backstop only fires when the token is immediately preceded by colon, equals, or at, but here it is preceded by a slash. No prefix rule matches and it is not hex. So the 24-char lowercase slug slips every rule. Webhook relays, magic-link/invite URLs, and share links commonly use lowercase-only slugs and land verbatim in connection-failure text, which is exactly what a Doctor report is meant to be pasted publicly. The same gap applies to diceware-style passphrases appearing as bare tokens.

Recommended fix (either works; the first is cleaner and carries no security tradeoff):

1. Preferred: keep the internally generated labels out of the generic redact() pipeline at the call site. model_registry_providers and the MCP server ids / reverse-DNS names are a small, enumerable, non-secret set the diag server generates itself, so exempt those specifically rather than loosening the global heuristic that also runs over free-form error text and URLs where real secrets live.

2. If you would rather keep the regex heuristic: add a regression test proving the bypass class is still masked (redact of the hooks.example.com URL above must still mask the slug), and narrow the exemption so a bare lowercase run outside a recognized identifier / URL-userinfo / key-value context is still caught (only exempt the specific internal-label shapes #701 names).

Minor: the three still-masked cases in the new test all contain digits, so none actually exercise the lowercase-only backstop the code comment relies on. Worth adding a digit-free case.

Really appreciate the contribution. This is close, it just needs the leak vector closed before it can land on a surface that gets pasted into public reports.

FerroxLabs:
Rebased onto `main` and pushed one hardening commit (`3c4f00ed2`) — thanks for the fix, the diagnosis is right and the Doctor report was genuinely being garbled by its own catch-all.

One gap in the entropy lookahead, though. `(?=[A-Za-z0-9_-]*[A-Z0-9])` exempts **every** all-lowercase run — but an identifier is only lowercase-and-safe because a *separator* breaks it up. An unbroken lowercase run of 24+ is token-shaped, not name-shaped, and none of the other layers catch it:

- `KEY_VALUE_REGEX` needs a key name in front (`token: …`, `api_key=…`)
- `DELIM_TOKEN_REGEX` needs a `:`, `=` or `@` immediately before
- the hex rule needs 32+ chars

So a bare token in free text had no rule left. Concretely, on your branch as pushed:

```
redact('auth failed for zzzytqwerlkjhgfdsamnbvcxsw, retrying')
  before: auth failed for zzzytqwerlkjhgfdsamnbvcxsw, retrying   <-- in the clear
  after:  auth failed for ••••cxsw, retrying
```

That string is masked on `main` today, so this was a live regression in a redaction path rather than a hypothetical.

The added rule is `/\\b[a-z0-9]{24,}\\b/g`. `_` and `-` are outside the character class, so it cannot match a snake/kebab/reverse-DNS name — all three of your preserved identifiers stay byte-identical, including `model_registry_providers` at exactly 24 chars. Tests added alongside yours to pin both directions.

Verified the rule semantics against every assertion in the file (yours and mine, 11/11) in an isolated harness. I did not run the suite locally — this is a fork branch, so the real gate is CI. Auto-merge goes on once it's green.

FerroxLabs:
Second push (`2e10def91`) — the red on **Code Quality** and **Security Audit E2E** was not the regex, it was the test file's location:

```
conciergeDiagServer.redact.bun.test.ts(10,38): error TS2307: Cannot find module 'bun:test'
```

Both of those jobs run `tsc`. Every other `*.bun.test.ts` under `src/` is individually listed in `tsconfig`'s `exclude` array; this one wasn't, so tsc compiled it and couldn't resolve `bun:test`.

I could have just added it to the exclude list, but there's a second problem that fix would have hidden: **`.bun.test` is never referenced in `pr-checks.yml`**, so the bun suite doesn't gate PRs at all. The regression would have sat there enforcing nothing.

So I moved it to vitest at `tests/unit/process/resources/builtinMcp/conciergeDiagServer.redact.test.ts`, beside the existing `conciergeDiagServer.test.ts` which already imports `redact` from the same module. The bun carve-out is for code that genuinely needs the bun runtime (sqlite drivers, cost repos); a pure regex test needs none of it.

Same assertions, now actually enforced. Typecheck clean, 11/11 green locally.

Issue #740, opened 2026-07-06
[Bug]: Workflow can show complete while bge-m3 chat calls fail with provider 400


arc-web:
### Platform

macOS (Apple Silicon)

### Wayland Version

Observed in a current desktop setup; source reviewed against `main` / `v0.11.17`.

### Bug Description

A workflow can appear to finish successfully while the actual conversation contains repeated provider failures because the selected model is not chat-capable.

In the observed run, the Workflow view showed only the user prompt and the right rail showed `Workflow Complete` after roughly 40 minutes. The useful error was only visible after manually switching to Conversation view:

```text
Provider error: API error 400: {"error":{"message":"\"bge-m3:latest\" does not support chat","type":"invalid_request_error","param":null,"code":null}}
```

`bge-m3:latest` is an embedding/retrieval model, so it should not be selected or used for chat/workflow agent turns.

### Steps to Reproduce

1. Configure/use a provider that exposes `bge-m3:latest`.
2. Select or allow Wayland to use `bge-m3:latest` for a chat/workflow run.
3. Start a workflow.
4. Watch the Workflow view and completion rail.
5. Manually switch to Conversation view.

### Expected Behavior

- Non-chat models such as `bge-m3:latest` should be classified as embedding/retrieval models and excluded from chat/workflow model selection.
- If a provider rejects a workflow chat request, the Workflow view should surface the actionable provider error.
- A run with repeated provider failures should not present as clean success without visible error context.

### Actual Behavior

- The Workflow view can appear blank or unresponsive.
- The right rail can show a green `Workflow Complete` state.
- The actionable provider error is visible only in Conversation view.
- The same raw provider error repeats many times:

```text
"bge-m3:latest" does not support chat
```

### Additional Context

Static source review suggests two likely contributors:

- Embedding/retrieval model classification is conservative: `CatalogAssembler.looksLikeEmbedding()` currently checks for `embed` only, which misses `bge-m3`-style model IDs.
- Some capability rules are additive and can add `chat` via catch-all rules after embedding/rerank rules.
- Workflow completion UI currently renders the green completion card for `status === complete`; the workflow run driver treats `errored` as terminal for all-terminal completion, so workflow success/error presentation may need separate review.

Related but not identical:

- #682 covers opaque draft-generation failures.
- #389 covered a different Ollama capability mismatch: tool support vs chat support.
- #629 covers init/key recovery UI.

[...]

Issue #660, opened 2026-07-04
[Core] Honest capability availability — boot advisory + stop silent capability/image drops (audit, #656)


FerroxLabs:
Part of the silent-gate/silent-fail audit (rolls up to #656). Anti-pattern: a capability is unavailable and the agent silently drops it or invents a cause instead of honestly reporting why. Verified line numbers in wayland-core.

HIGHEST-LEVERAGE FIX (do this first): inject a boot-time unavailable-capabilities-and-why advisory into the agent system context. Doctor already computes this list out of band; surface the same to the model. This converts the auto-hide + the whole API-key column from silent to honest in one move.

FINDINGS:
1. P0 tool auto-hide is invisible to the model. crates/wcore-tools/src/registry.rs:114-120 drops any tool whose backend is unconfigured from the schema with only a dev-log. Gates vision, image_generation, transcription, tts, video_analyze, discord, google_meet, homeassistant, cron (bootstrap.rs:768,772,799,856,865,876). Asked to generate an image with no key, the agent fabricates I do not have image generation instead of set OPENAI_API_KEY. Keep the tool out of the call schema, but the boot advisory must tell the model what is unavailable and why.
2. P0 inbound channel image answered blind. channel_media.rs:139-143 (enrich) + channel_dispatch.rs:278-280: image arrives with no vision backend -> _ => continue -> prompt carries a bare URL the model cannot fetch -> confident answer to an unseen image. Fix: emit an honest degraded-mode turn notice (this assistant cannot view images, set a vision key).
3. P0 vision failure/size-bound silently drops image. channel_media.rs:197-225 (describe_image) returns None on backend error/timeout/size -> same blind path. Fix: propagate a visible degraded notice, not None.
4. P1 provider tool-drop honesty (fold in): tools silently dropped for tool-incapable models. bedrock.rs:186 (predicate 62-78) and openai.rs:722-727 (ToolSupportCache tool_capability.rs:81-83) omit the tools array and send anyway; agent thinks tools are live, model answers in prose. Fix: honest model cannot call tools signal instead of silent drop.
5. P2 structural: ModelInfo (model_catalog.rs) has no supports_vision/tools/streaming; capability is guessed from name prefixes + reactive 400s, which is WHY selection can only silently drop. Add capability fields + consult at selection; implement real message_requires_vision (engine.rs:139-141 hardwired false).

[...]

PR #795, opened 2026-07-11
fix(models): exclude family-named embedding models from chat selection

FerroxLabs:
Embedding/retrieval models whose ids are not spelled with the literal
'embed' (bge-m3, gte-large, e5-mistral, voyage-3, ...) were offered in
the primary/workflow model picker: excludeFromPrimary only matched
'embed'/'rerank'/'image', so their capability came back undefined and
the picker included them. Selecting one for a workflow turn then failed
with a provider 400 ('does not support chat') while the Workflow view
still showed a green Complete.

Make excludeFromPrimary a superset of the embedding + rerank family
patterns (extracted to a shared, exported EMBEDDING_MODEL regex), and
reuse that same pattern in the models.dev CatalogAssembler so the
catalog kind/tags stay consistent with the picker. Fixes the selection
root cause in #740; the 'workflow shows complete despite failures'
presentation is tracked separately.

### DIFF
--- a/src/common/utils/modelCapabilities.ts
+++ b/src/common/utils/modelCapabilities.ts
@@ -20,6 +20,18 @@ import { isFluxModelId } from '@/common/config/flux';
 const FLUX_IMAGE_MODEL = /flux(?:[.-]?\d|-(?:dev|schnell|pro|kontext|realism|lora))/i;
 
 /**
+ * Embedding / retrieval model families. Many are NOT named with the literal
+ * `embed` (`bge-m3`, `gte-large`, `e5-mistral`, `voyage-3`, …), so the known
+ * families are matched by name. Exported so other classifiers (e.g. the
+ * models.dev catalog assembler) stay consistent with this one. See #740.
+ */
+export const EMBEDDING_MODEL =
+  /(?:^text-|embed|bge-|e5-|LLM2Vec|retrieval|uae-|gte-|jina-clip|jina-embeddings|voyage-)/i;
+
+/** Reranker / retriever models (cross-encoders) - never chat models. */
+const RERANK_MODEL = /(?:rerank|re-rank|re-ranker|re-ranking|retrieval|retriever)/i;
+
+/**
  * Capability matching regex patterns
  */
 export const CAPABILITY_PATTERNS: Record<ModelType, RegExp> = {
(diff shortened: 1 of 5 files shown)


FerroxLabs:
⚠️ **HOLD — do not merge. Collision / coordination in progress.**

I opened this without first spotting the pre-existing external PR #742 (@arc-web) for the same issue (#740) and Session B's active claim. Converting to draft and deferring.

**Why this isn't simply a duplicate of #742 — the two are largely complementary:**
- **#742** fixes the *catalog / detector* classification side: `CatalogAssembler`, `ModelCapabilityDetector`, and a new `modelCapabilityRules.ts` (`looksLikeEmbeddingModelId`).
- **This PR (#795)** fixes the *primary/workflow picker gate*: `src/common/utils/modelCapabilities.ts` `excludeFromPrimary`. The guid/workflow picker (`getAvailableModels` in `src/renderer/pages/guid/utils/modelUtils.ts`) includes a model iff `(function_calling===true||undefined) && excludeFromPrimary!==true`. For `bge-m3:latest`, `excludeFromPrimary` returned `undefined` (the pattern only matched `embed|rerank|image`), so it was **offered in the picker** — the actual selection root cause. **#742 does not touch `modelCapabilities.ts`**, so on its own it may classify bge-m3 as embedding in the catalog while the picker still offers it.

**Overlap:** both edit `CatalogAssembler.looksLikeEmbedding` — mergeable, but needs reconciliation.

**Recommendation (for Session B / maintainers):** prefer @arc-web's #742 (filed first, external contributor) for the catalog/detector side; fold in the `excludeFromPrimary` picker-gate fix from this PR (the part #742 lacks). I'll defer to whatever the owning session decides — this branch is just here so the picker-gate finding isn't lost.

FerroxLabs:
Closing in favor of **#797** (@B's hardened picker-gate PR) — the canonical fix for the `excludeFromPrimary` selection gate this draft held, now with token-boundary hardening (kuae/text-davinci guards) and scoped to `modelCapabilities.ts` only (no CatalogAssembler overlap with arc-web's catalog-side #742).

Recap of the 3-piece #740 fix: **#742** (catalog/detector) + **#797** (picker gate) + #660 (the "workflow shows Complete despite 400" display half, core lane). This #795 draft was Session A's early superset attempt; superseded and closed per coordination. No code lost — the picker-gate change lives in #797.

Issue #806, opened 2026-07-11
CI: #721 spawn-test settle loop hangs on sharded runners, blocking the merge train


FerroxLabs:
## Impact
Currently blocking **PR #742** (windows-2022 3/4) and **PR #805** (ubuntu-latest 3/4). Same test, two different PRs, two different OSes. Any PR unlucky enough to land on shard 3/4 gets a red required check.

## The test
`tests/unit/process/services/ijfwSystemService.applyPendingUpgrade.test.ts`
> `#721: spawn-test settles false via its 5s timeout when the child emits only garbage`

```
Error: Test timed out in 30000ms.
  ❯ tests/unit/process/services/ijfwSystemService.applyPendingUpgrade.test.ts:303:3
```

## This is a HANG, not slowness — and the previous fix misdiagnosed it
`fc6cea0b1` ("test: give the #721 garbage-child settle loop a 30s budget", #775) raised the budget from 10s to 30s on the theory that loaded CI shards were legitimately slow. It is now blowing through the raised 30s budget too, which is the tell that the premise was wrong.

Measured on this branch:

| where | result |
|---|---|
| the test file alone, locally (3 runs) | **passes in 5–7ms** |
| whole `--shard=3/4` locally (3042 tests) | **passes in 11ms** |
| CI shard 3/4 (ubuntu + windows) | **times out at 30000ms** |

5ms versus >30s is ~6000×. Contention does not produce that. The promise is not settling on CI at all, so the settle loop burns all 200 rounds and then `await run` blocks forever.

## Why the shape is fragile
```js
vi.useFakeTimers({ toFake: ['setTimeout', 'clearTimeout'] });   // setImmediate stays REAL
const run = ijfwSystemService.applyPendingUpgrade().finally(() => { settled = true; });
for (let i = 0; i < 200 && !settled; i++) {
  for (let j = 0; j < 4; j++) await flush();     // real macrotasks
  await vi.advanceTimersByTimeAsync(500);        // fake clock
}
await run;   // <-- hangs forever if the loop exhausted without settling
```
It hand-interleaves a **real** macrotask queue (the child emits via `setImmediate`, fs is real) with a **fake** clock. Whether it settles depends on the real queue draining in the exact number of `flush()` calls budgeted per round. That is a race, and the sharded runner loses it.

`await run` with no timeout of its own is what converts a lost race into a 30s hang rather than a fast, legible failure.

## Suggested direction (not yet attempted)
[...]

Issue #721, opened 2026-07-05
[Bug]: one non-JSON stdout line SIGTERM-kills the ijfw MCP child — tolerate garbage NDJSON lines instead of killing the connection


FerroxLabs:
## Summary
One non-JSON stdout line from the ijfw MCP child kills the whole connection: mcpWireProtocol.decode() throws DecodeError on any well-terminated line that fails JSON.parse (src/process/services/ijfw/mcpWireProtocol.ts:74-79), and ijfwMcpClient.onStdout() responds by SIGTERM-killing the child and entering degraded mode (ijfwMcpClient.ts:349-359). Customer log shows the loop firing 8x in one day: decode error - killing child { err: invalid JSON line: Unexpected token 'b', "build.buil"... } followed by child exit { code: null, signal: SIGTERM }.

Verified attribution: the polluting line comes from the child's own stdout (dedicated pipes at ijfwMcpClient.ts:326 make cross-stream pollution impossible; Wayland's buildStorage emits no log output — the auditor's attribution was wrong). The customer's ijfw version/extension console.logs a build.* progress line to stdout, violating MCP stdio framing. But our reaction is the defect: NDJSON is self-synchronizing at newlines, so a garbage line cannot desync the protocol — killing the child is unnecessary and turns third-party log noise into a connection outage.

## Fix
1. Tolerant framing: in decode(), skip unparseable well-terminated lines (return droppedLines count + truncated samples instead of throwing); onStdout logs them at warn. This matches the official MCP SDK stdio transport behavior (log-and-skip).
2. Keep the kill/quarantine path ONLY for genuine resource-abuse DecodeErrors (line > MAX_LINE_BYTES, buffer > MAX_BUFFER_SIZE).
3. Unit test: interleave a plaintext build.* line between two valid JSON-RPC responses; assert both responses resolve and the child is NOT killed.
4. Report upstream to the ijfw plugin that their server (customer's version) writes progress lines to stdout; diagnostics must go to stderr.

## Evidence
Customer audit PDF p.1 + section 3.2 (exact strings trace to mcpWireProtocol.ts:78, ijfwMcpClient.ts:351/:353/:337).

FerroxLabs:
[...]

PR #810, opened 2026-07-11
fix(806): stop CI merge-train hang from spawn-verify settle test

FerroxLabs:
The #721 spawn-verify settle test faked the clock and interleaved real fs
macrotasks with fake-clock advances over 200 rounds; settlement depended on
how many real event-loop turns the OS granted per round, which starves under
parallel CI shard load and hangs (>30s false-red blocking the merge train).
Raising the wall budget twice did not help — it is a race, not slowness.

Add an IJFW_SPAWN_VERIFY_TIMEOUT_MS env seam (default 5000, mirroring the
existing IJFW_AUTO_INSTALL tunable) and rewrite the test to run with real
timers at a 30ms timeout. The garbage-only child never yields a valid reply
or exit, so the real timeout is the sole settle path and the flow resolves
deterministically in real time — no fake/real interleaving.

### DIFF
--- a/src/process/services/ijfwSystemService.ts
+++ b/src/process/services/ijfwSystemService.ts
@@ -575,6 +575,16 @@ async function retryOnEbusy<T>(op: () => Promise<T>, attempts = 5): Promise<T> {
   throw lastErr instanceof Error ? lastErr : new Error('EBUSY retry exhausted');
 }
 
+/**
+ * Verify-probe timeout. Defaults to 5s in prod; overridable via env (same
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Superseded by #813, which merged at 12:49Z and closed #806. #813 is the adversarially-reviewed, mutation-pinned version (silent-child mutant killed; windows-2022 3/4 + ubuntu 3/4 shards pass live). This branch conflicts against the landed fix and is redundant. Closing to clear the merge queue; reopen if a gap surfaces. — Overwatch

Issue #774, opened 2026-07-09
[Bug]: Agent stops permanently on any mid-run error instead of recovering/resuming


JimOnemi:
### Platform

Windows

### Wayland Version

v0.11.17

### Bug Description

When any error occurs mid-task, the agent halts and never resumes. It does not recognize that it failed, does not retry or work around the error, and does not attempt to continue the task. Observed with multiple error types:

Connection lost: Failed to generate content: Connection error.
Failed to generate content: Connection error.
API Error: 400 Flux.BadRequestError: DeepseekException – {"error":{"message":"Failed to deserialize the JSON body into the target type: messages[31]: missing field 'tool_call_id' ..."}}

In every case, if the user isn't watching, the task sits dead until someone manually prompts "retry" or "move past it and keep going" — at which point the agent resumes fine. Session state is intact; the only missing piece is automatic recovery.

<img width="825" height="400" alt="Image" src="https://github.com/user-attachments/assets/95497cc8-6f74-4c34-ae81-cd08f6651aa5" />
<img width="825" height="307" alt="Image" src="https://github.com/user-attachments/assets/61cdc43b-5752-49ec-b54d-817207b8d627" />

<img width="825" height="307" alt="Image" src="https://github.com/user-attachments/assets/f878bf13-8f25-42b5-8be5-c71ecb495226" />

### Steps to Reproduce

Start a multi-step task.
Trigger any mid-run failure: transient network drop, or a provider-side 400 (e.g., malformed request with missing tool_call_id).
Error banner appears; agent stops entirely with no retry or notification.

### Expected Behavior

Transient errors (connection, timeouts, 5xx): auto-retry with backoff, then resume the task.
Request errors (4xx like the tool_call_id deserialization failure): detect, repair or drop the malformed message, and continue — the agent can already do this when manually told to "keep going."
If recovery is impossible after capped attempts, notify the user clearly rather than idling silently.

### Actual Behavior

Agent stops on the first error of any kind.
No retry, no self-correction, no failure notification.
A manual "retry" or "keep going" prompt resumes the task successfully, confirming recovery is feasible.

PR #814, opened 2026-07-11
fix(acp): recover from a mid-run error instead of halting the agent

FerroxLabs:
## The bug

Any mid-run error killed the turn outright. The banner went up, the agent stopped, and the task **sat dead until a human typed "retry"** — at which point it resumed fine. That last part is the whole diagnosis: session state was always intact, and the only missing piece was automatic recovery.

Underneath, `PromptExecutor.handlePromptError` set the session back to `active`, emitted a banner, flushed *other* queued prompts — and **dropped the failed one on the floor**. The `AUTH_REQUIRED` branch three lines above already did the right thing (`pendingPrompts.unshift(content)`).

## The fix

`execute()` now wraps the prompt in a bounded retry (3 attempts, exponential backoff), reusing the existing `computeBackoff`/`sleepWithAbort`.

The retry runs **inside** the awaited turn, deliberately. `AcpAgentManager.sendAgentMessageWithFinishFallback` awaits `sendMessage`, and on a rejection it paints a turn-error banner and synthesizes a `finish` that releases the loading state. Retrying inside keeps the turn in flight, so a blip we recover from never flashes a spurious "turn failed". No manager changes.

## What it will NOT replay, and why

This is the part that matters. Replaying a prompt re-asks the model to carry out the request, so it is only safe under narrow conditions.

- **Only a failure the LIVE agent itself answered** (`-32603` — where ACP bridges dump the provider's error, including this issue's own `Failed to generate content: Connection error`).
- **Never stream-death** (`PROCESS_CRASHED`, `CONNECTION_FAILED`). If the pipe broke we cannot know what the agent already *did* — a `tool_call` notification it wrote can be lost with the pipe. It is also **already owned**: `onDisconnect` → `resumeFromDisconnect()` respawns the agent and re-flushes the pending prompt itself. Two recovery mechanisms racing over one session is worse than either alone.
- **Never a turn that ran a tool.** A file written, a command shelled out — replaying could run it twice. A human typing "keep going" is making that call with their eyes open; we are not.
- **Only genuinely transient detail.** The replay decision is an **allowlist**, not a denylist: a denylist over the `-32603` catch-all leaks by construction. Unknown fails **closed**, which just leaves the user where they were before this issue.

## Known limit (stated, not hidden)

The no-double-execution guarantee rests on backends reporting tool activity before failing — via a `tool_call` update, a permission request, or `fs/write_text_file`. **ACP says SHOULD, not MUST.** An agent that runs a tool and reports nothing could still have that turn replayed.

Truly *resuming* a tool-bearing turn (rather than restarting it) needs the engine to expose a resume primitive. That is **#457** (`needs:core`), and it is not something the desktop can fake safely. This PR does not pretend otherwise.

## Review

Three rounds of adversarial review, each of which found a real defect — including in my fixes to its previous findings:

1. The retry bound itself to `lifecycle.client`, a **live getter**. On a crash, `resumeFromDisconnect` *synchronously* spawns a replacement before its first `await`, so an "is there still a client?" check passed against a half-initialized one and fired the prompt into a session it had never loaded. My tests were also **hollow** — 4 of 5 guards deleted clean, and the cancel test passed via a TypeError path that never touched cancellation.
2. I fixed that race in the retry loop and left **the same race standing in `handlePromptError`** (`PROCESS_CRASHED` is `retryable: true`, and `resuming → active` is a legal transition). And the rate-limit regex ended in `\b` — `_` is a word character, so `rate_limit_error`, `insufficient_quota`, `context_length_exceeded` **all sailed through and got replayed**.
3. The post-backoff guard didn't re-read `turnRanTool`, so the safety property rested on SDK dispatch ordering rather than holding by construction; and the ownership guard sat below the `AUTH_REQUIRED` branch, the last path still racing the respawn for the queue.

## Verification

- **437 tests green** across the ACP + task suites; typecheck clean.
- Every guard is **pinned by mutation testing** — 8/8 killed: the `turnRanTool` re-check, the ownership guard, the transient allowlist, the `AbortSignal`, `PROCESS_CRASHED` replay, and all three `noteToolActivity` wiring points. (The 9th, dropping `cancelAll`'s abort, is an *equivalent* mutant — `cancelAll` calls `cancel()`, which already aborts.)
- Two of the new tests had been passing **for the wrong reason** (one threw inside `assertPathAllowed`; one never entered the sleep). Both were caught by mutation and fixed.
- A 23-string corpus of real provider messages, transient vs final.

Refs #774. The mid-run *resume* half remains #457 (engine lane).


FerroxLabs:
⚠️ Not auto-landing — real **Code Quality (oxfmt) failure**, not a flake or BEHIND. Oxlint passes; oxfmt fails on 4 unformatted files in this diff:
```
oxfmt src/process/acp/session/AcpSession.ts src/process/acp/session/PromptExecutor.ts tests/unit/process/acp/session/PromptExecutor.retry.test.ts tests/integration/process/acp/session/AcpSession.prompt.test.ts
```
Run that (auto-fix), commit, push — then it's green + up-to-date and auto-merge fires. (Flagged by desktop [A] per Overwatch monitoring ask; this is C4/#774's lane so leaving the 1-line format fix to the author.)

Issue #729, opened 2026-07-05
[Feature]: Use a green status badge for active/enabled profiles instead of amber


dariobros:
### Feature Description

The current **"Active"** badge for the selected profile is displayed in amber/orange. While it matches Wayland's branding, amber is commonly associated with a warning, pending state, or something that requires attention.

Since the profile is fully active and enabled, a green status indicator would communicate the state more clearly and align with common UI conventions.

### Problem Statement

The current color can create a small moment of hesitation because many users subconsciously associate:

- 🟠 Amber = pending, paused, warning, or not fully active

- 🟢 Green = active, enabled, healthy, running

For a quick visual scan, green provides immediate confirmation that this is the currently active profile.

### Proposed Solution

Change the **"Active"** badge color from amber/orange to green.

Alternatively:

- Keep the orange accent for branding throughout the UI.

- Reserve green specifically for positive status indicators such as:

  - Active

  - Enabled

  - Connected

  - Running

  - Healthy

This would improve visual consistency with common desktop application design patterns.

---

### Feature Category

UI/UX Improvement

### Additional Context

The screenshot below shows the current "Active" badge using an amber/orange color.

While visually attractive, it resembles a warning or pending state more than a confirmed active state.

Using green for status badges would make profile status instantly recognizable with almost no cognitive effort and better match user expectations across modern applications.

<img width="908" height="437" alt="Image" src="https://github.com/user-attachments/assets/da5b566f-43e0-40e5-96c9-f9ab6cb8974f" />

FerroxLabs:
Fixed by #837 (merged): status badges now use the success-green semantic token for active/enabled state instead of the brand accent/amber. Packaged-Windows confirmation rides A's release-readiness pass.

PR #837, opened 2026-07-12
fix(ui): status badges use success green, never the brand accent (#729)

FerroxLabs:
The profile 'Active' badge was painted with --wc-accent (#ff6b35), the brand
colour. Orange reads as a warning or a pending state, so a profile that was
fully active looked like it needed attention — the hesitation the reporter
described.

The badge family already had the right answer next door: .badge.connected (the
runtime 'Running' pill) was green via --wc-success. Two positive states, two
different colours, one primitive.

Collapse both onto a single semantic .badge.ok and write the rule down: green =
active/enabled/connected/running, amber = pending, red = error; the brand accent
marks SELECTION (nav rail, focus ring) and never health. That carve-out matters —
the sidebar correctly uses the accent for the selected item, and a blanket
'active is green' rule would have repainted the whole nav rail.

Pins the chain, not just the class name: the badge takes .ok, .ok resolves to the
success token, and no badge variant may reference the accent. Repainting .ok
orange while keeping its name fails the suite. ProfilesPane had no test at all.


FerroxLabs:
**Independent cross-audit (lane A) — ✅ clean, no blocking/should-fix.**

Verified against the real files: all WCore status badges now converge on the `--wc-success` semantic token (the only accent-painted status badge, `.badge.activeBadge`, is deleted; both `ProfilesPane.tsx:131` and `RuntimePane.tsx:138` repoint to `.badge .ok`). Non-status accent usages (segmented selection, primary btn, switch-on, rail-active) are correctly preserved — the "accent = selection, never health" carve-out matches reality. Token discipline correct (semantic, not hex), mode-invariant per the existing convention, and the DOM test is non-hollow (pill body asserted separately + a CSS-scan guard that no `.badge.*` variant references `--wc-accent`).

Two NITs, optional: (1) the `activeBadge` "re-introducing it turns this red" comment slightly overclaims — the CSS rule is deleted so a re-added TSX class resolves to `undefined`; the real guard is the CSS-scan test (robust). (2) Out-of-scope/pre-existing: `.engineChip.stopped` uses a hardcoded `#f87171` instead of the existing `--wc-error` token — worth a follow-up, not this PR. LGTM.

Issue #842, opened 2026-07-12
A workflow parked on 'awaiting_input' notifies nobody


FerroxLabs:
Split out of #579 (task-completion notifications), where it is a deliberate, documented false negative rather than an oversight.

## The gap

`taskCompletionNotifier` excludes any conversation driven by a workflow whose `status === "active"`, because a workflow advances its parent conversation with a **hidden directive per step** — so notifying on those turns would raise a banner per step.

But `pause()` and `setRunMode()` in `WorkflowSessionService` mutate **`run_mode` only**; `status` stays `"active"`. So a workflow parked on `run_mode === "awaiting_input"` — i.e. **blocked pending a human action** — is also silenced.

That is precisely the moment the user most needs telling, and it is the exact scenario #579 exists for ("I would not have to keep checking manually"). A workflow can currently sit blocked on the user indefinitely, in silence.

## Why it was not fixed in #579

It is a **false negative, not spam**. Getting it wrong in the other direction (notifying per step) is the failure mode #579 was explicitly hardened against, so the conservative gate shipped first. This is strictly an improvement on top, not a regression: before #579 there were no notifications at all.

## Suggested fix

Let `run_mode === "awaiting_input"` through the gate with its own distinct copy — "Workflow needs your input" rather than "Task complete" — while continuing to suppress the per-step `running` turns. Needs a check that `awaiting_input` is entered once per park, not re-emitted per driver tick, or it reintroduces the storm.

Relevant: `WorkflowSessionService.ts` (`pause`, `setRunMode`, and the `awaiting_input` transitions), `taskCompletionNotifier.isUserFacingConversation`.

FerroxLabs:
Verified shipped in **Desktop v0.12.0**.

**Evidence:** b3cd0511a (PR #925) — behind, ahead 0, behind 19

**Detail:** taskCompletionNotifier.ts:129 adds the run_mode === 'awaiting_input' exception and :200 emits distinct "needs your input" copy. Both cite (#842). Absent at v0.11.18 — again supersedes PR #843's own "not fixed" note.

Ancestry was proven against the released v0.12.0 tag using the GitHub compare API (a commit is in the release only when status is behind or identical AND ahead_by is 0), and corroborated by reading the code at the tag rather than trusting the timeline alone.

Reopen if the symptom recurs on 0.12.0 or later.

Issue #838, opened 2026-07-12
Gemini/OpenClaw/NanoBot/Remote never emit turn-completion (starves notifier + workflow driver)


FerroxLabs:
Found while wiring #579 (task-completion notifications). Filing rather than folding in, because the blast radius is bigger than notifications.

## The gap

`ConversationTurnCompletionService.notifyPotentialCompletion()` is the canonical "this turn is over" signal. It is called by exactly **two** of the agent managers:

- `src/process/task/AcpAgentManager.ts:438` (success) and `:1794` (error)
- `src/process/task/WCoreManager.ts:812` via `handleTurnEnd()` at `:1444`

These four never call it — they only emit `finish` / `error` on the responseStream:

- `src/process/task/GeminiAgentManager.ts:882`, `:971`
- `src/process/task/OpenClawAgentManager.ts:165`, `:174`
- `src/process/task/NanoBotAgentManager.ts:93`, `:101`
- `src/process/task/RemoteAgentManager.ts:134`, `:141`

## Why it matters twice over

`conversation.turn.completed` has **two** main-process consumers, and both are starved for those backends:

1. **Task-completion notifications (#579).** A Gemini user who switches on "Notify when agent finishes" gets nothing, silently. The toggle appears to work.
2. **The autonomous workflow driver** — `src/process/utils/initBridge.ts:268` calls `recordAutonomousCompletion` / `applyStepTransition` off this same event. If a workflow step is dispatched to a Gemini/Remote worker, the completion may never be recorded and the step may never transition. That is the more serious half and should be verified first: it is either a live workflow bug or those backends are simply never used as workers.

## Suggested fix

Add `notifyPotentialCompletion` at each terminal emit site so the event means the same thing for every backend. Note this is **not purely additive** — it will start delivering completions to the workflow driver that it never received before, so it needs checking against `recordAutonomousCompletion` semantics rather than being dropped in.

The 1s per-conversation dedupe in `ConversationTurnCompletionService:24` should make the extra calls safe from double-fire.

FerroxLabs:
Verified shipped in **Desktop v0.12.0**.

**Evidence:** c6c50fba8 — behind, ahead 0, behind 18

[...]

Issue #579, opened 2026-07-03
[Feature]: Notification upon task completion


Orrlog:
### Feature Description

I would like to suggest an optional completion sound for Wayland.

When the app finishes a task, it would be helpful if it could play a small ding, bell, or notification sound. This would be especially useful for longer tasks where the user may not want to sit and watch the screen the entire time.

For example, if I submit a request and then switch back to my browser or work on something else, I may not realize the task has already finished. On longer tasks, I also end up going back and forth repeatedly to check whether it is done, only to find that it is still running. If I knew the app would play a sound when the task was finished, I would not have to keep checking manually.

An optional sound alert would let the user know when the most recent request or task is finished, without needing to constantly monitor the page. Ideally, this could be turned on or off in settings, so users who do not want audio alerts can disable it.

### Problem Statement

It solves the problem of users having to constantly monitor the screen while Wayland is processing a task.

### Proposed Solution

I would like Wayland to add an optional completion sound that plays when the latest task or request finishes processing.

The sound could be a simple ding, bell, chime, or other subtle notification tone. The main purpose is to let the user know that Wayland has completed the current task without requiring them to keep watching the screen or repeatedly switching back to check.

Ideally, this would be controlled by a setting, such as “Play sound when task completes,” so users can turn it on or off based on their preference. It would also be helpful if the sound only played when a task fully completes, not during every small status update.

### Feature Category

UI/UX Improvement

### Additional Context

This would be especially useful for longer-running requests where the user may be working in another browser tab, checking email, or doing something else on the computer while Wayland is processing.

Right now, the user either has to keep watching the Wayland screen or keep switching back to check whether the task is finished. If it is still running, that check was wasted. If it has already finished, the user may not realize it right away.

A small optional completion sound would make the app feel more practical during longer tasks, especially for users who multitask while waiting for Wayland to finish.

TradeCanyon:
Addressed by #843 (task-completion OS notification) and the #854 follow-up (both merged): quiet-hours default persistence, per-chat focus gate, secret redaction, and i18n. Packaged desktop confirmation rides A's release-readiness pass.

FerroxLabs:
Addressed by #843 (task-completion OS notification) and the #854 follow-up (both merged): quiet-hours default persistence, per-chat focus gate, secret redaction, and i18n. Packaged desktop confirmation rides A's release-readiness pass.

PR #843, opened 2026-07-12
feat(notifications): notify on task completion, sound optional (#579)

FerroxLabs:
The Notifications settings page already shipped four switches — agentFinished,
agentError, playSound and quietHours — all defaulted ON. Nothing in the main
process ever read any of them. Flipping them did nothing, and the "coming soon"
string that would have admitted this is never rendered. The page was not an
unfinished feature; it was a silent lie.

The reporter asked for a sound. A sound alone cannot say which task finished or
whether it failed, so the notification is the feature and the sound is the option.

The hard part is NOT notifying. conversation.turn.completed is a TURN event, not
a TASK event, and a naive listener is a spam cannon:

- ai_waiting_input also fires while the agent is parked on a tool/permission
  confirmation. WorkflowSessionService already states it: a non-zero
  pendingConfirmations means the step is NOT complete. Announcing "Task complete"
  there is both a lie and one banner per tool call.
- The upstream dedupe is keyed per CONVERSATION, so it does nothing about fan-out
  ACROSS conversations. A team round is one conversation per slot; an autonomous
  workflow is one child conversation per step. Both would banner each.
- A workflow's parent conversation is advanced by a hidden directive per step.
- Cron turns complete like any other, and CronService ALREADY notifies for those,
  so we would double-notify.
- Channel-inbound turns (Telegram/Lark/DingTalk/WeChat/iMessage) beat every one of
  the gates above: channels AUTO-APPROVE tool permissions, so pendingConfirmations
  is always 0, and the app is unfocused by construction because the user is on
  their phone — reading the very reply we would be announcing. Excluded by an
  allowlist on source, because ConversationSource is an open union and a new
  channel must not be able to opt itself in.

So the gate is positive: only a conversation the user is driving at the desk, that
has genuinely come to rest, earns a banner. Each exclusion is mutation-verified —
delete any one and the suite goes red.

It also fails CLOSED when the conversation cannot be read. SqliteConversationRepository
returns undefined for BOTH "no such row" and "the read failed" — it never throws —
and every exclusion above reads from that one object, so treating a failed read as
an ordinary chat would drop them ALL at once and restore the storm, under exactly
the DB pressure a long workflow creates. A missed banner is a nuisance; a banner
storm is the bug.

Quiet hours wrap midnight (the shipped default is 22:00-07:00). A naive
start<=now<end test is not merely wrong there, it is exactly inverted — silent
all day, loud all night.

The banner names the CONVERSATION, not the model: event.model.name falls back to
the backend string, so the old copy read "acp finished and is waiting for you".

Focus is checked across ALL windows — a popped-out conversation is its own
BrowserWindow, and its owner was being told about the turn they were watching.

Also fixes a dropped parameter: ElectronPlatformServices accepted an "icon" and
never passed it to Notification, so the app icon resolved by notificationBridge
was computed and thrown away.

Coverage is honest and bounded: only AcpAgentManager and WCoreManager emit turn
completion, so Gemini/OpenClaw/NanoBot/Remote will not notify (#838). A workflow
parked on run_mode 'awaiting_input' also stays silent today (#842) — a false
negative, not spam.


FerroxLabs:
**Independent cross-audit (lane A) — no BLOCKING; the hard part is well done, but 4 SHOULD-FIX before merge.**

Solid foundation (verified): correct main-process wiring (`initTaskCompletionNotifier` on `conversation.turnCompleted`, the same emitter the workflow driver uses), all completion paths covered (finish `ai_waiting_input` / error `error`, `pendingConfirmations>0` correctly excluded), no cron double-notify (`runtime.hasTask`/`cronJobId` gated — CronService owns those), genuinely fail-closed (getConversation returns undefined, never throws → no storm), correct quiet-hours midnight-wrap, sound on-by-default consistent with settings, and behavior-asserting tests (not `.not.toBe`).

**SHOULD-FIX:**
1. **Quiet-hours default never takes effect.** Settings *displays* 22:00–07:00 but `notifications.quietHours` is only persisted when a time field is edited (`persistQuiet`); the generic `persist` never writes it. Fresh install → key `undefined` → `resolveSilent` returns `false` → a user who never re-types the times still gets sound at 2am, contradicting the UI and the PR's own claim. Apply the 22:00–07:00 default in the main process when the key is absent (or persist on first load).
2. **Focus gate is app-wide, missing the primary #579 case.** `isAppFocused()` = any focused BrowserWindow → suppresses the banner. Start a long task in chat A, switch to chat B *in the same app*; A finishes → no banner, even though the user isn't watching A. Gate on whether the focused window is actually showing `event.sessionId`, not app-wide focus (at minimum call out the tradeoff).
3. **i18n en-US only.** The other 12 locales get English fallback for "Task complete/failed". Strings are keyed + interpolated correctly — just untranslated.
4. **Error body leaks the raw engine string** (`truncate(event.detail)`) to the OS notification center, which macOS persists in history — can contain absolute paths / URLs with tokens. Success body (own conversation title) is fine; scrub paths/URLs or use a generic error body.

NITs: autonomous-workflow *final* turn fires no banner (author-deferred, known gap); user-cancel flows the normal-finish path → labeled "Task complete" (the `'stopped'` branch of `isTaskComplete` is currently dead). Nice work on the gating/fail-closed core.

PR #597, opened 2026-07-03
feat(projects): add Project History timeline tab (#180)

FerroxLabs:
Addresses #180 (do not auto-close — release closes).

## What
Adds a first-class **History** tab to the project workspace: a selectable, filterable timeline of what happened on a project over time, with a detail pane, plain-English summaries, and an **Open chat** action.

- Filter pills: **All / Chats / Emails / Refs / Remote** (default All).
- Selecting a timeline item updates the detail pane (summary + related fields + action) instead of navigating away.
- Event types modelled: project created/updated, chats, email ingests, references saved from email, current reference inventory, and remote-attachment captured/imported/excluded/failed.

## How
- **`projectHistory.ts`** (new, pure): folds project creation, chats, ingested emails, reference saves and remote-attachment actions into one time-ordered event list. Summaries are derived honestly from existing data — no transcript summarisation.
- **`ProjectHistoryPanel.tsx`** (new): renders the pills, timeline list and detail pane. References load via `ipcBridge.project.listReference`. The email-ingest source has no backend on `main` yet, so it degrades gracefully to an empty list — project + chat + reference events still render, and email/remote events light up automatically once a backend lands.
- **`ProjectWorkspacePage.tsx`**: wires the History tab next to Memory.
- All user-facing copy goes through i18n (`projects.timeline.*`); `i18n:types` + `check-i18n` pass. Renderer-only change.

> `ProjectWorkspacePage.tsx` shows a larger diff than the logical change because touching the file triggers CI's `oxfmt --check`, which canonicalises the whole file (it predates the current oxfmt formatting). No semantic changes in the reflowed region.

## Tests
- `tests/unit/renderer/projectHistory.test.ts` — 12 unit tests covering event mapping by source/status, dedup of email-saved references vs inventory, sort order, filter buckets, per-source counts, and graceful empty-email degradation.
- Full suite green (11892 passed); lint (0 errors) / typecheck (clean for changed files) / format:check pass.

Needs a live-verify pass in the running app (user-facing UI): open a project → History tab, confirm pills filter, selection drives the detail pane, and **Open chat** navigates.

Issue #180, opened 2026-06-18
[Feature]: Add project History timeline with event summaries and filters


sazora:
## Summary

Project pages would be much more useful with a first-class **History** section: a project-level timeline that shows what happened over time across chats, email ingests, references, and remote attachments.

This should be more than a log or list of links. Each timeline item should have a human-readable title and, when selected, a detail pane with a short summary of what happened and the related project artifacts.

## Problem

Right now project context is split across multiple surfaces:

- chats
- memory / decisions
- reference files
- email-ingest records
- Mail Drop / remote attachment actions

For project work, users need to answer questions like:

- What has happened on this project recently?
- Which email added these references?
- Was that Mail Drop file imported or ignored?
- Which chat discussed this topic?
- What changed since the last time I opened this project?

A plain file list or chat list does not answer those questions well.

## Proposed UX

Add a **History** tab next to **Memory** inside each project.

The History view should default to **All** and show clickable filter pills at the top:

- All
- Chats
- Emails
- Refs
- Remote

The main view should be a timeline. Selecting a timeline item should update a detail pane instead of immediately navigating away.

Each event should show:

- event type
- title
- timestamp when available
- plain-English summary
- related fields, such as sender, source email, file name, attachment counts, remote-link status, etc.
- explicit action button when useful, such as `Open chat`

## Event types worth capturing

- project created / updated
- project chats
- project email ingests
- references saved from email attachments
- current reference inventory
- remote attachment captured
- remote attachment imported
- remote attachment excluded / ignored
- remote attachment import failed

## Notes from a working local prototype

A local prototype was built on branch:

`sazora/wayland:saz/project-email-intake-actions`

Relevant commits:

- `604e47c0a` - adds the project History tab and baseline timeline
- `5e3bb5185` - converts it into a selectable timeline with summaries/details
- `116e52f41` - adds clickable filter pills and remote pending/failed events

The prototype derives summaries honestly from existing project data. It does not pretend to summarize chat transcripts beyond what is available in the current project history data. A later improvement could add generated summaries for chat turns or email bodies when a model is configured.

## Why this matters

This turns project history into project memory in timeline form. For long-running project work, users need a way to scan what happened, filter by source, and drill into the event without manually jumping between Chats, References, Memory, and email-ingest internals.

FerroxLabs:
[...]

PR #844, opened 2026-07-12
fix(projects): drop timeline filters that can never match anything (#180)

FerroxLabs:
The project History timeline shipped in #597 with five filter pills. Two of them
— "Emails" and "Remote" — can never be non-zero, because the panel feeds their
source a hard-coded empty array:

    const emailHistory = useMemo<EmailIngestRecord[]>(() => [], []);

There is no email-ingest backend in this repo: no service, no table, no IPC
channel. The event types were carried over from the fork the reporter prototyped
on. So every user with a project sees two permanently-empty pills, and clicking
either shows an empty timeline. A filter that can never match anything is not a
placeholder, it is a broken control.

This is what the approved shape asked for — a generic project provenance timeline
built on the events Projects ACTUALLY emits, decoupled from the reporter's
email-intake types. The honest sources are the project's own create/update stamps,
the conversations assigned to it, and its reference files. Those already worked;
they were just surrounded by scaffolding for a backend that does not exist.

Removes the email/Mail-Drop/remote-attachment kinds, their filters, icons, the 26
now-unreferenced i18n keys, and the tests that asserted the dead behaviour. The
`reference` kind went with them — it was only ever produced inside the email loop,
so reference FILES (kind `inventory`) are what the Refs pill now selects.

Guarded by a test that fails if any HistoryFilter stops being satisfiable by a
kind the builder can genuinely emit. Re-add the pills WITH the backend that fills
them, not before.

Issue #180 stays open: the timeline exists, but this only makes it honest.


FerroxLabs:
**Independent cross-audit (lane A) — ✅ clean, would approve.**

Traced the "can never match" claim end-to-end and it holds: `ProjectHistoryPanel` hard-codes `emailHistory = []`, no email/remote data is ever threaded into a project (the `email-imap` channel's `references` is the RFC email-thread header, surfaced as a `chat` event, not a project reference), so `counts.email`/`counts.remote` were provably always 0 — dropping those filters hides no valid data. The `reference→inventory` repoint is complete (`kind:'reference'` was only emitted in the removed email loop; real refs are `inventory`). No persisted/deep-link filter state exists (local `useState` only), and no other consumer of the removed union members. The guard test is the right defense: it iterates the runtime `HISTORY_FILTERS` (not a hand-written literal), so re-adding a dead filter breaks compile + test.

Two NITs: (1) the dynamic `t(\`projects.timeline.filter.${key}\`)` loses static i18n validation (check-i18n skips `${` keys) — runtime fine, minor coverage drop. (2) Pre-existing (#597): the timeline namespace was only ever localized in en-US, so non-English users see fallback — this PR neither causes nor worsens it. LGTM.

Issue #750, opened 2026-07-09
[Feature]: Docs - Workflows documentation almost incomprehensible


cdvatx:
### Feature Description

The documentation at https://docs.getwayland.com/concepts/workflows/ is almost incomprehensible, even to people who are quite technical. Please allow for better documentation on this topic to reduce friction for users.

### Problem Statement

Understanding how ForgeFlows work.

### Proposed Solution

Rewrite docs with easier lingo, explain terms, provide examples.

### Feature Category

Other

### Additional Context

Docs

FerroxLabs:
**Rewrite ready** — I've rewritten `concepts/workflows.md` to be comprehensible (plain-language intro + analogy, every term explained — sub-agents, fan-out, pipeline, RON — concrete examples per step shape, library-first structure, progressive disclosure of the RON/ForgeFlow internals, all kept accurate against the current docs). Full drop-in below.

Delivery note: the getwayland-docs source (`~/dev/getwayland/web/`, an Astro/Starlight site) is a local non-git working copy and I can't find its source repo under my access, so I couldn't open a PR — flagging Overwatch for the repo/land path. Applied the change to the local working copy meanwhile.

```markdown
---
title: Workflows (ForgeFlows)
description: What a workflow is, why you'd use one, and how Wayland runs it.
---

A **workflow** is a saved, repeatable job you hand to an agent: a set of steps that work toward one outcome, written down once so you (or anyone you share it with) can run it again without re-explaining it.

Think of the difference between telling a colleague "research this competitor, then draft a one-page brief, then sanity-check the numbers" every single time, versus writing those steps down once and just saying "run the competitor brief." A workflow is the written-down version. Wayland calls its workflows **ForgeFlows**, but they're the same thing — the name just distinguishes them from ad-hoc chat.

![The Workflows library grouped into categories](/assets/img/screens/workflows.jpg)

## Why use one instead of just chatting

You *could* ask a single agent to juggle a multi-step task turn by turn. That works, but the agent has to hold the whole plan in its head, and you get a different result each time depending on how it decides to break the work up.

A workflow makes the structure explicit. You say up front *what the steps are and how they connect*, and Wayland runs them as a coordinated set of helper agents (**sub-agents**) — one per step — instead of one agent improvising the whole thing. That makes runs more consistent, lets independent steps run at the same time, and makes it obvious what the task will actually do before you launch it.

## The three step shapes

Most workflows are built from three kinds of steps. You can mix them:

[...]

PR #849, opened 2026-07-12
fix(providers): guide ChatGPT OAuth on headless hosts instead of hanging

### LINKED ISSUE #525: [Bug]: ChatGPT-subscription OAuth connect fails on headless/VPS installs (localhost-callback flow unreachable remotely)
User (Jiggaman) cannot connect a ChatGPT SUBSCRIPTION via OAuth ("Sign in with ChatGPT", the Codex-CLI path) on a headless / VPS Wayland install. Same class as #524 (connecting a provider on a cloud/VPS deploy fails), OAuth flavor.

NOTE: ChatGPT-subscription OAuth is a SUPPORTED feature (unlike Claude Pro/Max subscription OAuth, which we do NOT build - ToS). So this should work; the problem is the headless/remote environment.

Likely root cause: the OAuth connect flow assumes a LOCAL browser + a localhost callback/redirect (the Codex-CLI-style loopback auth). On a headless VPS there is no local browser and the localhost callback is unreachable from the user's machine, so the flow cannot complete.

Fix direction to investigate: a remote-friendly auth flow — a device-code / out-of-band flow (show a URL + code, user authorizes on their own machine, paste the code back), or a callback that routes through the deploy's public HTTPS origin instead of localhost. Applies to ChatGPT-sub and any other OAuth provider (e.g. X/Grok sign-in).

Part of the cloud-deployment hardening initiative (cloud = future pro tier). ACTION: map the current OAuth connect flow, confirm why headless/VPS breaks it, and design the remote-capable flow.

FerroxLabs:
Connecting a ChatGPT subscription on a headless / VPS install failed opaquely:
the browser PKCE flow opens a local browser (none exists) and listens on a
127.0.0.1 callback that's unreachable from the user's own machine, so the flow
just timed out after three minutes with a generic "sign-in failed".

OpenAI's OAuth offers no device-code grant and the Codex client pins the
loopback redirect ports, so a true device flow isn't available. Instead this
detects a headless Linux host (no DISPLAY / WAYLAND_DISPLAY) up front and,
rather than waiting on the doomed flow, returns actionable guidance: forward
the sign-in port over SSH (`ssh -L 1455:localhost:1455 user@host`) and finish
in a local browser, or reuse a `~/.codex/auth.json` from `codex login`. The
existing Codex-CLI credential reuse (tried first, no browser) already covers
the copy-the-file path.

The remote-WebUI OAuth denial is unchanged — a remote WS caller must never
drive a token mint.

Closes #525


FerroxLabs:
Docs: added a **Sign in with ChatGPT on a server** section to the getwayland headless guide (`start/run-on-a-server.md`) documenting the SSH port-forward (`ssh -L 1455:localhost:1455`) + Codex-CLI reuse flow, matching the in-app guidance. Applied to the local getwayland working copy; deploys via its Vercel pipeline (preview→prod on Overwatch/Sean, same as #750).

Issue #755, opened 2026-07-09
[Tracker] macOS: subprocesses blocked / agent cannot run commands (bundle signature lead)


FerroxLabs:
Canonical tracker for the macOS execution-failure cluster: users on fresh or updated installs report the agent **cannot run any commands / subprocesses are blocked**, in several cases rendering Wayland unusable.

Strongest root-cause lead: **#738 — macOS subprocesses are blocked when the app bundle signature is invalid** (Gatekeeper kills spawned children). If confirmed, this is a packaging/signing defect and would explain most of the cluster.

Duplicate / same-family reports:
- #738 — subprocesses blocked when app bundle signature is invalid (root-cause lead)
- #743 — "Again. Unable to run commands WCORE"
- #734 — failures with fresh install on fresh macOS
- #649 — Wayland starts 10s of instances under macOS (possible same signing/relaunch loop)

Related in-progress work (do not dupe-close without checking): #551 (macOS Intel), #520 (unable to use Wayland since 0.11.2), #618 (wcore agents unable to run commands).

Verification: check `codesign --verify --deep` on the shipped bundle + post-update state, and whether the updater invalidates the signature.

FerroxLabs:
**Root cause confirmed (live repro on a macOS arm64 machine, v0.11.17).** The updater is exonerated — the signature breaker is a **runtime write into the signed bundle by the ijfw MCP server**:

```
codesign --verify --deep --strict --verbose=2 /Applications/Wayland.app
  a sealed resource is missing or invalid
  file added: .../Contents/Resources/app.asar.unpacked/.ijfw/.layout-version
```

Chain: Wayland forks workers with `cwd = app.asar.unpacked` (ForkTask.ts:53-61, for WASM) → the ijfw MCP spawn passes **no cwd** (ijfwMcpClient.ts:316) so the child inherits it → ijfw's `safeProjectDir()` falls back to any *writable* cwd → its layout migration writes `.ijfw/.layout-version` + dirs inside the bundle → seal broken → hardened-runtime macOS denies exec for the app's children (`xcode-select … Operation not permitted`, exit 126 — exactly #738's log signature). ShipIt/Squirrel does whole-bundle swaps and never patches in place; the blockmap path (#726) reconstructs the zip before install, so neither can invalidate the seal.

**Fix directions:**
[...]

Issue #706, opened 2026-07-05
Bug report: IJFW Memory runtime Degraded error - degradation is cosmetic


elJayAdvisor:
### Platform

Linux

### Wayland Version

0.11.13 - core 0.12.23

### Bug Description

The IJFW Memory Status page show the Memory Status as degraded, but investigation revealed that it's just a cosmetic error. A bug report has been attached. 

Bug report written to wayland-ijfw-memory-degraded-bug-report.md (177 lines). 

It covers the summary, environment, reproduction steps, root cause with the spawn code and crash loop mechanism, log evidence, impact, why user-side fixes don't work, and two recommended fixes for Wayland. 

[wayland-ijfw-memory-degraded-bug-report.md](https://github.com/user-attachments/files/29668616/wayland-ijfw-memory-degraded-bug-report.md)

### Steps to Reproduce

Click on Settings
Click on IJFW Memory (Ferrox Labs)

### Expected Behavior

I wouldn't expect to have a memory runtime error

### Actual Behavior

The IJFW Memory Status page show the Memory Status as degraded, but investigation revealed that it's just a cosmetic error.

FerroxLabs:
Confirmed the reporter's root cause against the repo, and it's **broader than the cosmetic status** — flagging for a design + packaged-build pass rather than a speculative fix.

## Confirmed root cause (repo evidence)
The shipped binary disables the RunAsNode fuse: `scripts/afterPack.js:61` → `[FuseV1Options.RunAsNode]: false` (in place since v0.9.6-rc.1, 2026-06-07). So `ELECTRON_RUN_AS_NODE=1` is **ignored** in every packaged build, and `ijfwMcpClient.ts:335-336`:
```ts
const env = buildChildEnv({ ELECTRON_RUN_AS_NODE: '1', ... });
const child = spawn(process.execPath, [entry], {...});
```
launches a second full Wayland instead of running `server.js` as node → stdout log-noise poisons the JSON-RPC pipe → `decode error - killing child` → single-instance detection → respawn loop → the persistent "Degraded". Exactly as reported. It's cosmetic only because memory also works via Codex's direct-`node` MCP config.

## Broader than #706
[...]

PR #850, opened 2026-07-12
fix(ijfw): spawn a real JS runtime, not the fused app binary (#706)

FerroxLabs:
Packaged builds harden the app binary in scripts/afterPack.js with
FuseV1Options.RunAsNode: false (SEC-ELEC-05), applied unconditionally
to every packaged build. Once that fuse is blown, ELECTRON_RUN_AS_NODE=1
is a silent no-op: spawning process.execPath boots a full Electron app
(new window, no stdio JSON-RPC) instead of Node, so the IJFW/MCP
children never complete their handshake and crash-loop -- surfacing to
users as "IJFW Memory: Degraded". Works in dev (unpackaged Electron is
unfused and honours the env var), breaks packaged.

Add resolveJsRuntime(): unpackaged -> the app binary as Node (dev
behaviour preserved exactly); packaged -> the Bun runtime already
shipped in the bundle (resources/bundled-bun, resolved via
getBundledBunDir) and already used for the npx-MCP path; packaged
without bundled Bun -> system node. Never the app binary when packaged.

Apply it to the four sites that spawned process.execPath as Node: the
IJFW MCP client, the IJFW spawn-test probe, the builtin @wayland MCP
launcher, and safeSpawn (node/npm/npx). Bun runs plain .js, ESM .mjs
and npm-cli.js as a drop-in Node replacement. The SEC-007 trusted
npm-cli.js resolution is unchanged -- only the interpreter running it
changes, and Bun lives inside the signed bundle.

Also propagate IS_PACKAGED into forked utility-process workers
(ElectronPlatformServices.worker.fork), alongside the existing DATA_DIR.
Workers fall back to NodePlatformServices, whose isPackaged() reads that
env var; without it a worker in a packaged build reports isPackaged=false
and would resolve the app binary as Node -- reintroducing the same
crash-loop the fuse guarantees. The IJFW spawn sites document
forked-worker execution as an expected scenario (#755).

Requires packaged-build verification before close (per Overwatch).


wayland:
Please don't tag me in irrelevant things.

PR #857, opened 2026-07-12
fix(win): remove wayland: protocol + Run-key residue on uninstall (#490)

### LINKED ISSUE #490: [Bug]: Wayland appears to have broken my electron apps
### Platform
Windows 11 25H2

### Wayland Version

wayland 11.8 

## Summary

After installing Wayland Agent 11.8 on Windows 10, multiple Electron applications on the same machine stopped launching correctly. During follow-up investigation, Wayland-related registry artifacts were still present even though the Wayland files themselves were no longer present on disk.

I am not claiming confirmed causation here. This report is limited to the observed timeline, the cross-app impact, and the residual Wayland-specific system state that remained afterward.

## Environment

- OS: Windows 11 25H2
- Wayland version: 11.8
- Electron apps observed failing: Hermes Desktop, Discord

## Steps observed

1. Install Wayland Agent 11.8.
2. Launch Electron apps on the system.
3. Observe launch failure affecting more than one Electron app.
4. Inspect system state after Wayland is no longer present on disk.

## Observed behavior

- Hermes Desktop and Discord both failed to launch correctly after the Wayland failed install event.
- Running the apps directly produced repeated GPU-process crash behavior followed by shutdown. 
- Wayland launched after silent failed install. Waycore agent (GPT 5.4) noted that logs showed a failed install and that version was 11.7.
- Version display in settings showed 11.8 despite failed install.
- At investigation time, expected Wayland files were missing from disk, but Wayland-related registry artifacts were still present.

## Expected behavior

- Installing Wayland Agent should not be followed by launch failures across unrelated Electron apps.
- Uninstalling or removing Wayland should not leave behind Wayland-specific startup entries, protocol handlers, AppContainer mappings, or search-index references.

### 1. Timing

- The Electron launch failures were first observed after installing Wayland Agent 11.8.
- The issue affected multiple Electron apps on the same machine, not just one application.

### 2. Wayland-specific registry state remained after files were gone

Wayland-related registry artifacts were still present even though the associated files were already missing from disk.

Examples found during cleanup:

- `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`
  - `electron.app.Wayland = "C:\Program Files\Wayland\Wayland.exe" --start-on-boot`

- `HKCU\Software\Classes\wayland\shell\open\command`
  - `"C:\Program Files\Wayland\Wayland.exe" "%1"`

- Multiple AppContainer mappings with display name:
  - `Wayland-Core Sandbox`

- Windows Search index references to:
  - `C:\Users\Wyck\.wayland\`

### 3. The residual state was not limited to one registry value

User-level cleanup removed Wayland-related entries from several locations, including:

- startup value: `electron.app.Wayland`
- protocol handler: `HKCU\Software\Classes\wayland`
- RunNotification residue
- AppSwitched / JumplistData entries
- TileProperties key
- AppCompatFlags entries referencing prior Wayland executables/installers
- MuiCache entries referencing Wayland
- 18 `Wayland-Core Sandbox` AppContainer mapping keys

### 4. Machine-level Wayland references also remained

Four machine-level HKLM Windows Search index entries still referenced `.wayland` and required elevation to remove:

- `HKLM\Software\Microsoft\Windows Search\CrawlScopeManager\Windows\SystemIndex\WorkingSetRules\19`
- `HKLM\Software\Microsoft\Windows Search\Gather\Windows\SystemIndex\Sites\LocalHost\Paths\18`
- `HKLM\Software\WOW6432Node\Microsoft\Windows Search\CrawlScopeManager\Windows\SystemIndex\WorkingSetRules\19`
- `HKLM\Software\WOW6432Node\Microsoft\Windows Search\Gather\Windows\SystemIndex\Sites\LocalHost\Paths\18`

## Disk state at investigation time

These expected Wayland locations were already missing:

- `C:\Program Files\Wayland\Wayland.exe`
- `C:\Program Files\Wayland\Uninstall Wayland.exe`
- `C:\Users\Wyck\.wayland` (removed by hand after using uninstaller)
- `C:\Users\Wyck\AppData\Local\wayland-updater` (removed by hand after using uninstaller)

This suggests that whatever uninstall/removal path occurred did not fully clean up related Windows state.

## Why this seems worth investigating

This report does not prove that Wayland caused the Electron failures. However, the following combination appears significant enough to investigate:

- the failures began after installation of Wayland Agent 11.8
- more than one Electron app was affected
- Wayland-specific registry state remained after the product files were gone
- the leftover state spanned file system, startup, protocol, sandbox/AppContainer, compatibility-cache, and search-index areas

## Requested investigation

Please investigate:

- whether Wayland Agent 11.8 modifies Windows or Electron-related runtime state that can persist after uninstall
- whether the uninstall path leaves behind AppContainer, protocol-handler, startup, or search-index artifacts
- whether those artifacts are expected, and if so, whether they should be cleaned up more fully

FerroxLabs:
Closes #490.

## Problem
The generated NSIS uninstaller removes only electron-builder's own install/uninstall registry keys. Two HKCU entries Wayland writes at **runtime** are left dangling at a now-deleted executable after uninstall:

- the `wayland:` protocol handler — `app.setAsDefaultProtocolClient` (`src/index.ts`)
- the start-on-boot Run entry — `app.setLoginItemSettings` (`applicationBridge.ts`)

## Fix
Add a `customUnInstall` macro (a real electron-builder hook — `app-builder-lib/templates/nsis/uninstaller.nsh`) to both `resources/windows-installer-x64.nsh` and `windows-installer-arm64.nsh`, deleting both per-user keys:

```
DeleteRegKey   HKCU "Software\Classes\wayland"
DeleteRegValue HKCU "Software\Microsoft\Windows\CurrentVersion\Run" "${PRODUCT_NAME}"
```

`DeleteReg*` on an absent key is a no-op, so this is safe whether or not start-on-boot was ever enabled. Both are HKCU to match exactly what Electron wrote (never widened to HKLM). Kept in sync across both arch files.

## On the "broke my other Electron apps" claim
Investigated and **could not reproduce / found no mechanism** — Wayland only ever writes its own per-app HKCU keys, never anything shared across Electron apps. Recording as **unreproducible**; this PR fixes the concrete registry-residue half.

## Verification
NSIS is declarative and only exercised when electron-builder packages the Windows installer — there is no unit surface. Real verification is the **packaged-Windows installer/uninstall smoke on SEANDESKTOP** (install → confirm both HKCU keys present → uninstall → confirm both keys gone). Holding land for that packaged verify per the Windows-leg discipline; no macOS/Linux impact.

### DIFF
--- a/resources/windows-installer-arm64.nsh
+++ b/resources/windows-installer-arm64.nsh
@@ -35,3 +35,17 @@ FunctionEnd
   nsExec::Exec `powershell -NoProfile -ExecutionPolicy Bypass -Command "$$d = '$INSTDIR'; $$self = $9; Get-Process -ErrorAction SilentlyContinue | Where-Object { $$_.Id -ne $$self -and $$_.Path -and $$_.Path.ToLower().StartsWith($$d.ToLower()) } | Stop-Process -Force -ErrorAction SilentlyContinue"`
   Sleep 1500
 !macroend
+
+; #490: the generated uninstaller removes only electron-builder's own
+; install/uninstall registry keys. Two HKCU entries Wayland writes at RUNTIME are
+; left behind, dangling at a now-deleted executable:
+;   - the `wayland:` protocol handler (app.setAsDefaultProtocolClient, src/index.ts)
+;   - the start-on-boot Run entry (app.setLoginItemSettings, applicationBridge.ts)
+; Remove both per-user keys on uninstall. DeleteReg* on an absent key is a no-op,
+; so this is safe whether or not start-on-boot was ever enabled. Both are HKCU to
+; match exactly what Electron wrote (never widen to HKLM).
+; (Keep in sync with windows-installer-x64.nsh.)
+!macro customUnInstall
+  DeleteRegKey HKCU "Software\Classes\wayland"
+  DeleteRegValue HKCU "Software\Microsoft\Windows\CurrentVersion\Run" "${PRODUCT_NAME}"
+!macroend
--- a/resources/windows-installer-x64.nsh
+++ b/resources/windows-installer-x64.nsh
@@ -45,3 +45,17 @@ FunctionEnd
   nsExec::Exec `powershell -NoProfile -ExecutionPolicy Bypass -Command "$$d = '$INSTDIR'; $$self = $9; Get-Process -ErrorAction SilentlyContinue | Where-Object { $$_.Id -ne $$self -and $$_.Path -and $$_.Path.ToLower().StartsWith($$d.ToLower()) } | Stop-Process -Force -ErrorAction SilentlyContinue"`
   Sleep 1500
 !macroend
+
+; #490: the generated uninstaller removes only electron-builder's own
+; install/uninstall registry keys. Two HKCU entries Wayland writes at RUNTIME are
+; left behind, dangling at a now-deleted executable:
+;   - the `wayland:` protocol handler (app.setAsDefaultProtocolClient, src/index.ts)
+;   - the start-on-boot Run entry (app.setLoginItemSettings, applicationBridge.ts)
+; Remove both per-user keys on uninstall. DeleteReg* on an absent key is a no-op,
+; so this is safe whether or not start-on-boot was ever enabled. Both are HKCU to
+; match exactly what Electron wrote (never widen to HKLM).
+; (Keep in sync with windows-installer-arm64.nsh.)
+!macro customUnInstall
+  DeleteRegKey HKCU "Software\Classes\wayland"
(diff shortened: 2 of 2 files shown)


FerroxLabs:
Closing in favor of #860 (A's #490 implementation), which is more correct: it deletes the Run value `electron.app.Wayland` — Electron's default AppUserModelID, since the app never calls setAppUserModelId — whereas this PR deleted `${PRODUCT_NAME}`=`Wayland`, the wrong value name (would not match the key Electron actually writes = a no-op). #860 also scrubs the StartupApproved\Run disabled-marker, guards against electron-builder running the uninstaller during updates, and adds a regression test. Superseded; #490 tracked by #860.

Issue #827, opened 2026-07-11
[Bug] Playwright MCP on Windows: server green but no browser tools exposed/callable


FerroxLabs:
Community report (2026-07). Playwright MCP server shows connected (green), but the session exposes zero Playwright/browser tools.

Reported symptom: "No Playwright tool is exposed in this session. The MCP server shows green on your side, but I have no callable browser actions available to invoke." Platform: Windows.

**Problem:** an MCP server that connects successfully is not having its advertised tools registered into the session's callable tool catalog (Windows).

**Fix direction:** investigate MCP tool discovery/registration on Windows — the server's tool list is not being surfaced to the model. Check the MCP client tool-registration path and any Windows-specific gaps.

Relates to #113 (browser tool non-functional) and per-assistant MCP scoping (#111).

PR #859, opened 2026-07-12
fix(mcp): resolve npx to bundled Bun on chat session injection

FerroxLabs:
A Playwright MCP server (`npx @playwright/mcp`) showed green in Settings
but exposed zero tools in a live chat on Windows. The Settings "Test
Connection" probe rewrites `npx` to the bundled Bun runtime, but the live
session-injection paths passed `npx` verbatim to the external ACP bridge,
whose spawn cannot resolve npx on a machine with no system Node - so the
MCP child never started and no tools loaded. Nothing re-validated
tools/list on the session path, so the failure was silent.

Route every stdio session-injection site through one shared helper,
`resolveStdioMcpCommand`, so the rewrite can never diverge from the probe
again. Applied at all three sites: McpConfig.fromStorageConfig,
buildAcpSessionMcpServers, and buildWCoreUserStdioMcpServers. The helper
lives in its own module so the many tests that vi.mock shellEnv keep the
export. Tier-A McpProtocol is left unchanged (its command/args use
distinct builtin-runtime conditions that a naive merge would regress).

refs #827

### DIFF
--- a/src/process/acp/session/McpConfig.ts
+++ b/src/process/acp/session/McpConfig.ts
@@ -2,6 +2,7 @@
 import type { IMcpServer } from '@/common/config/storage';
 import type { AcpMcpCapabilities } from '@/common/types/acpTypes';
 import type { McpServer } from '@agentclientprotocol/sdk';
+import { resolveStdioMcpCommand } from '@process/utils/mcpStdioResolve';
 
 type MergeParams = {
   userServers?: McpServer[];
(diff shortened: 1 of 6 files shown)


FerroxLabs:
Superseded by #858, which merged for #827 and covers all 5 npx→bundled-Bun consumers (buildAcpSessionMcpServers, buildWCoreUserStdioMcpServers, McpConfig.fromStorageConfig, plus WCoreMcpAgent.toWCoreConfig and GeminiAgentManager that this PR missed). I independently verified #858 ⊇ this PR; closing to avoid a partial-fix regression. #490 fix continues in #860.

PR #926, opened 2026-07-25
TEMP F-01 verification A: docs-only PR must satisfy the required checks

FerroxLabs:
Throwaway verification branch for Milestone F-01. Do not merge, do not review. It will be closed and deleted once the gate behaviour is recorded.

Case A of three: the diff is docs-only (readme.md), so the unit shards should not run and the three required Unit Tests checks should report green with the reason stated, while Code Quality still runs for real.


FerroxLabs:
F-01 verification complete. Case A recorded: docs-only PR, all four required checks green with the unit shards correctly skipped. Closing the throwaway branch.

PR #950, opened 2026-08-11
feat(agent): add Wayland Nano as a first-class built-in ACP agent

FerroxLabs:
## What

Registers **Wayland Nano** (`wnano`) as a first-class built-in agent alongside Wayland Core:

- `ACP_BACKENDS_ALL.wnano` — `cliCommand: wayland-nano`, native ACP over stdio (no npx bridge), streaming on
- `AgentRegistry.createWNanoAgent()` — always listed, merged right after `wcore`
- Brand parity with Wayland Core: orange-on-black (`#ff6b35` on `#141414`) Lucide **circle-gauge** logo (`wayland-nano.svg`, exact `wayland.svg` wrapper), backend label, agent-modes entry, about blurb in all 12 locales
- `AcpConnection` spawn case (required — generic cliPath group)
- `detectCustomAgents()` skips legacy `acp.customAgents` rows pointing at the `wayland-nano` CLI so existing profiles don't list it twice

## Notable decisions

- **Single `default` mode** — verified against the agent's actual ACP advertisement (`availableModes: [default]`; model selection goes through `session/set_model`). Listing modes it rejects would break `session/set_mode`.
- No `fluxCompat`/`skillsDirs` claimed — no chip shown until evidenced.
- Out of scope (follow-ups): binary bundling/resolver, dedicated settings page, onboarding filter, hero-card lock, cron `VALID_BACKENDS` whitelist.

## Verification

- 92/92 unit tests pass across the 7 touched/related suites (incl. 4 new dedup tests, logo/modes/selection coverage)
- `tsc --noEmit` clean · `check-i18n` passes · oxlint/oxfmt 0 errors

Agent side: FerroxLabs/wayland-nano-k3 (CI 6/6 green; panel-certified RC).

PR #951, opened 2026-08-11
feat(agent): bundle wayland-nano into packaged Desktop (prepareWaylandNano + resolver + packaged-gate)

FerroxLabs:
## What — Desktop-embedding parity for Wayland Nano, mirroring Wayland Core's style

Stacked on #950 (this PR's base is the wnano integration branch).

- **scripts/prepareWaylandNano.js** — full mirror of prepareWaylandCore.js: downloads pinned GitHub release assets from FerroxLabs/wayland-nano, SHA-256 verifies against **scripts/bundled-wnano-shasums.json** BEFORE extract/copy/execute, publisher attestation on strict builds, stages to resources/bundled-wayland-nano/{platform}-{arch}/. Fail-closed everywhere (live-verified: strict mode rejects 'latest', missing shasums abort).
- **src/process/agent/wnano/binaryResolver.ts** — same chain as wcore: userData override → bundled resources → dev resources → PATH. AcpAgentManager resolves it BEFORE the cliCommand fallback; Unix spawn path is quote-aware so bundled paths with spaces survive.
- **electron-builder.yml** extraResources + macOS signIgnore; **verify-packaged-resources.js** treats the wnano bundle as critical (digests, provenance, attestation, exact inventory, --wnano-runtime flag); **build-with-builder.js** strict prepare per target + prune list.
- shasums skeleton is deliberately empty and **fails closed** — filled same-commit as the first real release (follow-up noted in the file).

## Release-asset contract (matches wayland-nano release.yml as committed)

URL: .../releases/download/{vTag}/wayland-nano-{VERSION}-{pkg}.zip (VERSION without 'v'; pkg ∈ darwin-arm64, darwin-x64, linux-arm64, linux-x64, win32-x64; zip contains only wayland-nano[.exe] at root). win32-arm64 intentionally unsupported (fail-closed).

## Verification

- 105+ unit tests green: prepareWaylandNano 11/11, binaryResolver 5/5, acpConnectors 27/27, verifyPackagedResources 62/9-skip (pre-existing darwin-sweep skips), prepareWaylandCore 10/10 no regressions
- typecheck clean · oxlint/oxfmt clean
- Live smoke: dev warn+skip manifest, strict rejects 'latest', strict version with empty shasums aborts — all fail-closed

## Follow-ups (by design, block dist builds until landed)

1. Same-commit pin: DEFAULT_WNANO_VERSION, bundled-wnano-shasums.json fill, publisher-attestations.json policy for FerroxLabs/wayland-nano (after the first release ships).
2. ConversationSideQuestionService cliCommand fallback left untouched (minimality) — wire to resolveWNanoBinary() if side questions should use the bundled binary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -209,6 +209,7 @@ lint-results.txt
 docs/superpowers/
 resources/bundled-bun
 resources/bundled-wayland-core
+resources/bundled-wayland-nano
 resources/bundled-officecli
 resources/bundled-constitution-fs
 resources/hub
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -121,6 +121,9 @@ extraResources:
   # aionrs binary (pre-compiled per platform/arch)
   - from: resources/bundled-wayland-core
     to: bundled-wayland-core
+  # Wayland Nano first-party agent binary (pre-compiled per platform/arch)
+  - from: resources/bundled-wayland-nano
+    to: bundled-wayland-nano
   # Native anchored Constitution filesystem helper. The exact target helper
   # and adjacent manifest are staged from source before the ASAR is compiled;
   # the independently generated authority is compiled into main-process code.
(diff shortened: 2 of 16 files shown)


FerroxLabs:
Held out of 0.12.0. This is real, wanted work, but it cannot ship until wayland-nano publishes ATTESTED releases. Evidence, all by execution:

1. The packaged-build gate requires a signed publisher attestation. verifyWNanoRuntime in scripts/verify-packaged-resources.js returns false unless publisherVerified === true, which requires a matching entry in scripts/supply-chain/publisher-attestations.json AND a manifest carrying verified: true.

2. wayland-nano v0.1.0 has NO provenance attestation. All five release assets return HTTP 404 from the attestations API. Known-positive control: the same check against wayland-core v0.13.0 SUCCEEDS and returns sourceDigest 8f9c66842e9cca1c7d1ceddb9f39174fc2e901f8, exactly matching its policy entry. So the method works and the negative is real.

3. As-is this breaks the release build. prepareWaylandNano is called from build-with-builder.js with requireVerified: true for every package target, DEFAULT_WNANO_VERSION is still the literal string latest, and bundled-wnano-shasums.json ships as an unfilled placeholder with no tag entries. Running it throws: Strict wayland-nano preparation requires an exact pinned release tag; "latest" is forbidden.

What is already done and ready for whoever picks this up: all five v0.1.0 archives were downloaded and verified against the published wayland-nano-0.1.0-shasums.json asset (all five match), and the extracted-binary digests were computed. So the checksum half of the manifest is a solved problem. The blocker is purely the attestation.

To unblock, in this order:
1. wayland-nano's release workflow publishes provenance attestations (actions/attest-build-provenance) and a release is cut with them.
2. Add a wayland-nano policy entry to publisher-attestations.json with the real signerWorkflow, sourceRef and sourceDigest from that attestation.
3. Fill bundled-wnano-shasums.json with archiveSha256 and binarySha256 per asset, and pin DEFAULT_WNANO_VERSION to the exact tag in the SAME commit.
4. Rebase onto main and verify with a full packaged build, not just CI.

Until then Nano ships via npx waylandnano@0.1.0, which is verified working. The offline failure path was tested: with a cold cache and no registry it fails in about 1 second with a clear ENOTCACHED error rather than hanging.

Leaving this open as the tracked follow-up for 0.12.1. Note its base is the stale feature/wayland-nano branch, so its mergeability field is meaningless.

PR #952, opened 2026-08-12
feat(agent): inject multi-provider credentials into wnano spawns (C8)

FerroxLabs:
## Summary

Desktop half of **C8 multi-provider parity** for `wnano` (design: `shared/reviews/panel-tui/C8-provider-parity-design.md` §6/§7, Desktop bullets of §9; §5 payload invariants and §8 security invariants honored). Today `resolveFluxRouting` returns `'unknown'` for wnano, so no Flux credential flows at all. This PR adds the full injection arm:

- **`FLUX_API_KEY_FILE` handoff (Q4)** — a new `wnano` arm in `resolveFluxRouting` emits the PATH (never the secret) of a per-conversation key file that Desktop writes under userData: atomic temp-then-rename (mirrors `codexAuthFile.ts`), mode 0600 on POSIX (Windows relies on the userData directory ACL — documented, no POSIX bit checks there), deleted in `kill()` at teardown, key value never logged. Mutual exclusivity, wnano-style: ambient `FLUX_API_KEY`/`FLUX_TEST_KEY` are stripped so the connected key wins; Desktop never emits `FLUX_API_KEY` (it stays the documented manual/dev-only fallback for non-Desktop launches). Unlike the surface arms, native provider keys are deliberately NOT stripped — Nano consumes them as its multi-provider credentials. The handoff is unconditional (not gated on the model pick or the routeThroughFlux toggle), so routing stays stable across `set_model` switches and no respawn churn is introduced.
- **`WAYLAND_NANO_PROVIDERS` payload (§6.2)** — injected at the same merge point as `buildConnectedProviderEnv`. JSON array `[{provider, models, hasKey}]`: no secrets, no endpoint/routing fields (entries are rebuilt with exactly the three contract keys), provider ids constrained to Nano's known set, deduped (first wins), deterministic order (known-set order), bounded to ≤64 entries / ≤256 models per provider / ≤128 chars per id / ≤32 KiB serialized with deterministic drop-trailing-then-trim. `hasKey` is advisory only. Models come from the connected providers' persisted model-registry catalogs (`getRegistryCatalog` + custom model ids); `flux-router` advertises Desktop's fixed tier list (Nano owns the live Flux catalog itself). **Note:** the brief pointed at `providerCatalog.generated.json` as the model source, but that file carries no model lists (and none of Nano's 17 known ids) — the registry catalog is the only Desktop source of per-provider model ids for connected providers.
- **OAuth bearer injection (Q1b)** — for each advertised OAuth provider, refresh at spawn ONLY when <600s remain (no churn), inject the ACCESS token only as `WAYLAND_NANO_OAUTH_BEARER_<NORMALIZED_ID>` plus non-secret `..._EXPIRES_AT_UNIX_SECS`. Refresh tokens are never injected; a bearer with unknown expiry is never emitted (the expiry metadata is contractual). v1 wires **xai only** — the only Desktop OAuth provider in Nano's known id set (`chatgpt-subscription` is not in the set and needs the deferred Responses wire); refresh goes through `xaiRefreshToken()`, which prefers the engine-rotated bundle (#391), so there is no refresh-token rotation race.

Nothing else changes: the wnano registry entry, picker, and mode table already fit (§6.4). `WAYLAND_HOME` is not passed (struck proposal).

## Test plan

- [x] `tests/unit/task/wnano/fluxRoutingWnano.test.ts` — wnano arm: `FLUX_API_KEY_FILE` emitted, `FLUX_API_KEY` never emitted, ambient flux keys stripped, provider keys kept, unconditional handoff, stable routing across model switches, native fallback when unconnected/unwritable
- [x] `tests/unit/task/wnano/providersPayload.test.ts` — schema exactness, hostile endpoint/secret fields never emitted, unknown ids excluded, dedup, deterministic known-set order, all four bounds enforced (64/256/128/32 KiB)
- [x] `tests/unit/task/wnano/oauthBearer.test.ts` — access-token-only env, refresh called <600s / not called when fresh / boundary at exactly 600s, no refresh token in env, expiry metadata emitted, unknown-expiry skip, refresh-failure skip, suffix normalization + collision guard
- [x] `tests/unit/task/wnano/fluxKeyFile.test.ts` — atomic write (no temp sibling left), 0600 on POSIX (skip on Windows), per-conversation isolation, hostile-id sanitization, teardown cleanup, failure paths never throw
- [x] `tests/unit/AcpModelSelectorWnanoColonId.dom.test.tsx` — **Q2 build condition**: `openai:gpt-5.6-terra` / `anthropic:claude-opus-4-8` round-trip through the picker (advertise → render → select → `set_model` echo with the exact colon id), friendly `name` displayed instead of the raw id
- [x] `bunx vitest run tests/unit/task/ tests/unit/AcpModelSelector*.dom.test.tsx` — 12 files, 115 passed / 1 skipped
- [x] `bun run typecheck` (tsc --noEmit) — clean
- [x] `bunx oxlint` + `oxfmt --check` on all touched files — clean
- [x] Full `bunx vitest run` — 15697 passed, 61 failed in 7 files; **all 61 failures reproduce on the pristine base branch** (host-specific: EPERM on `~/.wayland-server` storage, `ERR_DLOPEN_FAILED` native module, recovery/wcore suites). Zero new failures from this PR.

**CI note:** Desktop CI red is PRE-EXISTING (constitution-fs pin) — documented in the design §6 and deliberately not fixed here.

**Base branch:** `feature/wayland-nano`, stacked on the wnano chain exactly like #951 (wnano does not exist on `main` yet).

### DIFF
--- a/src/process/task/AcpAgentManager.ts
+++ b/src/process/task/AcpAgentManager.ts
@@ -26,7 +26,7 @@ import type {
   AcpSessionConfigOption,
 } from '@/common/types/acpTypes';
 import { ACP_BACKENDS_ALL, getCurrentWrapperVersion, getFluxCompat } from '@/common/types/acpTypes';
-import { isFluxModelId } from '@/common/config/flux';
+import { FLUX_MODEL_IDS, FLUX_PROVIDER_ID, isFluxModelId } from '@/common/config/flux';
 import { ExtensionRegistry } from '@process/extensions';
 import { getDatabase } from '@process/services/database';
 import { ProviderRepository } from '@process/providers/storage/ProviderRepository';
@@ -89,6 +89,15 @@ import { extractTextFromMessage, processCronInMessage } from './MessageMiddlewar
 import { ConversationTurnCompletionService } from './ConversationTurnCompletionService';
 import { resolveFluxRouting, type FluxRoutingResult, type RoutingDecision } from '@process/task/fluxRouting';
 import { readConnectedFluxKey } from '@process/connectors/fluxKey';
+import {
+  NANO_KNOWN_PROVIDER_IDS,
+  buildWaylandNanoProvidersPayload,
+  buildWnanoOAuthBearerEnv,
+  cleanupWnanoFluxKeyFile,
+  writeWnanoFluxKeyFile,
+  type WnanoOAuthBearerSource,
+  type WnanoProviderEntry,
+} from '@process/task/wnano';
 import type { McpConfigProjection } from '@process/acp/session/McpConfig';
 import { createMcpSessionState, type McpSessionBackend, type McpSessionState } from '@/common/mcp/sessionReceipt';
 import { createMcpSessionDigestKey } from '@process/services/mcpServices/mcpSessionTruthGate';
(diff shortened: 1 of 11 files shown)


FerroxLabs:
Superseded. Verified by execution, not inference: cherry-picking 7030ae0f1 onto main produced an EMPTY patch ("The previous cherry-pick is now empty"), so every line of this PR is already on main via the #956 squash.

Closing. No work is lost.

PR #962, opened 2026-08-15
fix(build): accept linux musl sharp natives and guard OfficeCLI mutation

FerroxLabs:
Two more release-build failures from the v0.12.0 attempts. Attempt 2 got materially further: macOS arm64 reached Apple notarization with everything bundled, signed and verified, and died there on a transient `NSURLErrorDomain -1009`. These are the two real code failures behind it.

## Linux: musl sharp natives rejected

`verifyWhatsAppNativeTarget` compared the installed `@img` inventory against an exact list that omitted the musl builds. bun's `--os`/`--cpu` filters carry no libc dimension, so a linux install always lands both the glibc and the musl sharp packages:

    sharp-linux-x64  sharp-libvips-linux-x64
    sharp-linuxmusl-x64  sharp-libvips-linuxmusl-x64

The exact-inventory comparison therefore rejected every linux build. Both musl packages declare `os: ["linux"]` and `cpu: ["x64"]` and still have to pass the per-package os, cpu and native-identity assertions, so only the expected list needed to know about them. Nothing is loosened: a foreign-platform or foreign-arch package is still rejected.

This check arrived in #925 and had never been through a real linux release build.

Verified across `linux/x64`, `linux/arm64`, `darwin/arm64` and `win32/x64`. darwin and win32 inventories are unchanged by this commit.

## OfficeCLI: mutation guard, long-path config home, and logging

Following #961, three hardening changes:

- **The pinned digest is re-asserted after the probes.** The probes execute the binary, and this updater has replaced its own executable in place. A build can no longer package bytes that differ from the ones it verified, and the failure now names the cause instead of surfacing as a confusing version mismatch.
- **The isolated config home is resolved to its long form.** `os.tmpdir()` on Windows runners is an 8.3 short path (`C:\Users\RUNNER~1\...`); a consumer that canonicalises it differently would miss the config written there and silently fall back to updates enabled. This is the leading suspect for why #961 held on macOS and Linux but not on the Windows runners.
- **It logs like the other runtimes.** `prepareOfficeCli` was the only bundled runtime that prepared silently, which is why the Windows failure could not be read from the build logs at all.

## Verification

`#961`'s fix confirmed working on a real Windows machine, running the exact CI probe sequence (`help docx/xlsx/pptx`, `--help`, `watch --help`, then `--version`): version stayed `1.0.136` and the executable's SHA-256 was unchanged across the run. Without it the same binary rewrote itself to `1.0.144`.

Full `prepareOfficeCli` run passes locally on both the fresh-download and reuse paths. `prepareOfficeCli.test.ts` and `verifyPackagedResources.test.ts`: 79/79 pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -209,6 +209,7 @@ lint-results.txt
 docs/superpowers/
 resources/bundled-bun
 resources/bundled-wayland-core
+resources/bundled-wayland-nano
 resources/bundled-officecli
 resources/bundled-constitution-fs
 resources/hub
--- a/.planning/HANDOFF-2026-08-15-PUBLISH-0.12.0.md
+++ b/.planning/HANDOFF-2026-08-15-PUBLISH-0.12.0.md
@@ -20,17 +20,17 @@ because I labelled an option by its mechanism instead of its consequence.
 
 ## 1. State right now
 
-| Thing | State |
-|---|---|
-| Desktop `main` | `aae785322` [V] |
-| Version on main | **0.12.0** [V] |
-| `v0.12*` tag | **NONE** [V] — nothing published |
-| Open PRs | **#960** (the one to merge), **#951** (close as superseded) |
-| #960 head | `8dbfab151` [V] |
-| #960 CI | 15 pass / 5 pending / **1 advisory fail** — see §2 |
-| Trust root branch | `release-trust-v1` @ `aae785322`, protected [V] |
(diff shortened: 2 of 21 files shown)


FerroxLabs:
Reopening from a clean base: this branch descended from the pre-squash 0.12.0 branch and conflicted with main. Same commit, replayed on main, in the follow-up PR.

PR #964, opened 2026-08-15
fix(build): seed the real OfficeCLI config, skip unpublished Nano targets

FerroxLabs:
Three release-build failures from attempt 3, plus one diagnostic gap. Attempt 3 cleared the Linux musl-sharp check and reached the installed-payload smoke, which is further than any previous attempt.

## OfficeCLI still self-updated on Windows

#961 disabled the updater by redirecting `HOME` and `USERPROFILE` at the child process. That held on macOS and Linux and did nothing on the Windows runners.

Proved on a real Windows machine: with the environment home pointing at a config saying `autoUpdate: true` and the real user profile saying `false`, the binary honoured the **real profile** and did not update. On Windows the config directory resolves through the shell API, so the environment override is ignored.

An earlier test appeared to show the override working; that result was confounded by the updater's own `lastUpdateCheck` throttle, not the config.

The per-user config is now seeded for the duration of the probes and restored byte-for-byte afterwards, with the environment isolated as well for hosts that do honour it. Verified locally that the developer's own `~/.officecli/config.json` comes back with an identical SHA-256 after a full prepare.

## wayland-nano publishes no win32-arm64 runtime

`Unsupported wayland-nano target: win32-arm64` failed the entire Windows arm64 build. This gap is mine: 0.12.0 pinned five runtimes for six package targets.

That target now skips the bundle with an explicit log and keeps the npx launcher, which #951 deliberately retained as the backstop for exactly this case. It is also no longer demanded of the packaged-resource verifier.

## The installed-payload smoke never declared Nano

`platform-package-smoke.mjs` called the verifier without `--wnano-runtime`, which the verifier requires and refuses to infer, so every platform failed after packaging with:

    at least one --wnano-runtime is required; wayland-nano target identity must never be inferred

## Diagnostic: the cleanliness gate now names the dirty paths

`source worktree is not clean; refusing immutable commit/tree attestation` refused a build without saying what changed, which is undiagnosable once the runner is gone. The gate is unchanged; it now lists the offending paths (capped at 20).

## Verification

`prepareOfficeCli.test.ts`, `prepareWaylandNano.test.ts` and `verifyPackagedResources.test.ts`: 90/90 pass. Full `prepareOfficeCli` run passes locally on both the fresh-download and reuse paths. `isSupportedWNanoTarget` confirmed false for `win32-arm64` and true for `win32-x64`.

### DIFF
--- a/scripts/build-with-builder.js
+++ b/scripts/build-with-builder.js
@@ -798,6 +798,13 @@ try {
   // scripts/bundled-wnano-shasums.json carries the signed release checksums.
   for (const platform of packagePlatforms) {
     for (const arch of packageArchitectures) {
+      if (!prepareWaylandNano.isSupportedWNanoTarget(platform, arch)) {
+        console.log(
+          `wayland-nano publishes no ${platform}-${arch} runtime; skipping the bundle. ` +
+            'Nano still launches on this target through the npx fallback.'
+        );
+        continue;
+      }
       prepareWaylandNano({
         platform,
         arch,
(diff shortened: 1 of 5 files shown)


FerroxLabs:
Reopening from a clean base: this branch descended from the pre-squash #963 branch and conflicted with main. Same commits, replayed on main, in the follow-up PR.

Issue #940, opened 2026-08-06
Bug report: Cannot connect Apple Ecosystem MCP - resource file not present in app bundle


martinlass7:
### What happened

Cannot connect Apple Ecosystem MCP connector. Ran tests with AI help. The conclusion:

Apple Ecosystem MCP connector fails because the app is trying to launch:

/Applications/Wayland.app/Contents/Resources/app.asar.unpacked/out/main/builtin-mcp-apple.mjs

but that file is not present in the Wayland Desktop 0.11.18 app bundle.

I searched the bundle with:

find "/Applications/Wayland.app" -iname "*apple*" -o -iname "*builtin-mcp*"

The only relevant results were:
builtin-mcp-concierge-diag.js
builtin-mcp-search-skills.js
builtin-mcp-image-gen.js
node_modules/run-applescript

There is no builtin-mcp-apple.mjs.

Core is updated to 0.12.26 and the Desktop override path now points correctly to /usr/local/bin/wayland-core, so this looks like a missing bundled connector script or stale connector configuration in the Desktop app.

### Environment
- App: 0.11.18
- Engine: wayland-core 0.12.26
- OS: darwin x64 (23.6.0)

### Diagnostics
**Scheduled tasks** (cron_jobs): 0
**MCP servers** (mcp.config): 8
- `wayland-search-skills` — Enabled but exposes 0 tools — it likely failed to connect or registered nothing; check its command, args, or credentials.
- `wayland-concierge-diag` — Enabled but exposes 0 tools — it likely failed to connect or registered nothing; check its command, args, or credentials.
- `com.microsoft-playwright-mcp` — Enabled but exposes 0 tools — it likely failed to connect or registered nothing; check its command, args, or credentials.
**Providers** (model_registry_providers): 2
**Workspace** (projects + conversations): 20
- `VoicePrint, etc.` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `We continue the novel series project. Stand by for` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `Give me a 100-word summary of what you understand ` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
[...]

Issue #693, opened 2026-07-04
[0.12.0 L3] Non-Bypassable Command Floor + Learned Egress/Approval


FerroxLabs:
Catastrophic-command floor below approval and --force; on-disk LearnedPolicy backing the always-allow memory. wcore-tools (new command_floor.rs), a different crate so it runs parallel to the engine.rs chain. Lane: Core-2. Depends: gate. MUST land before L2. Part of #686.

FerroxLabs:
## Command-floor half — built and graded (branch `feat/non-bypassable-command-floor`, commit `01717623`)

**The premise was half right, and the half that was wrong matters.** Verified at HEAD (`45005c9d`) before building anything:

- `policy.rs:729` is a *string inside an advisory message*, not an enforcement site.
- The **credential-exfiltration denylist** and the **P2b unsaved-work guard** were **already non-waivable** — called unconditionally on all four `BashTool` entry points, consulting no flag, no config field and no environment variable, with the unsaved guard failing *closed* on expiry. Two of the four candidate floor rules were already there. Pinned by test rather than reimplemented.

**What genuinely had nothing underneath it**, measured with `WAYLAND_SANDBOX=none` (the shape tier 2 produces) and a known-positive control in each arm proving the guard layer was reached:

1. `BashTool` authored `<root>/.git/hooks/pre-commit` — arbitrary code execution as the operator on their next commit.
2. `BashTool` read and could rewrite the active profile's `permissions.toml` — the learned-grant store that decides what the agent may do without asking. **A command that appends to it has disabled the guard it is running under, permanently.**

Both are refusals this codebase *already makes* for the in-process file tools. The tree records why they never reached Bash, at `workspace_policy.rs:1791`:

> a write-deny inside a granted root would have to be expressed to the OS sandbox, and `SandboxManifest` has no `fs_write_deny` — so it would hold only for the in-process file tools and **fail open for `Bash`**, which is two answers to one question.

### The floor

`crates/wcore-tools/src/bash/command_floor.rs`, wired first in all four `BashTool` entry points. Two rules:

1. **Repository control surface** — `.git/hooks`, `.git/config`, `.wayland-core`, `.wayland-core.toml`. Host-wide, not workspace-scoped: the workspace-scoped form *did not fire on the red arm at all*, because the only thing keeping a command away from another repository's hooks was the sandbox.
[...]

PR #970, opened 2026-08-16
feat: add guarded autopilot mode for Grok

sazora:
## Summary
- adds a per-conversation `Guarded Autopilot` mode for the Grok ACP backend
- keeps Grok on its native `default` ACP mode and lets Wayland own guarded approval handling; it does not enable raw YOLO/full-auto behavior
- prevents a paired remote WebUI client from arming `autoGuarded`, while preserving ordinary remote ACP mode changes

## Safety behavior
Routine Grok permission requests are auto-approved in guarded mode. Existing catastrophic shell-command checks remain explicit confirmations. This intentionally does not add Grok to `FULL_AUTO_MODE`.

## Verification
- focused unit/red-team suite: 38 passing
- TypeScript typecheck: passed
- lint: no errors (existing repository warnings remain)
- production package: passed

The mode is scoped to a local desktop conversation and is not a sandbox or a replacement for workspace/egress controls.

### DIFF
--- a/src/common/adapter/bridgeAllowlist.ts
+++ b/src/common/adapter/bridgeAllowlist.ts
@@ -580,6 +580,21 @@ export function isRemoteDeniedConfigWrite(name: string, data: unknown): boolean 
 }
 
 /**
+ * Guarded Autopilot is a local-desktop-only control. The paired WebUI's token
+ * proves a remote browser, not that its holder may arm unattended host tool
+ * approval. Keep ordinary ACP mode switching available remotely, but reject
+ * only the Wayland-owned guarded mode at the WebSocket trust boundary.
+ *
+ * This value-level guard mirrors `isRemoteDeniedConfigWrite`: a buildProvider
+ * handler has no remote/local signal, while the WebSocket adapter does.
+ */
+export function isRemoteDeniedAcpModeChange(name: string, data: unknown): boolean {
+  if (name !== 'subscribe-acp.set-mode') return false;
+  const mode = (data as { data?: { mode?: unknown } } | null | undefined)?.data?.mode;
+  return mode === 'autoGuarded';
+}
+
+/**
  * Emitter/broadcast (main -> client) names that must NOT be forwarded to a
  * remote WebSocket peer. Inbound denial (isAllowedForRemote) stops a peer
  * INVOKING a provider; this stops a peer passively RECEIVING an emitter stream.
--- a/src/process/services/recovery/recoveryCapture.ts
+++ b/src/process/services/recovery/recoveryCapture.ts
@@ -186,9 +186,10 @@ async function addFileToEpoch(hash: ReturnType<typeof createHash>, filePath: str
 async function addPathToEpoch(
   hash: ReturnType<typeof createHash>,
   candidate: string,
-  excludedTopLevel: ReadonlySet<string> = new Set()
+  excludedTopLevel: ReadonlySet<string> = new Set(),
+  maxEntries = MAX_RECOVERY_INVENTORY_ENTRIES_PER_ROOT
 ): Promise<void> {
-  let remaining = MAX_RECOVERY_INVENTORY_ENTRIES_PER_ROOT;
+  let remaining = maxEntries;
   let stat: Awaited<ReturnType<typeof lstat>>;
   try {
     stat = await lstat(candidate);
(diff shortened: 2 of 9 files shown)


FerroxLabs:
Thank you for this, and separately thank you for going back to confirm #940 was still broken after the release. That kind of follow-through is genuinely valuable and it did not go unnoticed.

I have given this a full adversarial pass, because it is approval-behaviour code in the security shell and it deserved more than a skim. I want to be straight with you: the verdict is not-yet, but most of the reasoning is about scope rather than craft. Your reading of the trust boundary is right. The WebSocket adapter is the correct place for a value-level gate, isRemoteDeniedConfigWrite is the correct thing to mirror, settleRejectedInvoke is used correctly so the WebUI does not white-screen, and the subscribe-acp.set-mode wire name is right (I chased that as a suspected typo; it is not one). That is a better grasp of this codebase than most first contributions to it.

Five things need resolving before this can land.

**1. The guardrail only inspects execute-kind tool calls.**
classifyDestructiveToolCall returns not-destructive for every other kind, so in autoGuarded a delete, move, fetch, or MCP tool call is auto-approved with no inspection at all. Compare workspaceTrust.ts, where Cowork, our most conservative axis, allows only read, search and edit, and the comment explicitly excludes delete and move because destructive file ops must always prompt. As written, Guarded Autopilot would be less strict than Cowork while being presented in the UI as the guarded option. This needs an explicit kind allowlist before the command classifier runs.

**2. The classifier is not strong enough to carry a user-facing promise.**
It is a pattern denylist, and the file header is honest that it is a backstop rather than a sandbox. That is a fair description for the internal unattended paths it serves today. The new dropdown copy makes a stronger guarantee than the code delivers, and quoting and path-shape variations matter more than they look. We are hardening this separately; I would rather land that first so you are building on something firmer. Either way the labelling and the guarantee need to agree.

**3. The remote gate closes one door of several.**
isAllowedForRemote is default-allow, and none of these paths is currently denied: create-conversation carrying extra.sessionMode, update-conversation with mergeExtra writing the same field, agent.config.storage.set writing the acp.config preferred-mode key, and team.set-session-mode. Each of those arms autoGuarded without ever touching acp.set-mode. That third one is very close to the workspace.trustLevel side door described in the comment you modelled this on, so it is worth re-reading. It is also worth looking just above it at the outbound rule, where we concluded that a second hand-maintained list is exactly how these two drift apart, and derived the rule structurally instead. The same argument applies to a fourth predicate here.

**4. Selector and engine can diverge silently.**
AgentModeSelector prefers dynamic modes, then cached modes, then the static list, and grok will never advertise a Wayland-invented autoGuarded. Once grok reports its own modes, validInitialMode falls back to Default while the manager is still in autoGuarded and still auto-approving. Because the effect only sets local state, it never calls set-mode and so never reconciles. The same path may also make the option unselectable, which would render the feature inert. The Wayland-owned mode needs to be unioned onto capability modes rather than replaced by them.

**5. Please split out the recoveryCapture change.**
The slow Windows fixture is a real problem and worth fixing. But that test was proving the production 20,000-entry cap holds on an attestation path, and after the change it proves that a cap of one rejects two, via a new parameter on the production signature. That deserves its own PR and its own discussion rather than riding along with a feature.

None of this is wasted work. The adapter gate is sound and worth keeping; it needs the declarative paths alongside it. If you want to take another pass, I am happy to review it again, and I would start with the kind allowlist in point 1 since it is the smallest change with the largest effect. If you would rather hand it off at this point that is completely fine too, just say the word and we will pick it up from here with credit to you.

One practical note: main has moved a fair way under this branch, including two changes to bridgeAllowlist.ts, so it will need a rebase before it is mergeable.

FerroxLabs:
Core-lane triage: **removing `area:core`** — this PR contains no Core code.

All nine changed files are TypeScript in the Desktop tree (`src/common/adapter/`, `src/process/`, `src/renderer/`, `tests/unit/`). There are no `crates/**` changes, so nothing here is reviewable or testable by the wayland-core lane, and the label was routing it to a lane that cannot act on it.

Two things for the Desktop lane, neither of which is a Core call:
- The PR is currently **CONFLICTING** against its base and was last updated 2026-08-16. It needs a rebase before it can be assessed.
- The safety claim ("routine Grok permission requests are auto-approved in guarded mode; existing catastrophic shell-command checks remain explicit confirmations") is an approval-policy change and deserves a review against the Desktop approval surface. Core has no equivalent gate to cross-check it against, and the engine-side non-bypassable command floor is separately tracked in #693 — worth reading together, but they are not the same control.

Leaving `area:desktop-ui`, no other change.

PR #973, opened 2026-08-16
fix(macos): sign the nested binaries Apple rejects, keep the pin at stage

FerroxLabs:
Unblocks macOS for 0.12.0. Every notarization submission so far has come back Invalid.

## What Apple actually said

Fetched with the notary-log workflow from #971 (submission 26749782-31d9-4fe1-b698-3bb888ccada0):

    "status": "Invalid",
    "statusSummary": "Archive contains critical validation errors"

14 nested Mach-O binaries, each flagged "The binary is not signed with a valid Developer ID certificate", "The signature does not include a secure timestamp", "The executable does not have the hardened runtime enabled":

- bundled-wayland-core/darwin-arm64/wayland-core
- bundled-wayland-nano/darwin-arm64/wayland-nano
- bundled-constitution-fs/darwin-arm64/wayland-constitution-fs
- 11 whatsapp-bridge natives (sharp .node/.dylib, plus bare-fs/bare-os/bare-url prebuilds for darwin-arm64, darwin-x64 and ios-x64-simulator)

Apple was never timing out. notarytool exits 0 even when the final status is Invalid, so notarizeDmg.js saw only the subsequent stapler failure and reported "stapler staple failed or timed out ... re-run once Apple's notary service recovers". For a deterministic rejection that advice is wrong, and the three built-in retries could never succeed.

## Cause

mac.signIgnore excluded exactly those 14 files. The rule behind it is sound but was applied too widely:

- bundled-bun/bun is Developer ID signed by Jarred Sumner, bundled-officecli/officecli by AionUi Inc., both hardened and timestamped. Apple accepted both. Preserving a real publisher signature is correct.
- Our three runtimes are only ad-hoc / linker-signed, and the whatsapp natives, all new in 0.12.0, are unsigned. Excluding those means shipping unsigned code, which Apple refuses outright.

v0.11.18 carried no signIgnore at all, signed all of them, and notarized. Verified on the shipped 0.11.18 dmg: wayland-core there is Developer ID Application: Ferrox Labs, LLC with flags=0x10000(runtime), so signing it is proven runtime-safe, and that release had zero whatsapp natives.

## Change

Narrow signIgnore to the two genuinely upstream-signed binaries. Everything else is signed by us.

Signing rewrites bytes, so on darwin the packaged gate can no longer compare them to the pinned upstream digest. That guarantee moves to where it is enforceable: prepareWaylandCore and prepareWaylandNano already download, checksum and attest each binary at stage time, and the staged manifest must still record that verified digest. The packaged gate additionally proves the shipped bytes are the ones we signed.

Off darwin nothing is re-signed, so exact byte equality stays enforced there, unchanged.

## This tightens one check rather than loosening it

The constitution-fs gate used a bare codesign --verify --strict. That also passes on an ad-hoc signature, so it could not have detected the exact state that caused this outage. It now uses a requirement pinned to the Ferrox Labs team, like the others.

Verified by execution against real artifacts:

| binary | expected | result |
|---|---|---|
| v0.11.18 wayland-core (Developer ID: Ferrox Labs) | accept | exit 0 |
| v0.12.0 wayland-core (ad-hoc) | reject | exit 3 |
| v0.12.0 bun (Developer ID: Jarred Sumner) | reject, not our identity | exit 3 |

Note for reviewers: codesign -R treats a bare argument as a path to a requirement FILE. Inline requirement text must start with '='. Without it codesign exits 1 with "invalid requirement specification" and fails closed on every binary, including correctly signed ones. That is why the constant carries a leading '='.

## Tests

87 pass across the two directly affected suites, plus 27 in the three contract suites. Full unit suite: 17,666 passing.

Three contract tests asserted that electron-builder.yml still contained the removed patterns; they now assert the opposite, with the reason. The test doubles for both signature checks were made faithful: they previously returned true unconditionally, which would have made every darwin byte-drift assertion in the file vacuous. They now model what codesign guarantees, that a signature is only valid while the bytes are unchanged.

New coverage: a darwin binary that is not Developer ID signed by us is rejected; the signature gate is exercised on exactly the runtimes we sign.

## Follow-up, not in this PR

Our three runtimes should be Developer ID signed in their own release pipelines, the way bun and OfficeCLI already are. That would restore end-to-end byte pinning and make signIgnore correct for them. It needs Apple credentials in three repos.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -306,20 +306,30 @@ mac:
   gatekeeperAssess: false
   entitlements: entitlements.plist
   entitlementsInherit: entitlements.plist
-  # Preserve the exact independently pinned nested executable bytes. OfficeCLI
-  # is already Developer ID signed by its publisher. The released wayland-core
-  # and wayland-nano binaries are linker-signed. Re-signing any of them with
-  # Electron helper entitlements changes its pinned digest and grants a much
-  # broader authority surface than the standalone tool was built with; the
-  # post-package gate rechecks the exact bytes from Contents/Resources.
+  # Exclude ONLY nested binaries that already arrive Developer ID signed by
+  # their own publisher, so packaging preserves that publisher signature instead
+  # of replacing it with ours. Verified on the shipped 0.12.0 dmg:
+  #   bundled-bun/*/bun            -> Developer ID Application: Jarred Sumner
+  #   bundled-officecli/*/officecli -> Developer ID Application: AionUi Inc.
+  # Apple accepted both during notarization precisely because they carry a valid
+  # Developer ID signature, a secure timestamp and the hardened runtime.
+  #
+  # Nothing else may be listed here. Apple refuses to notarize an app that
+  # contains an unsigned or merely ad-hoc/linker-signed Mach-O, and every binary
+  # we bundle that upstream does NOT sign (our own wayland-core, wayland-nano and
+  # wayland-constitution-fs, plus the whatsapp-bridge sharp and bare-* natives)
+  # is exactly that. Listing them here made all six macOS notarization
+  # submissions for 0.12.0 come back "Invalid" with 14 rejected files. v0.11.18
+  # carried no signIgnore at all, signed all of them, and notarized.
+  #
+  # Their pinned upstream bytes are still verified - at STAGE time, where
+  # prepareWaylandCore / prepareWaylandNano download, checksum and attest them
+  # before packaging. Signing is a deliberate act applied to already-verified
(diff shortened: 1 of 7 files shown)


FerroxLabs:
Superseded by the stage-time signing approach. Two independent cross-audits rejected this design: signing during packaging breaks the constitution helper's RUNTIME digest check (embedded in app.asar), which would have shipped a notarized but unlaunchable app, and replacing byte equality with signer identity allowed substitution. Reopened as a new branch that signs each binary BEFORE its digest is recorded and keeps every byte-level check intact.

PR #974, opened 2026-08-16
fix(macos): Developer ID sign staged runtimes so the app can notarize

FerroxLabs:
Unblocks macOS for 0.12.0. Supersedes #973, which two independent cross-audits rejected.

## What Apple actually said

Fetched with the notary-log workflow from #971:

    "status": "Invalid",
    "statusSummary": "Archive contains critical validation errors"

14 nested Mach-O binaries, each flagged "not signed with a valid Developer ID certificate", "signature does not include a secure timestamp", "executable does not have the hardened runtime enabled": wayland-core, wayland-nano, wayland-constitution-fs, and 11 whatsapp-bridge natives (sharp .node/.dylib plus bare-fs/bare-os/bare-url prebuilds).

Apple never timed out. notarytool exits 0 even when the final status is Invalid, so the build saw only the later stapler failure and reported a transient outage, which is why three retries per attempt could never succeed.

bundled-bun and bundled-officecli are Developer ID signed by their publishers (Jarred Sumner, AionUi Inc.) and Apple accepted both. v0.11.18 carried no signIgnore, signed everything, and notarized; it also shipped zero whatsapp natives, which are new in 0.12.0.

## Approach: sign at STAGE time, not at packaging time

Each binary is signed before its digest is recorded, and mac.signIgnore is left unchanged so electron-builder never re-signs. prepareConstitutionFs already had exactly this shape; it just signed ad-hoc with --timestamp=none, which Apple rejects.

The ordering is what makes this safe. The rejected alternative, signing during packaging, breaks two things:

- wayland-constitution-fs is re-hashed at RUNTIME against an authority embedded in app.asar (constitutionFsBinary.ts). A signature applied after that digest was recorded fails every launch with CONSTITUTION_FS_BINARY_UNVERIFIED: a notarized but unlaunchable app. A regression test pins the ordering and was confirmed to fail when the signing step is moved after the digest.
- It forced the packaged gate to accept "signed by us" in place of byte equality, which permits substitution.

## Byte identity is preserved, and authenticity no longer depends on the manifest

Bundle manifests gain binary.stagedSha256, the digest of the bytes actually staged; binary.sha256 still carries the verified upstream digest, so provenance is unchanged. The gate requires the shipped bytes to equal stagedSha256 exactly, and requires the staged bytes to be either the pinned upstream bytes verbatim or a darwin-signed derivative.

The manifest ships inside the app and is not independently authenticated, so stagedSha256 alone would let an attacker swap in a different, or older but still validly signed, binary and rewrite the digest beside it. Both auditors raised exactly that. It is closed cryptographically: each binary is signed with a code-signing IDENTIFIER of the form `<binaryName>.<pinned-upstream-sha256>`, and the gate's codesign requirement asserts that identifier. The identifier lives inside the signature, so it cannot be altered without the signing key. darwinSigningIdentifier() refuses to produce one without a valid pinned digest.

The whatsapp natives are signed in the source tree before the bridge is validated and copied, so verifySourceMirror still compares source and bundle byte for byte with no exemption at all. They are detected by Mach-O magic rather than file extension, so an extensionless native cannot be skipped.

## Requirement correctness

The requirement pins the Developer ID CA and Developer ID Application marker OIDs, not just subject.OU, which alone would also accept an Apple Development or Mac Distribution certificate from the same team.

Note for reviewers: codesign -R reads a bare argument as a path to a requirement FILE, so inline requirement text must start with '='. Without it codesign exits 1 with "invalid requirement specification" and fails closed on correctly signed binaries too.

Verified by execution against real artifacts:

| check | expected | result |
|---|---|---|
| v0.11.18 wayland-core (Developer ID: Ferrox Labs) | accept | exit 0 |
| v0.12.0 wayland-core (ad-hoc) | reject | exit 3 |
| v0.12.0 bun / officecli (other publishers) | reject | exit 3 |
| matching identifier in requirement | accept | exit 0 |
| mismatched identifier | reject | exit 3 |
| Mach-O magic detection on the real bundle | .bare and .node true, package.json false | as expected |

## Tests

121 pass across the six affected suites, including a new signDarwinStagedBinary suite. The sign-before-digest ordering test was confirmed to fail when the ordering is reversed, so it genuinely catches the regression.

Unrelated and pre-existing: tests/unit/process/task/acpAgentManagerNpmFallback.test.ts fails on a machine that has wayland-nano installed at ~/.local/bin, because the test does not isolate PATH. It passes in CI.

## Follow-up, not in this PR

Our three runtimes should be Developer ID signed in their own release pipelines the way bun and OfficeCLI already are, which would remove the need to sign them here at all.

### DIFF
--- a/scripts/build-with-builder.js
+++ b/scripts/build-with-builder.js
@@ -18,7 +18,11 @@ const prepareBundledBun = require('./prepareBundledBun');
 const prepareWaylandCore = require('./prepareWaylandCore');
 const prepareWaylandNano = require('./prepareWaylandNano');
 const prepareOfficeCli = require('./prepareOfficeCli');
-const { signDarwinStagedBinary, resolveDarwinSigningIdentity } = require('./signDarwinStagedBinary');
+const {
+  signDarwinStagedBinary,
+  resolveDarwinSigningIdentity,
+  darwinSigningIdentifier,
+} = require('./signDarwinStagedBinary');
 const prepareConstitutionFs = require('./prepareConstitutionFs');
 const { verifyThirdPartyExecutableLedger } = require('./supply-chain/verifyThirdPartyExecutableLedger');
 const { writeCapabilitySeal } = require('./capability-seal/verifyCandidateCapabilitySeal');
@@ -155,8 +159,20 @@ function viteBuildExists() {
   return fs.existsSync(path.join(mainDir, 'index.js')) && fs.existsSync(path.join(rendererDir, 'index.html'));
 }
 
-function shouldSkipViteBuild(skipViteFlag, forceFlag) {
+function shouldSkipViteBuild(skipViteFlag, forceFlag, signedDarwinPackaging = false) {
   if (forceFlag) return false;
+  // Reusing a stale app.asar while the Constitution helper is re-signed ships an
+  // app that packages cleanly and then refuses to launch: the helper is
+  // re-hashed at runtime against the authority compiled INTO app.asar, and
+  // signing (with a fresh secure timestamp) changes the helper's bytes every
+  // time. The adjacent manifest would agree with the new binary while the
+  // embedded authority still described the old one, so the packaged gate passes
+  // and every launch fails with CONSTITUTION_FS_BINARY_UNVERIFIED.
+  if (signedDarwinPackaging && skipViteFlag) {
+    throw new Error(
+      '[build] --skip-vite cannot be combined with signed macOS packaging: the Constitution authority compiled into app.asar must be rebuilt whenever the helper is re-signed, or the packaged app will not launch.'
+    );
+  }
   if (skipViteFlag) return true;
 
   // Auto-detect: skip if build exists and hash matches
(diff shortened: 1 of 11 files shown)


FerroxLabs:
## Cross-audit round 2: four confirmed findings fixed

Both independent auditors returned NO-GO on the previous revision. Every confirmed finding is now fixed in 0d29dd7.

**1. A release build could ship the unsigned upstream binary.** The gate accepted either the pinned upstream bytes or a signed derivative, so a darwin release could ship genuine code with no hardened runtime, which Apple would reject anyway. build-with-builder now passes --require-darwin-signature whenever the build actually had a Developer ID identity, and the gate then demands the identifier-bound signature. Builds with no identity stay buildable. The decision comes from the build environment, never from data inside the package, so it cannot be flipped by editing a manifest.

**2. The constitution-fs gate accepted ad-hoc signatures.** It used a bare codesign --verify --strict, which succeeds on ad-hoc and so could never have detected the state that caused this outage. It now requires a real Ferrox Labs Developer ID.

**3. Symlinked natives were skipped.** A package manager linking a native in from a shared store would leave it unsigned while electron-builder resolved and packaged it. Symlinks are now followed while they stay inside the bridge's node_modules, deduplicated by real path.

**4. FAT64 Mach-O was not detected.** Added 0xcafebabf and its byte-swapped form, so an extensionless 64-bit universal native cannot fall outside the sweep.

## Refuted after reading the code rather than the claim

- constitution-fs signing is already guarded by `platform === 'darwin'`; it cannot run codesign on a Linux ELF.
- verifyBridgeLock only parses package.json and bun.lock text and checks directory existence, and whatsapp-bridge-source.json pins 9 files, none of them natives. Signing natives cannot break frozen-lock validation.
- No verified-cache path exists in either prepare script, so none can record an unsigned digest for a signed binary.
- wayland-constitution-fs has no upstream digest to lose: it is compiled locally from this repo, and the authority exists to bind app.asar to the on-disk helper.

174 tests pass across the seven affected suites; tsc clean.

PR #993, opened 2026-08-17
fix(memory): un-latch the Memory runtime probe (#891) + arm the MCP build gate (#940)

### LINKED ISSUE #891: [Bug]: Memory runtime shows "Degraded (not reachable) / Memory did not respond" while IJFW memory-server is up and healthy (1.6.5, answering /api/health)
### What happened?

On the IJFW Memory (Ferrox Labs) settings page, "Setup status" reports:
- IJFW installed: OK (installed and up to date)
- CLIs detected: OK (4 detected)
- Memory runtime: Degraded (not reachable)

Clicking **Test** returns: "Memory did not respond. Check the install status above."

However, the IJFW memory runtime is actually up, current, and reachable. Its
dashboard/health endpoint responds correctly:

    $ curl -s http://127.0.0.1:37891/api/health
    {"ok":true,"status":"ok","version":"1.6.5","uptime":16647,
     "ledgerPath":"/Users/dazza/.ijfw/observations.jsonl","obsCount":55}

So Wayland's Memory-runtime health check is a false negative: it reports the
runtime as unreachable while the runtime is healthy (v1.6.5) and serving on
127.0.0.1:37891. Memory itself works in sessions (observations load, memory
MCP tools answer); only Wayland's status panel/Test cannot see it.

Notes from debugging:
- There was earlier a version-skew: the dashboard daemon was still running an
  older 1.6.3 build (started ~28h earlier) while on-disk code was 1.6.5, and
  every fresh start logged "daemon already running -- exiting". Restarting the
  daemon brought /api/health to 1.6.5. The Wayland panel still showed
  "Degraded / did not respond" after that.
- Restarting the Wayland app did NOT change the result. The ijfw dashboard
  daemon runs independently of the app, so an app restart does not restart or
  re-probe the runtime.
- The Wayland desktop log (~/Library/Logs/Wayland/2026-07-15.log) contains no
  entry for this probe (no "memory runtime", "did not respond", or ijfw health
  lines), so the failure reason is not surfaced anywhere the user can see.

Suspected cause: the Memory-runtime probe targets something other than the
live /api/health endpoint (or uses a stricter timeout/version/contract check),
and fails to detect a healthy 1.6.5 runtime. The probe result is also not
logged.

### Steps to reproduce

1. Have IJFW Memory installed and running (memory-server 1.6.5, dashboard
   listening on 127.0.0.1:37891, /api/health returns ok).
2. Open Settings and go to the IJFW Memory (Ferrox Labs) page.
3. Observe "Memory runtime: Degraded (not reachable)".
4. Click **Test**.
5. Observe "Memory did not respond. Check the install status above."
6. In a terminal, confirm the runtime is actually healthy:
   `curl -s http://127.0.0.1:37891/api/health`  -> `{"ok":true,...,"version":"1.6.5"}`

### Expected behavior

With the memory runtime up and /api/health returning ok, the Memory runtime
status should show healthy/reachable and Test should pass. If the probe fails,
the reason (endpoint tried, HTTP status, timeout, version mismatch) should be
written to the Wayland log so the failure is diagnosable.

### Actual behavior

Memory runtime is permanently reported as "Degraded (not reachable)" and Test
returns "Memory did not respond", despite the runtime being healthy and
answering /api/health on 127.0.0.1:37891. No probe detail is logged.

### Wayland version

0.11.18

### Operating system / platform

macOS 26.5.2 (build 25F84), Apple Silicon (arm64), MacBook Air M5. Install via
GitHub release auto-updater (channel latest-arm64).

### IJFW / memory-server version

@ijfw/memory-server 1.6.5 (from ~/.ijfw/mcp-server/package.json).
Dashboard daemon serving /api/health on 127.0.0.1:37891 reports version 1.6.5.

### Relevant logs / diagnostics

/api/health (healthy):

    {"ok":true,"status":"ok","version":"1.6.5","uptime":16647,
     "ledgerPath":"/Users/dazza/.ijfw/observations.jsonl","obsCount":55}

Wayland desktop log: no probe entry found (grep for "memory runtime" /
"did not respond" / ijfw health returns nothing).

ijfw dashboard.log (during the earlier skew): repeated

    [ijfw-dashboard] daemon already running -- exiting

### Additional context

- The status panel's install/CLI checks pass; only the runtime reachability
  check fails.
- Two possible product fixes: (a) point the Memory-runtime probe at the same
  /api/health the runtime actually serves, and (b) log the probe attempt +
  failure reason so users can self-diagnose.

FerroxLabs:
Two independent fixes. Closes #891; hardens the build half of #940.

## 1. #891 — Memory runtime reported Degraded while working

Both mechanisms in the report reproduced exactly as described.

**Mechanism A (latched status).** The mount probe in IjfwSetupStatus.tsx was a one-shot effect keyed on installOk, and the Test button never touched the runtime row. The first ok:false the component ever saw pinned the row to "Degraded (not reachable)" for the whole session — including after a Test that passed.

**Mechanism B (a failure with no probe).** ensureSpawned in ijfwMcpClient.ts refuses to fork for 5s after a failed spawn and throws inside that window without attempting anything. That refusal was returned as errorReason spawn_error — a spawn failure that never happened. A probe landing in the window collected a synthetic failure, which mechanism A then made permanent.

Fix:

- The backoff now raises a distinct RespawnBackoffError, surfaced as a new spawn_backoff reason (added to IjfwErrorReason plus a string in all 12 locales). Callers can now tell "we did not ask" from "the runtime is down".
- The mount probe retries exactly once, 5.5s later — past the 5s backoff window — and only for spawn_backoff. A genuine failure still settles on the first answer, so a dead runtime is not hammered: two probes per mount, maximum.
- A Test result now drives the runtime row in both directions, because it is the latest evidence available.

The backoff itself is untouched: same window, same refusal to fork. Only what it reports changed.

## 2. #940 — the build half

Four @wayland connectors (apple, imap, news, cal.com) are missing from the shipped app because they bundle from the sibling waylandmcp tree. That tree has no GitHub remote, so CI cannot check it out. **This PR does not fix that** — no repo was pushed, created or vendored. What it fixes is that the failure was silent.

scripts/build-mcp-servers.js warned-and-skipped in both failure modes (source tree absent, bundle step threw). Both are now fatal. The only way past is WAYLAND_ALLOW_MISSING_MCP=1, which the failure message names, and which prints a ::warning:: line naming the dropped connector when used.

**The gate is armed but currently bypassed in CI.** Every workflow that builds (build-and-release, _build-reusable, build-matrix, build-manual, pr-checks, publish-npm) sets WAYLAND_ALLOW_MISSING_MCP=1 with a comment saying to delete it the moment waylandmcp is checkout-able. Deleting one line then makes a silent drop impossible. Without the bypass this PR would break every build today, which is why it is there and why it is commented rather than quiet.

Also: WAYLAND_MCP_SRC is now authoritative. An explicit override that silently falls back to a different tree is the same class of quiet substitution.

**Not done — the Library card filter.** useMcpLibrary builds its catalog synchronously from a build-time Vite glob. Knowing whether a connector's script actually exists is a main-process filesystem fact, so filtering there needs a new IPC surface, an async state in a currently synchronous hook, and a policy for what to show when the check itself fails. That is not cheap or reliable at that point in the UI, and the build gate closes the hole at the source. Left for #940 follow-up.

## Verification

- bunx tsc --noEmit: clean.
- New #891 tests fail before the fix, pass after. Pre-fix: 4 failed (expected 'pending' to be 'checking'; called 1 time not 2; expected 'pending' to be 'ok'; expected 'ok' to be 'pending') plus the client test (expected 'spawn_error' to be 'spawn_backoff'). Post-fix: IjfwSetupStatus.dom 20/20, ijfwMcpClient 24/24.
- New #940 tests: 6/6 pass. The end-to-end one runs the real script with a missing source and asserts a non-zero exit; pre-fix that same test recorded exit 0 (the silent skip), post-fix it exits 1.
- ijfw + settings + memory + i18n suites: 796 passed across 64 files. scripts + mcp suites: 363 passed across 31 files.

DO NOT MERGE without review.


FerroxLabs:
CI note: the first run had one Windows shard fail on tests/unit/process/services/recovery/recoveryCapture.test.ts - bounds content hashing and rejects a 20,001-entry authority tree, a 30s timeout on a test that synchronously writes 20,001 files. It touches nothing in this diff. Re-ran the failed jobs: all 22 checks now pass.

wayland:
STOP TAGGING ME!

PR #1005, opened 2026-08-17
fix(team): stop teammates hanging in Processing and reconcile leader state

### LINKED ISSUE #980: Teams: leader holds contradictory team/artifact state — control plane does not reconcile with execution plane
## Summary

On 0.12.0, a three-agent Team executed real work end to end, but the **leader repeatedly held contradictory views of team and artifact state**. The execution plane worked; the control/task plane did not stay consistent with it.

Reported by a user on Facebook after upgrading 0.11.18 to 0.12.0 on Qubes. Reporter has screenshots and artifacts and has offered a tighter repro.

## Topology

- Backend: Gemini for all three members
- Model: defaulted to Kimi K2.7 Code, switched to Kimi K3 (see linked issue on the stuck Processing state)
- Members: Backend Architect (leader), Code Reviewer (worker), QA Engineer (verifier)

## What worked (execution plane)

- Leader dispatched real work to the worker
- Worker independently inspected the shared project and produced a substantive artifact (~11 KB)
- Worker reported completion to the leader
- Verifier was dispatched and independently audited the worker artifact, producing its own ~14 KB artifact
- **Resilience:** an external API billing interruption killed the run mid-flight. After billing was restored the existing Team recovered without being recreated. Agent context survived, the worker artifact survived, and the verifier resumed its in-progress audit.

## What failed (control plane)

The leader alternated between mutually exclusive beliefs during a single run:

1. The worker/verifier artifact **existed** and **was missing**, at different points
2. Teammates were reported as **failed** while the verifier had demonstrably resumed and was actively working
3. Task-board writes **appeared not to stick**

## Impact

The leader cannot be trusted to drive completion logic, because its view of the team is not derived from, or reconciled against, the state the execution plane actually persisted.

## Open variable, do not close on this

The reporter is deliberately **not** yet calling this a Teams bug. This run used the Gemini backend; he plans to repeat the same topology and workload on the OpenCode backend, which he has historically found reliable, before attributing the behaviour to Teams itself.

**This issue should stay open pending that rerun.** If OpenCode reproduces it, it is Teams. If it does not, the fault is in the Gemini backend or session path underneath.

## What to verify

- Where the leader's view of teammate status and artifact existence is sourced from, and whether it is ever reconciled against on-disk/session truth
- Whether task-board writes are acknowledged before they are durable
- Whether a failed status is sticky once set, even after the teammate resumes

Related: #889 (make tool effects idempotent or reconcilable)

### LINKED ISSUE #983: Teams: Gemini backend defaults to Kimi K2.7 Code and members hang in Processing with no error
## Summary

On a fresh three-agent Team using the Gemini backend, the model defaulted to **Kimi K2.7 Code** and all three members sat in a stuck **Processing** state. Manually switching the members to **Kimi K3** cleared it and the run proceeded normally.

Reported by a user on Facebook after upgrading 0.11.18 to 0.12.0 on Qubes.

## Two separable questions

**1. Why does it default to Kimi K2.7 Code here?**
The members were created against the Gemini backend, so a Kimi model arriving as the default needs explaining. Either the default resolution is picking a model the selected backend cannot properly drive, or the backend/model pairing is being resolved from stale or shared state.

**2. Why does the failure present as an indefinite Processing state rather than an error?**

This is the part that matters most. A model that cannot serve the request should surface a failure the user can act on. Instead the UI sat in Processing with no error, and the only route out was the user guessing to change the model.

That symptom is the same class as the turn-never-finishes bug already root-caused on the wcore path, where the terminal frame was never registered so the UI never left the running state. **This is a different backend, so do not assume the same cause** — but the honest-failure-surfacing requirement is identical.

## What to verify

- How the default model is resolved for a newly created Team member, and whether the selected backend constrains that set
- Whether the Gemini/Kimi K2.7 Code path returns an error that is being swallowed, or genuinely hangs with no terminal frame
- Whether a member stuck in Processing has any timeout or user-visible escape other than manually changing the model

## Priority note

The stuck state is the user-facing defect. Even if the default turns out to be legitimate, an unservable model must fail loudly rather than hang.

FerroxLabs:
Fixes #983
Fixes #980

Both issues live in the same teammate status lifecycle, so fixing one alone leaves the other open.

## #983 - members hang forever in "Processing"

Three things had to be true at once for a member to wedge with no error and no way out.

1. The inactivity watchdog was armed AFTER await agentTask.sendMessage(). That await is not bootstrap-only: on the gemini backend the worker resolves send.message at END OF TURN (src/process/worker/gemini.ts), and the fork round-trip carries no deadline of its own. So the watchdog could never guard the turn most likely to hang. The member sat at active forever, the catch in wake() never ran, and nothing reached the user. It is now armed before the send, and disarmed when the send rejects.

2. ForkTask.postMessagePromise settled only on the child's callback message. A child that crashed, was killed, or failed to spawn left every caller pending forever. In-flight waiters are now tracked by pipeId and rejected from the child's exit and error handlers. Exit rejects even when expected, because kill() and the child's own complete/error frame are still exits and any un-posted callback is never coming. An opt-in timeoutMs is also available, with deliberately no default - send.message is turn-scoped, not request-scoped, so a blanket deadline would abort healthy long turns.

3. The stall handler never released the wake lock, so the slot stayed in activeWakes and TeamSessionService.restartAgent refused with "Cannot restart while wake in progress" - the stuck member had no escape hatch at all. The right rail compounded it by rendering Restart only for failed, never for the active state a wedged member actually sits in. The lock is released when the watchdog fires, and Restart now renders for active too, behind a confirmation.

Five call sites had to be taught to handle the new rejection or a dead worker would surface as an unhandled rejection: the Gemini auto-approve path, the yolo auto-confirm in BaseAgentManager, the Gemini MCP confirm bridge, the generic confirm bridge, and ChannelMessageService (whose surrounding try/catch is synchronous and could not catch a rejected promise).

## #980 - leader holds contradictory state about its team

1. setStatus wrote the in-memory array only, though TeamAgent.status is a persisted field of TTeam.agents that TeamSessionService reads back. Status is now persisted - through a dedicated repository method, not the generic update. That distinction is the whole fix: repo.update is a whole-row read-modify-write from the caller's snapshot, so persisting the manager's roster would have reverted every field the manager does not own. changeAgentBackend writes agentType straight to teams.agents and never touches the live manager, so the next status write would have put the old backend back - wrong CLI in the rail, wrong CLI spawned on the next session load - and concurrent renameTeam, setSessionMode, workspace change and promoteToStanding were exposed the same way. updateAgentStatuses reads, merges by slotId and writes inside ONE transaction, and touches only the agents and updated_at columns. The manager hands over {slotId, status} pairs, never agent bodies, so it has nothing stale to write back.

2. Nothing reconciled the roster after a restart. A persisted active belongs to a process that died mid-turn, so TeamSession reconciles it to pending on load - no process, needs a full role prompt, exactly what restartAgent leaves behind - and writes that back. Reconciliation is an explicit load-path call rather than constructor behaviour, so a manager built around genuinely mid-turn agents is not clobbered. killAgentProcess also flips the live slot to pending, matching what both of its callers immediately write to the DB.

3. The task Watchdog rewrote the board with zero notification. It re-queues, terminates and completes-through tasks, and wrote only to team_event_log, which no agent reads. There is no mailbox write anywhere in Watchdog.ts, which is why a leader could keep reporting that a member was working a task the sweep had already taken away. The sweep now collects its rewrites and hands the whole batch to a notifier once, after the re-entrancy guard is cleared and without awaiting it; TeamSessionService writes one mailbox message per task but wakes each affected leader exactly once. Batching is not cosmetic: a leader wake is turn-scoped, so notifying per task inside the sweep meant N serial leader turns and a dropped 60s tick for the duration, suspending zombie detection exactly when the team was already degraded.

4. The leader had no push channel: ipcBridge.team.agentStatusChanged reaches only the renderer. Member failures are now written to the leader's mailbox. The 429 path is included, guarded to the transition into failed so a burst of quota errors does not flood the leader.

5. failed was sticky. finalizeTurn returned only an active agent to idle, and the heartbeat re-arm required an existing timer that the timeout had already deleted, so a teammate that recovered on its own stayed dead in the leader's view forever. Soft failures (inactivity, 429) are tracked separately: streaming or a completed turn brings the slot back and re-arms the watchdog, while a hard crash, a kill, or a hard send failure drops the grant so nothing can resurrect a process that is gone.

Owner rejection was already fixed in #994 and is untouched.

## Wake ownership

Releasing the wake lock when the watchdog fires made a second concurrent wake reachable for the first time. wake #1 could still be pending inside sendMessage while wake #2 was live, and its late rejection would disarm wake #2's watchdog and mark a healthy turn failed - the #983 symptom itself, reintroduced one wake later. Each wake now claims a generation token and the whole post-send tail mutates slot state only while it still owns it. The watchdog, a kill and a crash all retire the generation so an abandoned send settles inert.

## Restart safety

During a normal turn the main process refuses a restart while the wake lock is held, so the new active-state button is only a toast. But a member that recovered from a soft failure is back to active WITHOUT re-entering the wake lock, and there the restart is accepted and kills a live streaming CLI mid-turn. Restarting from active therefore goes through a confirmation that says what it will do; restarting a failed member stays one click, which is the point of that control. The new strings are translated in all twelve locales.

## Verification

Every new test was run against the pre-fix code first and fails there.

- The 13-test status-lifecycle block in tests/unit/team-TeammateManager.test.ts: 11 fail against ferrox/main, measured by reverting every src file this branch touches and running the file. The 2 that pass are negative controls - a crashed member must not be un-failed by a straggler frame, and a rejected send must not double-escalate. (An earlier revision of this description said "8 of 10"; that was wrong, and this is the re-measured number.)
- tests/unit/ForkTaskBoundedCall.test.ts: 6 of 7 fail on the base, by timing out at 10s - the timeout is the hang itself.
- The wake-ownership regression fails on the pre-fix branch with "expected 'failed' to be 'active'", which is precisely the healthy turn being killed by the stale wake. The hard-failure-grant regression fails with "expected 'idle' to be 'failed'", which is precisely the dead member being advertised to the leader as available.
- The repository regressions fail with "repo.updateAgentStatuses is not a function", and the wiring regression with "expected update to be called 0 times, but got 1".
- The sweep-batching regression fails on the pre-fix branch.

Commands run:

- npx vitest run over tests/unit/team-TeammateManager.test.ts, team-TeamSession.test.ts, team-SqliteTeamRepository.test.ts, tests/unit/process/team, ForkTaskBoundedCall.test.ts and the locale parity suite - 40 files, 547 tests, all passing.
- The wider set of every test file referencing TeammateManager, TeamSession, ForkTask, BaseAgentManager, GeminiAgentManager, TeamRightRail, conversationBridge, ChannelMessageService, ChatActions or SqliteTeamRepository - 71 files, 716 tests, 713 passing. The 3 failures are all acpAgentManagerNpmFallback.test.ts, which fails identically with every src change on this branch reverted: it expects "npx waylandnano@0.1.1" but resolves the real wayland-nano binary on the developer's PATH.
- npx tsc --noEmit - clean.
- npx oxlint on every changed source file - 0 errors.
- npx oxfmt over exactly the files this branch touches, committed separately; npx oxfmt --check now reports all 31 matched files correctly formatted.

## Judgement on the remaining race

The status-only merge alone would have narrowed the window rather than closed it, because repo.update rewrites every column from the caller's snapshot - so a status writer that re-read the row would still revert a rename committed between its read and its write. That is why the merge lives in the repository inside a single better-sqlite3 transaction and writes only the agents and updated_at columns, instead of in TeamSession. With that, there is no interleaving left to lose: a concurrent writer either commits before the transaction's read and is preserved in the merge, or after the transaction's write and wins outright. The remaining exposure is unchanged from before this branch and belongs to the other writers, which still use the whole-row update and can still lose each other's updates; narrowing that is a separate change and is not made worse here.

## Two self-inflicted CI failures, fixed

Both were real and deterministic, not flake, and both came from this branch.

The generated i18n key type file was not regenerated after the two new locale strings were added. check-i18n.js fails on a stale i18n-keys.d.ts, which reds out two required checks rather than one: the I18n Check job runs the script directly, and Code Quality runs it again through the prek i18n hook because locale files changed. Regenerated with scripts/generate-i18n-types.js, which is why a generated file appears in this diff. The unrelated es-ES warnings that job also prints are pre-existing and warning-only: a pristine ferrox/main tree emits the same two warnings and still exits 0.

Bounding the fork round-trip created three unhandled-rejection paths that only surface once a promise actually rejects. The Gemini confirm bridge chained catch straight onto task.confirm, which is typed void and returns nothing in every implementation except Gemini's, so it threw for the existing bridge test's mock. TeammateManager fired the leader wake unguarded in maybeWakeLeaderWhenAllIdle and in the testament wake at the end of handleAgentCrash, and vitest fails an entire shard on one unhandled rejection. All three now behave like the other confirm sites: the durable mailbox entry is already written, so a failed wake logs and its content rides the next one.

Reproduced locally with the same shard split CI uses. Before: shard 1/4 exit 1 with "Unhandled Rejection: fork task child exited before responding" from team-TeammateManager.test.ts, and shard 3/4 exit 1 with "TypeError: Cannot read properties of undefined (reading 'catch')" at geminiConversationBridge.ts:31. After: shard 1/4 exit 0 (4788 passed), shard 3/4 exit 0 (4536 passed), shard 4/4 exit 0 (4720 passed). Shard 2/4 exits 1 locally on the three pre-existing acpAgentManagerNpmFallback cases only, which resolve the real wayland-nano binary on this machine's PATH and pass in CI. check-i18n.js exits 0, oxfmt --check reports all 33 files correctly formatted, oxlint reports 0 errors, tsc --noEmit is clean, and check-ui-tokens.js exits 0.


FerroxLabs:

This audit was run on 2026-08-17 against head 9e0cce88c and its verdict was never posted here. Recording it now so the PR carries its own evidence.

**Round 1 verdict: FIX-FIRST.** Findings:

1. New unbounded await at ChannelMessageService.ts:397 — the exact defect class issue #983 exists to kill. Reproduced as PENDING after a simulated hour with a live child. The diff BUILT the fix and never wired it: the new options.timeoutMs on ForkTask had zero production callers.
2. The lost-update race was not closed, contrary to the commit message. Five unchanged whole-row writers at TeamSessionService.ts:1184, 1238, 1299, 1377 and 1651 clobber the new live status writes.

**Round 2 (delta audit of the fix) verdict: GO.** Head 62fd38216 is cleared on audit grounds.

One correction to the record: an earlier team-count figure quoted in review was wrong. The real numbers are 48 files and 606 tests, not 47 and 594. The discrepancy is because describeNativeSqlite silently skips locally and hard-fails in CI.

Finding 2 is tracked separately as issue #1057 (five whole-row writers still lose concurrent teammate status). It is not a merge blocker for this PR.

FerroxLabs:

PR #1006, opened 2026-08-17
fix(mcp): stop per-tool switches lying on non-Codex backends

### LINKED ISSUE #998: Per-tool switches in the MCP Library are inert on Wayland Core and ACP backends
## Summary

In the MCP Library you can switch individual tools off per server. **On Wayland Core and on the ACP backends, that switch does nothing.** The tool stays callable. The UI gives no hint, and two in-source comments state the opposite.

Found while researching #475.

## Evidence — every allowedTools consumer on main

`allowedTools` becomes an actual runtime constraint on a launched engine in **exactly one place**:

    src/process/task/codexConfig.ts:90
    if (server.allowedTools !== undefined) entry.enabled_tools = [...server.allowedTools];

Every other reference is something else entirely:
- **UI state** — DetailPage.tsx:611-625, where the switch is rendered and stored
- **A display-only budget count** — toolBudget.ts:19
- **Storage schema** — storage.ts:988
- **Change-detection hashes** — mcp.ts:118, GeminiAgentManager.ts:116, mcpSessionTruthGate.ts:221 (Gemini uses it purely to notice config changed, not to restrict anything)
- **Comments asserting behaviour that does not exist** — mcpSessionConfig.ts:82 ("allowedTools still trims tools within whatever servers stay active") and toolContract.ts:14 and :38

`buildWCoreUserStdioMcpServers` and `buildAcpSessionMcpServers` emit **no tool list at all**, and the wcore launch profile carries only `mcp_servers = [...]` (envBuilder.ts:423-435) — a server-level allowlist with no per-tool dimension.

## Why this matters more than a cosmetic bug

The obvious case is Google Workspace: a user switches off 107 of its 122 tools to cut the flood, and all 122 remain callable on the default engine.

**The serious case is safety.** Someone who switches off a destructive tool — a delete, a send, a payment — reasonably believes they have disabled it. On Wayland Core they have not. A control that appears to restrict capability and silently does not is worse than no control, because it manufactures false confidence.

## Secondary defect in the same area

The count-versus-cap nudge that warns about large tool inventories fires only for `openai` and `gpt-5` model families (src/common/mcp.ts:145). Anthropic and Flux users connecting a 122-tool server get **no warning at all**.

## Recommended fix

Make the switch honest. Either:
1. **Enforce it** — thread `allowedTools` into the wcore profile and the ACP session config so the engine receives a per-server tool allowlist, or
2. **Say so** — if per-tool scoping genuinely belongs to Core's curation layer rather than Desktop, disable the switch on backends that cannot honour it and explain why in the UI.

Option 1 is the right answer if the plumbing supports it. Option 2 is acceptable and still far better than today. **What is not acceptable is leaving a control that lies.**

Either way, fix the two comments that assert the trimming happens — they will mislead the next reader exactly as they misled this investigation.

Size S to M depending on which option.

## Verification

A test asserting that a server with `allowedTools` set to a strict subset produces a launch configuration restricted to that subset, per backend. It should fail today for wcore and the ACP backends and pass for Codex.

FerroxLabs:
Fixes #998

## The lie

allowedTools became a real runtime constraint in exactly one place on main: Codex's generated config.toml, as enabled_tools. Everywhere else it was UI state, a display-only budget count, or a change-detection hash. A user who switched a destructive tool off believed it was disabled. On Wayland Core and the ACP agents it stayed fully callable, and two in-source comments asserted the opposite.

## Split verdict, per backend

**Option 1 (enforce) where the plumbing exists:**

- Codex - already enforced via enabled_tools. No behaviour change; now covered by a regression test so it cannot silently rot.
- Gemini - NEW. allowedTools is now carried as includeTools on the aioncli-core MCPServerConfig, for stdio and hosted connectors alike. aioncli-core's mcp-client filters discovered tools through it (a tool survives only when includeTools is absent or names it), so a switched-off tool never registers.

**Option 2 (say so) where it is genuinely impossible from Desktop:**

- Wayland Core - the launch profile is mcp_servers = [...] with no tool dimension. ProfileConfig has only mcp_servers: Option<Vec<String>>, McpServerConfig has no per-tool field, and the AddMcpServer wire command carries name/transport/command/args/env only. Core's one tool-level retain is a bootstrap-builder persona allowlist over the WHOLE registry (builtins included), reachable only from the ACP path and stripped before crossing the wire. There is nothing Desktop can send.
- The ACP agents (claude, qwen, wnano, grok, kimi, opencode, ...) - the session/new mcpServers descriptor has no per-tool field in the protocol.

For those, the MCP Library's tools panel now states plainly which engines enforce the switches, that the others receive the whole connector so a tool switched off there is still callable, and points at the connector switch as the control that does work.

The switches stay active rather than greyed out because they are not globally inert - they are a real constraint on Codex and Gemini, and they scope the desktop candidate pool everywhere. Greying them out would break two backends to be honest about the rest.

## One source of truth

TOOL_ALLOWLIST_ENFORCING_BACKENDS in src/common/mcp.ts. The UI banner reads its engine names off that constant, so the text cannot drift from the behaviour, and a test cross-checks the list against what each launch builder really emits. Adding a backend to the claim without the plumbing fails the suite.

## Comments corrected

- mcpSessionConfig.ts - "allowedTools still trims tools within whatever servers stay active" was false; replaced, and buildAcpSessionMcpServers / buildWCoreUserStdioMcpServers / buildWCoreSessionMcpServers now document that they are server-level only and why.
- toolContract.ts (two places) - the candidate pool is a desktop-side view, not proof a switched-off tool is unreachable.

## Second commit: two defects found in review

**Disable all must silence a connector, not break it.** The existing Disable all button persists allowedTools: [], which the new Gemini path turned into includeTools: []. aioncli-core's connectAndDiscover treats zero prompts AND zero tools as a failed connection: it throws, emits error feedback and sets the server DISCONNECTED. Switching every tool off would have produced an error toast and a red connector rather than "connector on, tools off" - worse than the bug being fixed. An empty allowlist now drops the connector from the launch entirely, before beginMcpSession so no expected receipt is declared that can never arrive. [] still survives as [] at the aioncli layer, which is correct there; the decision belongs one level up, at whether to emit the server at all.

**The nudge must not invent a vendor limit.** Making resolveModelToolCap always return a number made the nudge always eligible, and its copy names the model and the number - a Claude user with 130 tools would have read "claude-sonnet-4-5 caps at 128". Anthropic publishes no such tool-array cap, so that is a false attributed claim, and the 85% near band would have made it a permanent banner for anyone with a large inventory. Fixing a control that lied by shipping a warning that lies is the same defect wearing a different hat.

resolveModelToolCap now returns { limit, documented }:

- documented cap (OpenAI/gpt-5, where exceeding it really does 400) keeps the near+over bands and the "{{model}} caps at {{cap}}" copy;
- advisory ceiling gets "{{count}} tools enabled. Long tool lists degrade tool selection on most models - scope servers for this chat or switch models." It names neither the model nor a number, is not styled as an error, and fires only once the count is genuinely past the ceiling;
- an unwired documented flag is treated as advisory, so the attribution cannot leak back in by omission.

I deliberately did NOT add per-vendor cap entries I could not verify from documentation. The wording branch gets the same protective effect without asserting a vendor doc I have not read.

Existing nudge tests were updated at the CALL SITE only, to state that gpt-5's 128 is a documented cap. Every assertion in them is unchanged, and all six pass with and without the second commit.

## Verification

Runtime outcome, not just config shape - tests/unit/process/task/geminiEmptyAllowlistLaunch.test.ts:

- the real aioncli-core discoverTools is executed and returns zero tools for includeTools: []
- the real discoverPrompts is executed and returns zero prompts for a prompt-less server
- the installed connectAndDiscover is pinned to still turn that pair into a throw plus MCPServerStatus.DISCONNECTED, so the workaround's premise breaks loudly if the dependency changes
- the real GeminiAgentManager.getMcpServers is driven end to end and proven to omit the connector, emit exactly the enabled tools when scoped, emit no includeTools when unscoped, and declare no expected session receipt for the dropped connector

npx vitest run tests/unit/process/task/geminiEmptyAllowlistLaunch.test.ts -> PASS (7) FAIL (0)

RED proven by execution before each fix, not by reasoning:

- Gemini enforcement, with GeminiAgentManager.ts reverted: PASS (10) FAIL (2), both Gemini cases failing "expected undefined to deeply equal [ 'search_files' ]"
- backend-aware cap, with common/mcp.ts reverted: a throwaway probe on resolveModelToolCap('anthropic', 'claude-sonnet-4-5') received undefined
- empty-allowlist drop, with the filter removed: 2 failed / 5 passed, "expected { workspace: { command: 'uvx', ...(4) } } to not have property workspace" and "expected [ 'workspace' ] to deeply equal []"
- advisory copy, with ConnectorsFlyout.tsx reverted: 3 failed / 18 passed, "expected '130 tools enabled - claude-sonnet-4-5...' not to contain 'caps at'" - and the six pre-existing documented-cap tests passed in that same run, which is the proof they were not weakened

Green, all run after the fixes:

- ConnectorsFlyout.dom + geminiEmptyAllowlistLaunch + mcpToolAllowlistEnforcement + ComposerAddMenu.dom + toolBudget + localeKeyParity -> PASS (142) FAIL (0)
- geminiMcpFingerprint + geminiAgentManagerCrash + geminiAgentManagerThinking + geminiAbortRecovery + geminiBootstrapRejection + mcpStdioSpawn + DetailPage.toolToggles.dom + GuidActionRow.dom -> PASS (45) FAIL (0)
- codexSessionMcp + wcoreUserMcpInjection + mcpSessionConfig.activeServers + getCandidateTools + desktopProfileSplice -> PASS (68) FAIL (0)
- npx tsc --noEmit -> No errors found
- npx oxlint on the changed sources -> 0 warnings, 0 errors
- npx oxfmt --check over exactly the 37 files this PR touches -> all correctly formatted

Two new locale keys (mcpLibrary.detail.toolScopeNotice, conversation.composerMenu.toolNudgeAdvisory) added to all 12 locales with real translations; i18n-keys.d.ts regenerated with the repo script.

## Known, deliberate, not fixed here

The Gemini switch takes effect on the NEXT worker start, not live: computePreviewGeminiMcpFingerprint returns empty unless the MCP session-truth preview gate is on, so a toggle does not re-bootstrap a running worker. This is pre-existing, gated to MCP-2, and applies equally to the existing per-server toggles. Worth a tracked issue, not a change here.

Wayland Core and the ACP agents still cannot enforce a per-tool allowlist. That needs an engine-side and protocol-side change; the Core-side shape would mirror the existing only_for_assistant precedent on McpServerConfig. Worth a wayland-core issue.

Separately noted while tracing this, out of scope and unverified against a live agent: the ACP path injects MCP servers through session/new for every backend that advertises mcpCapabilities, with no codex-specific guard. If codex-acp advertises stdio, the same connector could be registered twice - once filtered from config.toml, once unfiltered from session/new.

### DIFF
--- a/src/common/mcp.ts
+++ b/src/common/mcp.ts
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import type { AgentBackend } from '@/common/types/acpTypes';
+
 /**
  * Shared MCP server-name helpers usable from BOTH the main process and the
  * renderer (the renderer cannot import `src/process/...`).
@@ -140,8 +142,11 @@ export function mcpSessionFingerprint(
  * Per-provider/model hard cap on the tool array a single request may carry
  * (OpenAI's limit is 128). Used ONLY to show the user a count-vs-cap nudge
  * (#348) — Wayland never truncates client-side; Wayland Core owns the smart
- * BM25 curation that actually fits the array, and Flux humanizes the 400. An
- * absent key => no known cap (the nudge then shows the count without a ceiling).
+ * BM25 curation that actually fits the array, and Flux humanizes the 400.
+ *
+ * Entries here are documented VENDOR caps. A provider that is absent is not
+ * uncapped — it falls back to {@link DEFAULT_TOOL_ARRAY_CAP}; see
+ * {@link resolveModelToolCap}.
  */
 export const PROVIDER_TOOL_LIMITS: Record<string, number> = {
   openai: 128,
@@ -149,13 +154,58 @@ export const PROVIDER_TOOL_LIMITS: Record<string, number> = {
 };
 
 /**
- * The known tool-array cap for a chat's target model, or `undefined` when the
- * provider/model has no known ceiling. Checks the model id first (e.g. `gpt-5`)
- * then the provider id (e.g. `openai`) so a capped model under any provider
- * still resolves. Informational only — see {@link PROVIDER_TOOL_LIMITS}.
+ * Advisory ceiling used when the target provider/model publishes no cap of its
+ * own (#998). Previously `resolveModelToolCap` returned `undefined` for
+ * everything that was not OpenAI, so an Anthropic or Flux chat carrying 122
+ * tools got NO nudge while the identical inventory warned on `gpt-5` — the
+ * warning was hardcoded to one vendor rather than being backend-aware.
+ *
+ * 128 is the tightest published ceiling across the providers Wayland routes to,
+ * and a router (Flux) can land a turn on any of them, so it is the honest floor
+ * to warn against. It is advisory only: nothing is ever truncated client-side.
  */
-export function resolveModelToolCap(providerId?: string, modelId?: string): number | undefined {
+export const DEFAULT_TOOL_ARRAY_CAP = 128;
+
+/**
+ * The tool-array cap to warn a chat's target model against. Checks the model id
+ * first (e.g. `gpt-5`) then the provider id (e.g. `openai`) so a capped model
+ * under any provider still resolves, and falls back to
+ * {@link DEFAULT_TOOL_ARRAY_CAP} so every backend gets a nudge rather than only
+ * the vendors listed in {@link PROVIDER_TOOL_LIMITS}. Informational only.
+ */
+export function resolveModelToolCap(providerId?: string, modelId?: string): number {
   if (modelId && modelId in PROVIDER_TOOL_LIMITS) return PROVIDER_TOOL_LIMITS[modelId];
   if (providerId && providerId in PROVIDER_TOOL_LIMITS) return PROVIDER_TOOL_LIMITS[providerId];
-  return undefined;
+  return DEFAULT_TOOL_ARRAY_CAP;
+}
+
+/**
+ * Backends whose LAUNCH CONFIGURATION actually carries the per-server
+ * `allowedTools` list to the engine, so a tool switched off in the MCP Library
+ * is genuinely not callable (#998):
+ *
+ *   - `codex`  — `enabled_tools` in the generated Codex config.toml
+ *                (`buildCodexMcpServerTable`).
+ *   - `gemini` — `includeTools` on the aioncli-core `MCPServerConfig`
+ *                (`buildGeminiStdioMcpConfig`); the runtime drops every
+ *                unlisted tool at discovery time.
+ *
+ * Every OTHER backend receives a SERVER-level allowlist only, and no amount of
+ * desktop-side filtering changes that: Wayland Core's launch profile carries
+ * `mcp_servers = [...]` with no tool dimension (`appendDesktopMcpProfile`, and
+ * Core's `ProfileConfig` / `McpServerConfig` have no per-tool field), and the
+ * ACP `session/new` `mcpServers` array has no per-tool field in the protocol.
+ * On those the switch is UI state plus a desktop candidate-pool filter, NOT an
+ * engine constraint — so the MCP Library says exactly that instead of implying
+ * a restriction that is not there.
+ *
+ * Adding a backend to this list without ALSO emitting its tool list re-creates
+ * the #998 lying control; `tests/unit/process/mcpToolAllowlistEnforcement.test.ts`
+ * cross-checks this list against what each launch builder really emits.
+ */
+export const TOOL_ALLOWLIST_ENFORCING_BACKENDS: readonly AgentBackend[] = ['codex', 'gemini'];
+
+/** Does `backend` carry per-server `allowedTools` through to its engine? */
+export function backendEnforcesToolAllowlist(backend: string | undefined): boolean {
+  return backend !== undefined && (TOOL_ALLOWLIST_ENFORCING_BACKENDS as readonly string[]).includes(backend);
 }
--- a/src/common/mcp.ts
+++ b/src/common/mcp.ts
@@ -160,23 +160,44 @@ export const PROVIDER_TOOL_LIMITS: Record<string, number> = {
  * tools got NO nudge while the identical inventory warned on `gpt-5` — the
  * warning was hardcoded to one vendor rather than being backend-aware.
  *
- * 128 is the tightest published ceiling across the providers Wayland routes to,
- * and a router (Flux) can land a turn on any of them, so it is the honest floor
- * to warn against. It is advisory only: nothing is ever truncated client-side.
+ * This is a rule of thumb about tool-selection quality, NOT a published limit,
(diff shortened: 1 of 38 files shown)


FerroxLabs:

Run on 2026-08-17 against the pre-fix head; the verdict was never posted here.

**Verdict: FIX-FIRST.** The finding that mattered:

Deleting the hosted includeTools left the full suite green at 59 of 59 while the banner still told the user tool restrictions were enforced on Gemini. In reality every SSE and HTTP connector reached Gemini with its full inventory, silently restoring bug #998. In other words the behaviour was untested: the guard could be removed without anything going red.

Fixed at head 8db930e0c. A delta audit of that fix is in progress; this PR is not cleared for merge until that audit reports.

FerroxLabs:
## Independent adversarial delta audit — head 8db930e0c

**Verdict: GO.**

The original FIX-FIRST finding is genuinely closed, verified by execution. Deleting the hosted includeTools spread (GeminiAgentManager.ts:669-670) now kills 2 tests, where the same deletion previously left 59 of 59 green:

    FAIL geminiEmptyAllowlistLaunch > emits includeTools for a scoped http connector
    FAIL geminiEmptyAllowlistLaunch > emits includeTools for a scoped sse connector
    AssertionError: expected undefined to deeply equal [ 'search_files' ]

The behaviour is no longer untestable-by-omission. The test drives the real getMcpServers() and asserts type, url and not.toHaveProperty('command'), so a regression cannot pass by falling through to the stdio builder.

Eight mutants, eight killed, each verified landed with diff:

    M1 delete hosted includeTools spread (the original finding)   2 tests die
    M2 OVER-FIX: always emit includeTools, [] when unscoped       2 tests die
    M3 remove the empty-allowlist filter                          2 tests die
    M4 drop server.allowedTools from the stdio builder call       1 test dies
    M5 add wcore to TOOL_ALLOWLIST_ENFORCING_BACKENDS             2 tests die
    M6 hardcode capDocumented true                                dies
    M7 unwire the modelCapDocumented prop                         dies
    M8 delete the DetailPage honesty banner                       dies

M2 is the one that matters most: emitting an empty includeTools for an UNSCOPED connector would silently disable every tool. That is now pinned, so the fix cannot be over-applied either.

Also clean: tsc exit 0, oxfmt clean on all 9 changed files, oxlint 0 errors. Both new locale keys present in all 12 locales. Dependent cluster 49 files / 548 tests pass. No hardcoded POSIX path literals compared against path.join output, so the windows-2022 shard is not at risk from this diff.

Not verified: no Windows or Linux shard was run, and the PR was not live-verified in a packaged app. This was a code and test audit.

FerroxLabs:

PR #1010, opened 2026-08-17
fix(memory): scope project quick-add and disclose dropped memory

### LINKED ISSUE #924: Bug report: Session-state handoff (.ijfw/) injects lossy or foreign context; one session produced another user's project
### What happened
The persisted session-state/handoff mechanism (.ijfw/ in the workspace root) repeatedly injected wrong or degraded context into fresh sessions, escalating over a day:

1. Lossy handoffs: sessions received "compressed summaries" of prior instructions with items silently dropped or renumbered, then executed the reconstructed (wrong) spec — e.g. a 5-item task list repeatedly lost its first item — or confidently re-reported already-committed work as freshly completed.

2. Cross-project contamination (severe): a fresh session in my Windows repo (D:\Development\UnlikelyAlliesV2, Next.js/Prisma) produced a long, looping implementation monologue for a completely unrelated project — an agent framework called "fatebook" at /Users/jonathandavies/dev/fatebook/ — a macOS path and a username that are not mine and do not exist on my machine. No files were written (paths nonexistent here), but the session was unusable, and a different user's paths appearing in my session is a serious data-hygiene concern.

3. State disavowal: the next fresh session received my LEGITIMATE project handoff and, when questioned, disclaimed my own project as "injected state from a previous session — possibly by someone else."

Possibly related: a mid-task generation failure "Request timeout after 483s ... contentGenerator.timeout" in a session carrying heavy injected state — context bloat suspected.

Expected: handoff state should be verbatim or verifiable rather than lossily summarized, strictly namespaced per workspace/user, and sessions should disclose what state was injected and from where. Transcript excerpts of the "fatebook" output available on request.

### Environment
- App: 0.11.18
- Engine: wayland-core 0.12.25
- OS: win32 x64 (10.0.26200)

### Diagnostics
**MCP servers** (mcp.config): 5
**Providers** (model_registry_providers): 4
**Workspace** (projects + conversations): 4
- `Next.js Supabase Portal Project Setup` — temporary workspace warning
- `Guest Magic-Link Authentication Loop Bug Trace` — temporary workspace warning
**Config paths**
- app: ~\AppData\Roaming\Wayland\config
- engine: ~\AppData\Roaming\wayland-core
**Recent errors** (~\AppData\Roaming\Wayland\logs)

Image

FerroxLabs:
Fixes #924

Two independent defects sat under one title. Both are fixed here, each with a regression test that fails on main.

## 1. A project-scoped memory could be written into a different project

The Memory composer defaults to "project" scope. On save, the main process resolved the destination from the archive index's first project, which is the most recently active IJFW project anywhere on the machine, ranked by memory-file timestamps. It has no relationship to the project the user is working in. Saving a note while in project B therefore appended it to project A's memory journal, and a later session opened in project A read it back as that project's own context.

The fix has two halves, because the first version of it was wrong in an instructive way.

Main is now strict. A project-scoped save resolves to an indexed project root or it is refused. There is deliberately no fallback to the global store: the global block is injected into every chat in every project, so redirecting a misplaced note there would reach further than the original bug, which misplaced it into exactly one project. Refusing is the only honest answer.

The renderer now names the project. An earlier revision of this PR read the project from useActiveBrainScope, which depends on ConversationProvider. The Memory page is a sibling route to /conversation/:id and mounts no such provider, so that hook always returned the app scope and the wiring was inert. The composer now takes the project list from the archive index the page already holds, pre-selects the page's project filter, offers a picker when there is more than one, and displays the resolved destination before the user saves. Previously the pill said "project" while the real destination was decided elsewhere and could differ.

## 2. The injected memory block hid what it had dropped

The global-memory block is the agent's whole record of the user's memory. Two paths shrank it with no marker at all:

- when the full body read failed, the 200-character list preview was substituted and presented as the complete entry, so the agent acted on a fragment with nothing signalling the remainder was missing;
- when an entry did not fit the remaining budget, it and every entry after it were dropped silently.

The oversized-body path already appended a truncation marker. These two now do the same, and the block states how many entries it left out.

## Remote surface

memory.set-quick-add is now denied to remote peers. The memory namespace is documented as open to a paired WebUI for reads with its destructive mutations denied; set-quick-add is a write that was sitting on the allowed side of that line, and it now carries a projectPath selector. Denying it restores the stated policy and keeps the selector off the remote surface. Local IPC is unaffected and the denylist only grows.

## Verification

- New DOM test drives the real /memory route tree, not a service harness, because a service-level test cannot see inert renderer wiring. It asserts the IPC payload carries a concrete projectPath. Against the earlier revision it fails with "expected undefined to be truthy".
- New service test stages two projects with different activity and asserts the entry lands in the one the caller named, and that no project-scoped call shape can reach the global journal.
- New test for the injected-block disclosure.
- Existing suites for the memory service, the memory bridges, the project-knowledge block, the memory renderer pages, the remote denylist and locale parity all pass unchanged. No existing test was edited.
- tsc --noEmit clean, oxfmt clean.

## Not covered here

Global memory is injected into every chat regardless of project. That is the documented behaviour of #256 and has its own test suite asserting it, so it is deliberately left alone; whether it should stay global is a product question worth its own issue.

Closing this does not explain the foreign username in the original report. The mechanism fixed here is bounded by the user's own home directory, so another person's home has no path into it. The likelier origin is the import surface, which writes externally sourced content straight into the global brain that then reaches every chat. Worth a separate issue.

Two pre-existing items noticed and deliberately not touched: memory files are written 0644 while the IJFW atomic-write and lock helpers use 0600, and importBridge resolveMemoryDir contains a branch whose body is only comments, so it always returns the global dir regardless of its doc comment.

### DIFF
--- a/src/common/adapter/bridgeAllowlist.ts
+++ b/src/common/adapter/bridgeAllowlist.ts
@@ -545,6 +545,13 @@ const REMOTE_DENIED_KEYS: ReadonlySet<string> = new Set([
   'memory.update-entry',
   'memory.delete-entry',
   'memory.restore-archived-entry',
+  //     #924: set-quick-add is a WRITE that was sitting on the allowed side of
+  //     the read/write line stated above. It creates an on-disk memory entry,
+  //     and it now carries a projectPath selector that steers WHICH indexed
+  //     project receives it. Denying it puts the namespace back in line with
+  //     the READS-only policy this block documents and keeps the new selector
+  //     off the remote surface entirely. Local Electron IPC is unaffected.
+  'memory.set-quick-add',
   // --- Project knowledge draft (reads arbitrary filePaths to feed the model) ---
   'project.generate-knowledge-draft',
   // --- Storage destructive / disk operations ---
--- a/src/common/adapter/ipcBridge.ts
+++ b/src/common/adapter/ipcBridge.ts
@@ -3221,7 +3221,7 @@ export const memory = {
   /** Quick-add a new memory from the renderer input. */
   setQuickAdd: buildProvider<
     { ok: boolean; error?: string },
-    { content: string; scope: 'project' | 'global'; type?: string }
+    { content: string; scope: 'project' | 'global'; type?: string; projectPath?: string }
   >('memory.set-quick-add'),
   /** Entries whose promotionScore meets the threshold. */
   getPromotionCandidates: buildProvider<PromotionCandidates, void>('memory.get-promotion-candidates'),
--- a/src/process/bridge/memoryArchiveBridge.ts
+++ b/src/process/bridge/memoryArchiveBridge.ts
@@ -43,6 +43,10 @@ const quickAddSchema = z.object({
   content: z.string().min(1).max(10_000),
   scope: z.enum(['project', 'global']),
   type: z.string().optional(),
+  // #924: the project the caller is actually in. The service accepts it only
+  // when it matches an indexed project root, so this is a selector, not a
+  // free-form write path.
+  projectPath: z.string().max(4_096).optional(),
 });
 
 const thresholdSchema = z.object({ threshold: z.number().int().min(0).max(100) });
@@ -187,7 +191,7 @@ export function initMemoryArchiveBridge(): void {
     const parsed = quickAddSchema.safeParse(args);
     if (!parsed.success) return { ok: false, error: parsed.error.message };
     try {
-      await svc.quickAdd(parsed.data.content, parsed.data.scope, parsed.data.type);
+      await svc.quickAdd(parsed.data.content, parsed.data.scope, parsed.data.type, parsed.data.projectPath);
       return { ok: true };
     } catch (err) {
       log.error('[memory-archive] setQuickAdd failed', { err });
--- a/src/process/services/memory/ijfwArchiveService.ts
+++ b/src/process/services/memory/ijfwArchiveService.ts
@@ -789,11 +789,49 @@ class IjfwArchiveService {
     };
   }
 
-  async quickAdd(content: string, scope: 'project' | 'global', type = 'observation'): Promise<void> {
+  /**
+   * Resolve the `.ijfw/memory` dir a project-scoped quick-add must be written
+   * to (#924).
+   *
+   * `projectPath` is the project the CALLER is in. It is accepted only when it
+   * matches an indexed project root, so an IPC caller can never steer a write
+   * at an arbitrary directory. When it cannot be resolved we fall back to the
+   * global store: writing the user's note somewhere predictable is correct,
+   * writing it into whichever unrelated project happens to be most recently
+   * active is not - that was the cross-project bleed in #924.
+   *
+   * The no-argument case keeps the historical `projects[0]` behaviour ONLY when
+   * there is a single indexed project, where it is unambiguous. With two or
+   * more, "most recently active" is a guess and we refuse to make it.
+   */
+  private resolveProjectMemoryDir(projectPath?: string): string {
+    const globalDir = path.join(os.homedir(), '.ijfw', 'memory');
+    const roots = this.index.projects;
+
+    if (projectPath && projectPath.trim()) {
+      const resolved = path.resolve(projectPath);
+      if (roots.some((p) => p.path === resolved)) return path.join(resolved, '.ijfw', 'memory');
+      log.warn('[memory-archive] quickAdd: unknown project path, writing to the global store', { projectPath });
+      return globalDir;
+    }
+
+    if (roots.length === 1) return path.join(roots[0].path, '.ijfw', 'memory');
+    if (roots.length > 1) {
+      log.warn('[memory-archive] quickAdd: no project path supplied, writing to the global store', {
+        candidates: roots.length,
+      });
+    }
+    return globalDir;
+  }
+
+  async quickAdd(
+    content: string,
+    scope: 'project' | 'global',
+    type = 'observation',
+    projectPath?: string
+  ): Promise<void> {
     const memDir =
-      scope === 'global'
-        ? path.join(os.homedir(), '.ijfw', 'memory')
-        : path.join(this.index.projects[0]?.path ?? os.homedir(), '.ijfw', 'memory');
+      scope === 'global' ? path.join(os.homedir(), '.ijfw', 'memory') : this.resolveProjectMemoryDir(projectPath);
     await fs.promises.mkdir(memDir, { recursive: true });
     const journalPath = path.join(memDir, 'journal.md');
     const now = new Date().toISOString();
--- a/src/process/services/memory/ijfwArchiveService.ts
+++ b/src/process/services/memory/ijfwArchiveService.ts
@@ -791,37 +791,37 @@ class IjfwArchiveService {
 
   /**
    * Resolve the `.ijfw/memory` dir a project-scoped quick-add must be written
-   * to (#924).
+   * to (#924), or null when no project can be named with confidence.
    *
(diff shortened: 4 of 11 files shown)


FerroxLabs:

Run on 2026-08-17 against the pre-fix head; the verdict was never posted here.

**Verdict: FIX-FIRST.** Two findings:

1. Issue #924 recurring, introduced by this PR. A background index refresh silently overwrites the user's chosen memory destination: the user picks destination beta and the entry lands in gamma, with no error and no notice.
2. The disclosure half is provably false. The skippedEmpty count is subtracted out of the user-facing notice, so a summary-only entry is dropped with no marker shown at all. The feature reports fewer drops than actually occurred.

Also for the record: a baseline figure quoted during review did not reproduce. The claim was 101; the measured value was 72. Treat quoted numbers on this PR as unverified unless re-measured.

A fix landed at head b2a5a4d7b. That fix has NOT yet been audited; a delta audit is running now. This PR is not cleared for merge until it reports.

FerroxLabs:
## Independent adversarial delta audit — head b2a5a4d7b

**Verdict: FIX-FIRST.** Three defects established by execution. One is a HIGH with real privacy impact.

### F1 — HIGH. A project-scoped save can land in the GLOBAL store

ijfwArchiveService.ts:400 injects the home directory into registryEntries as an ordinary project candidate, which is pre-existing. What is new is that THIS PR makes that same list both the write-destination allowlist (resolveProjectMemoryDir :810-823) and the user-facing picker (ComposerModal.tsx:85 and :367), and nothing filters the home entry out. useMemoryIndex.ts:125 passes it through verbatim.

The list is sorted by lastActive descending, so right after any global quick-add, drag-drop ingest or importer run the global store is projects[0], which is exactly what firstProjectPath seeds the destination from. The user sees "Saving to: <their username>", writes a project-private note, and it lands in the global brain that loadGlobalMemoryBlock injects into every chat in every project on the machine.

Measured with the real service and a populated home store, doing exactly what the composer does:

    landed in GLOBAL = true | landed in project = false

Also measured: on a fresh install that imported memories but has no IJFW project, roots.length === 1 resolves to the home directory, so the no-arg path lands in global too, with no picker rendered because projects.length > 1 is false.

This PR's own doc comment at :797-803 says this outcome would be worse than the bug it replaces. It ships a test named "a project-scoped save can never land in the global store" whose fixture never creates the home store, so buildIndex drops it and the invariant is never exercised.

### F2 — MEDIUM-HIGH. The #924 race is still open for a user who accepts the default

ComposerModal.tsx:103 re-seeds the destination only when userPickedRef is unset, and that flag is set ONLY by the select onChange at :371, which renders only when there is more than one project. A user who reads the destination line and accepts it never sets it. The index refetches on every change and re-sorts by lastActive, so a background agent write reorders the list and the destination moves between open and submit.

    shown on open:              project-alpha
    shown after background refresh: project-gamma
    submitted projectPath:      project-gamma

Known positive established: with the guard removed, the harness also fails the PR's own test, so it genuinely detects the original race.

### F3 — MEDIUM-HIGH. The preview-only disclosure is inert on the real path

knowledge.ts:211-216 clears previewOnly whenever a body is truthy, but getEntry never throws on an unreadable source: it catches at :746 and returns the 200-char preview as the body. So the marker never fires. Driving the real service:

    readable:   len 1918, complete
    unreadable: len 329, truncated mid-word, omitted computes to 0, no marker

That is verbatim failure mode 1 from the PR's own test-file header. The PR's test passes only because it forces getEntry to throw, a shape production never emits.

### Verified CLOSED — the skippedEmpty off-by-one

Measured across five shapes, headings plus disclosed accounts for 100% of input every time, and the block stays inside the char cap. A known positive was established by reverting to the previous version, which under-reported by 20 on a 60-entry case. This half of the PR is genuinely fixed.

### Windows hazard — clean

Checked for the pattern that has bitten three sibling PRs. The POSIX literals here never meet path.join or realpath: the renderer has no path module and splits on a separator-agnostic regex, and both sides of the destination comparison go through path.resolve.

### Test strength

Two mutants survived the PR suite: relaxing exact-root containment to a startsWith prefix check, and deleting the single-root fallback. The current code is correct on both, but nothing pins them.

Baseline measured on the PR-touched files: 85 tests across 4 files, all passing. Earlier figures of 101 and 72 quoted for this PR are different scopes and do not reproduce.

FerroxLabs:
## All three defects FIXED — commit 48ea45eca on the release integration branch

Each was reproduced first as a known positive, then closed, then mutation-tested.

### F1 — a project-scoped save reaching the global store

Reproduced on disk against the real service with the home store populated so buildIndex keeps it:

    INDEXED projects[0] = <home>
    quickAdd('PRIVATE project note', scope 'project', <home>)
      -> global journal exists = true, contains the note = true

and on a global-only install, quickAdd with scope 'project' and no path RESOLVED instead of rejecting, because the single-root fallback picked the home directory.

Fixed by adding an isGlobalStore flag to ProjectSummary, set in buildIndex through one isGlobalStoreRoot predicate, and filtering it out of the destination roots BEFORE both the named-path allowlist and the single-root fallback. The picker and its seed drop it too. Both cases now reject with unresolved_project_scope and the global journal is never created.

Deliberate global saves are untouched: getProjects() is unchanged so the Memory tab still browses the store, and an explicit global scope bypasses the project resolver entirely. That is pinned by an over-fix mutant.

### F2 — the race for a user who accepts the shown default

Reproduced through the real renderer tree: composer opened showing project-alpha, one index refresh with gamma most-recent, and the assertion failed with project-gamma both shown and submitted.

The flag that suppressed re-seeding was set only by the picker's onChange, which renders only when there is more than one project. It is now set as soon as the destination RESOLVES to a non-empty path, however it resolved. Seeding stays live while it is still empty so a cold open whose project list arrives late still fills in, and it is released on close so reopening re-seeds. Both of those are pinned by over-fix mutants.

### F3 — the inert preview-only disclosure

Reproduced against the real service, not a stub:

    readable entry      len 2038, marker false
    source deleted      len 338,  marker false, tail cut mid-content
    after the fix       len 380,  marker true, tail reads "(preview only - full entry unavailable)"
    readable case after byte-identical at 2038, still no marker

getEntry now returns bodyIsPreview, true only in the catch that substitutes the preview, and the composer reads that instead of assuming any truthy body is authoritative. getEntry was deliberately NOT turned into a thrower; its four callers were checked and the field is additive, so the bridge and renderer needed no change.

### Mutation results

17 mutants, all killed, each verified landed with diff against a post-fix baseline and restored. That includes 9 over-fix mutants: an always-true predicate blocking legitimate project saves, routing global scope through the project resolver, filter-everything and inverted-filter variants, always marking real bodies as previews, settling the destination so early a late project list is ignored, freezing so hard the picker stops winning, and never releasing the freeze on close.

The two mutants that previously SURVIVED are now pinned as well: relaxing exact-root containment to a startsWith prefix check, and deleting the single-root fallback. A new fixture with one project plus the global store pins that the exclusion happens BEFORE the length check.

tsc exits 0, oxfmt run on all 9 files. Suites green: 156 memory and projectKnowledge, 255 memory renderer DOM, 40 wiki. No test premise changed, only inert fixtures strengthened and cases added. No new user-facing strings, so no locale work.

### Deliberately not fixed here

An adjacent surface with the same root cause is filed as #1064: two wiki call sites take the most recently active project from the same list and will write wiki state into the home tree after any global save. It is pre-existing on main, verified by reading main directly, so it was not in scope. The isGlobalStore flag added here makes that fix cheap.

A Windows-only case-mismatch residual on the path comparison is recorded on #1064 as well. Reachability could not be demonstrated and it cannot be exercised from macOS, and the existing dedup shares the weakness, so an untested guard was correctly avoided.

FerroxLabs:

Issue #30, opened 2026-06-10
Bug: project-backed WCore conversations can drift to temporary workspaces


sazora:
## Summary
Project-backed WCore conversations should use the associated project's workspace path. A project chat with project metadata can drift to a temporary WCore workspace, causing the model/runtime to lose the project filesystem context.

## Expected behavior
If a conversation has project metadata, such as a project id, its WCore workspace should match the project workspace.

## Observed behavior
A project-backed WCore chat was routed to a temporary workspace even though it was associated with a real project. Repairing the conversation to the project workspace restored the expected behavior.

## Proposed fix
- On project chat creation, set WCore conversation workspace from the project workspace.
- On startup or periodic maintenance, repair non-custom project-backed WCore conversations whose workspace does not match the project workspace.
- Add a healthcheck/test that fails if any project-backed WCore chat points at a missing or non-project workspace.

## Acceptance criteria
- New project chats use the project workspace.
- Existing drifted rows are migrated/repairable.
- Missing workspace paths are detected before WCore launch.
- A regression test covers project id -> workspace sync.

FerroxLabs:
Partial fix shipped in v0.9.7.

The root cause of the drift is fixed: a chat created inside a project now resolves its WCore workspace from the project before the agent factory runs, so new project chats no longer fall through to a throwaway wcore-temp-* directory. A user-chosen custom workspace is still respected, and a project with no workspace still falls back to temp by design. This landed with a regression test covering the project-id to workspace binding (ConversationServiceImpl).

Leaving this open for the remaining two acceptance criteria, which did not ship yet:
- migrate/repair existing rows that already drifted before this fix
- detect a missing or non-project workspace before WCore launch (the healthcheck)

So new chats are correct from 0.9.7 on, but a conversation that already drifted stays drifted until the repair pass exists. Tracking that here.

All the best,

Wayland

FerroxLabs:
Fixed in v0.9.8: project WCore conversations are pinned to the project workspace at create and at the spawn chokepoint (enforceProjectWorkspace).

PR #1011, opened 2026-08-17
fix(projects): re-read project knowledge at agent spawn

### LINKED ISSUE #999: Project knowledge is frozen at conversation creation, and Teams have no project to inherit it from
## Two related gaps in project knowledge, both promised to a user and neither tracked until now

Filing these because they were committed to in conversation and never made it into the tracker. Both were verified against main before writing.

## 1. Project knowledge is frozen at conversation creation

Editing a project's knowledge — .wayland/CONTEXT.md — does not reach a conversation that is already open. The injection happens once, when the conversation is created, so a user who updates their project context has to start a new chat before the agent sees it.

That is not how anyone expects it to work. You edit the project's context precisely because the current conversation needs it.

## 2. A Team cannot inherit its project's knowledge, because a Team has no project

Verified on main: TTeam (src/common/types/teamTypes.ts:76-77) carries **workspace: string** and **workspaceMode** — a filesystem path and a mode. There is **no projectId** anywhere on the type.

So a Team is bound to a directory, not to a project. There is no link to follow to find the project's knowledge, which makes "Teams inherit their project's knowledge" not a missing feature so much as a missing relationship.

## Why these belong in one issue

They are the same root shape: **project knowledge is bound at the wrong moment and to the wrong object.** Fixing (1) alone leaves Teams with nothing to inherit. Fixing (2) alone gives Teams a project whose knowledge still freezes at creation.

## Relationship to #924

#924 is the mirror image of this and should be designed alongside it. There, **global** memory leaks into every project because the per-project scope resolver (getActiveBrainScope) is dead code with zero callers and the project store is never consulted. Here, **project** knowledge fails to reach the places that should have it.

Same underlying absence: nothing reliably resolves "which project is this conversation actually in", so one store over-applies and the other under-applies. **Whoever picks up #924 should read this issue first** — a scope resolver that fixes one should fix both, and building two separate answers to the same question is how this drifts again.

## Suggested sequencing

1. Establish a single authoritative project-scope resolver in the main process (#924 needs this too)
2. Give TTeam a project binding rather than only a workspace path
3. Make knowledge injection re-read on change rather than only at creation

Sizes are M, M and S respectively, but (1) is shared with #924 so it should be costed once.

FerroxLabs:
Fixes #999 in part. See "What is and is not fixed" below.

## The defect

A project chat's .wayland/ knowledge (CONTEXT.md, rules.md, decisions.md) was composed into that conversation's system-rules channel exactly once, in ConversationServiceImpl.createConversation, and then persisted into the conversation row as extra.presetRules / extra.presetContext. Nothing ever re-read it.

So a user who edits their project knowledge to fix a mistake never reaches the conversation they are in. Not on the next message, and not even after an app restart, because the frozen string is what got saved. The only remedy was starting a brand new chat.

## Established by execution

The regression suite drives the real spawn seam, WorkerTaskManager.getOrBuildTask, which is where every backend's manager is handed the conversation's extra. Run against the pre-fix source (parent commit restored on disk) it fails on exactly the frozen behaviour:

- an edit to CONTEXT.md after creation reaches the next agent spawn: expected the spawned rules to contain the new text, got the text as of creation
- emptied project knowledge is removed from the system-rules channel: the deleted block was still there
- a chat moved into a project after creation gets its knowledge at spawn: got the base rules only

The negative controls in the same file pass before and after, which is the known-positive check that the harness is really exercising the seam rather than failing for an unrelated reason.

## The fix

One shared helper composes the block, and the block is delimited by sentinels so it can be found and REPLACED rather than only appended. That is what makes the composition safe to run more than once.

It runs at creation, as before, and again in WorkerTaskManager immediately beside the existing #30 workspace correction, on the same persist-if-changed path, so the refresh also survives a restart.

Two related gaps close for free. ProjectServiceImpl.assignConversation and ConversationServiceImpl.createWithMigration both stamp extra.projectId without ever injecting knowledge, so a re-parented or migrated chat had none at all. Both now pick it up at spawn.

## Block boundaries

The block is delimited by a header and a matching footer, and removal cuts header-to-footer.

The separator blocks are joined with is a plain markdown thematic break. A --- in the user's own CONTEXT.md produces exactly those bytes, and substantive() passes it straight through into the composed block, so using the separator as the boundary fragments the block: only the leading fragment gets removed and everything after the user's own --- survives as an orphan. That would have left the corrected content in the prompt forever, grown extra on every spawn with a SQLite write each time (compounding with workflowAdvanceReset, which respawns once per workflow step), and missed the prompt cache on every spawn. Sentinels cannot appear in user prose, so none of that is reachable.

Both sentinel literals are stripped out of composed body text, so a hand-typed or pasted sentinel inside CONTEXT.md cannot close the block early.

The header literal is byte-identical to what the original injection wrote, so conversations created before the footer existed are still recognised. Those legacy blocks have no footer, so they are cut from the header to the start of the next separator-joined block, or to the end of the string. The global memory block is the only thing ever appended after project knowledge and it always opens with a bracket, so the legacy cut cannot swallow it. If a user's knowledge contains a thematic break followed by a line opening with a bracket the legacy cut stops early and leaves one stale fragment, which is bounded and one-shot because the replacement block carries a footer and every later refresh is exact. That is the safe direction: overshooting would silently delete the conversation's global-memory snapshot.

No migration, no schema change.

## Failure direction

A project lookup or a knowledge read that throws returns false without stripping, leaving the existing block in place. Losing the agent's context because a disk read blipped would be the wrong failure.

readIfExists previously collapsed every read error into an empty string. Only ENOENT actually means "no document": EIO, EACCES, or a read landing mid-write (writeProjectKnowledge truncates in place rather than writing a temp file and renaming) were indistinguishable from the user clearing the file, and the refresh would then strip the knowledge and persist the loss. Non-ENOENT errors now rethrow so the caller takes the keep-what-we-have branch.

A chat with no projectId is never touched and never triggers a lookup, so non-project chats do no extra work. The refresh sits after the task-cache check, so a cache hit does no disk IO.

## What is and is not fixed

Fixed: the knowledge is re-read whenever the agent for that conversation is BUILT. That is an app restart, the idle-agent rebuild, an explicit stop, a skipCache respawn, or the first spawn after the edit.

Not fixed: the next TURN of an agent that is already live. #999 is worded as knowledge never reaching an already-open conversation, and that is satisfied at the next spawn, not the next turn. A user who edits knowledge and immediately messages the same conversation whose agent is still running keeps the stale block until that agent is killed, which is the 5-minute idle timeout by default. Mutating a live engine session is backend-specific (wcore skips re-sending rules on resume, gemini binds at agent init, acp pushes presetContext only on the first message) and is deliberately not attempted here.

Out of scope: the Teams half of #999 (a Team carries a workspace path and has no projectId, so it has no project to inherit knowledge from). That is missing capability rather than broken behaviour, it needs a schema change, and this swarm is repairing defects only. #999 should stay open for both remaining items.

## Verification

- targeted regression file: 10 passed, 0 failed. Fail-before proven for every claim by restoring the pre-fix source on disk: 3 fail for the freeze, 2 more for the separator collision (OLD VALUE surviving in both the thematic-break and legacy-block cases), 1 more for the swallowed read error (the block stripped down to the assistant's base rules)
- surrounding suites, 153 tests across ConversationServiceImpl, both WorkerTaskManager suites, projectKnowledge and its redteam and reference-confinement suites, globalMemoryBlock (unit and integration), projectWorkspace, ensureProjectWorkspace, projectServiceWorkspace, projectMessages, buildAgentConversationParams, createConversationParams, wcorePresetRulesFallback, team-workspace-sync, projectBridge generateDraft and confinement, localeKeyParity: all passed
- tsc --noEmit: no errors
- oxlint on the touched files: 0 errors
- oxfmt --check on exactly the touched files: correct format
- no locale keys added or changed (this change has no user-facing string)
- no existing test relaxed, skipped or deleted


FerroxLabs:

Run on 2026-08-17 against the pre-fix head; the verdict was never posted here. Recording it because this one is security-relevant.

**Verdict: FIX-FIRST.** Findings:

1. HIGH. For conversations created before this PR, project knowledge the user DELETED stayed in the system prompt permanently. Demonstrated with a rotated credential that survived four agent spawns. The diff's own comment described the condition as bounded and one-shot; it is neither, because the orphaned block has no header for the healing pass to match on.
2. The read is unbounded, at roughly 2.2 ms of main-thread time per megabyte, while the sibling memory composer in the same file applies three separate caps.

One fix sketch was refused on executed grounds: a bare lastIndexOf of the separator followed by an open bracket leaves the orphan in place when a legacy block has a dated body and no memory block.

A fix landed at head 85ad4f965. That fix has NOT yet been audited; a delta audit is running now. This PR is not cleared for merge until it reports.

Note for anyone picking this up: auto-merge was enabled on this PR while these HIGHs were open and a fix was still in flight. It has been disabled. Do not arm a PR before its audit verdict is GO.

FerroxLabs:

PR #1012, opened 2026-08-17
ci: check out the @wayland MCP sources and arm the #940 gate

### LINKED ISSUE #940: Bug report: Cannot connect Apple Ecosystem MCP - resource file not present in app bundle

FerroxLabs:
Fixes #940

## What was wrong

The four bundled @wayland MCP connectors (apple-mcp, cal-com-mcp, imap-mcp, news-mcp) build from a separate repo that had never been pushed. PR #993 made a missing connector FATAL instead of a silent warning, but to keep CI green it had to set WAYLAND_ALLOW_MISSING_MCP=1 in six workflows. The gate was armed and immediately bypassed everywhere, which is the exact state it was written to end.

FerroxLabs/waylandmcp now exists (private) with a read-only deploy key provisioned as the WAYLANDMCP_DEPLOY_KEY secret, so the checkout can be real and the bypass can go.

## How the sources reach the build

Four workflows actually invoke scripts/build-mcp-servers.js. All four now check the sources out and go RED if they cannot:

- _build-reusable.yml build job, via scripts/build-with-builder.js
- build-matrix.yml build job, via the same
- build-and-release.yml publish-getwayland-npm job, via installer build-payload.mjs
- publish-npm.yml publish job, via the same

Each does the same two steps: actions/checkout with the deploy key to ./waylandmcp, then a step that MOVES that tree to the sibling position and installs there.

The move is the interesting part, and it is a containment fix rather than tidiness. esbuild resolves a bare import by walking node_modules UPWARD from the entry point. With the connector tree inside the workspace, that walk continues into the desktop app's ~2100-package node_modules. A connector importing something it does not declare, or declares at a different major, would then resolve SILENTLY against the app's copy instead of failing. Green build, wrong artifact, which is precisely the quiet substitution #940 exists to stop, and the new source-path log line would not catch it because the tree is correct while the dependency is not.

Proven by A/B on identical source, same script, only the layout differing:

- in-workspace: an undeclared import of a package only the app tree has bundles CLEAN
- sibling: the same source fails with Could not resolve

actions/checkout refuses any path outside github.workspace, which is why the checkout cannot go straight to the sibling position. So it lands in-workspace and is moved. That makes an in-workspace candidate unnecessary, and it is removed: build-mcp-servers.js keeps exactly the three sibling candidates it always had, so local dev is untouched and WAYLAND_MCP_SRC remains authoritative-and-single. The rm -rf before the move covers a self-hosted runner, where the sibling path sits outside the workspace git clean.

The script now also logs which tree each connector was bundled from. Multiple candidates resolving silently is the same failure class, and it is only visible if it is printed.

/waylandmcp/ is gitignored. That entry turned out to be load-bearing well beyond committability: scripts/platform-package-smoke.mjs:214 throws on ANY non-ignored untracked path and runs after the connector checkout, so without the ignore a leftover tree would hard-red every packaged build.

## Dependency layout

Verified from a clean clone, and the CI log matches byte for byte (254 packages installed, printing only the two root devDependencies): bun puts the real packages in waylandmcp/node_modules/.bun/ and creates packages/<pkg>/node_modules with symlinks into that store. Nothing is hoisted to the waylandmcp root. So the nodePaths per-package hint resolves, and the upward walk stops inside waylandmcp rather than reaching the app tree.

## Where the checkout deliberately does NOT go

pr-checks.yml loses its bypass line and gains nothing, and this PR's own CI run is what proved why. pr-checks runs electron-vite in PRODUCTION mode, and electron.vite.config.ts registers buildMcpServersPlugin only when mode is development, so nothing in that workflow has ever invoked build-mcp-servers.js. Its bypass was never load-bearing. An earlier revision of this PR did add the checkout there; the run showed the connector-resolution lines never appear, so it was removed rather than left feeding a build that ignores it.

That is also the whole answer on fork PRs. A fork gets no repo secrets and now needs none: the only workflow a fork can trigger does not build the connectors. Every path that does is push, tag, schedule or workflow_dispatch on this repo. So a release build can never silently ship without the connectors, and there is no bypass left anywhere in CI to make it do so.

build-manual.yml also only loses its bypass line. It has no build steps of its own and caller-level env does not propagate into a reusable workflow, so that line was already dead.

## Verification

- Bypass settings in .github/workflows: 6 before, 0 after. Three textual hits remain repo-wide: the constant in build-mcp-servers.js, its unit test, and one explanatory comment in pr-checks.yml.
- actionlint 1.7.12 across all six files: 67 findings before, 67 after. Zero new, including on the new bash run block. All six also parse with the yaml package.
- LIVE CI PROOF: on this PR's Security Audit E2E run an earlier revision carried the checkout, and Detect access, Checkout @wayland MCP connector sources and Install @wayland MCP connector dependencies all reported success on a real ubuntu runner. The log shows git remote add origin git@github.com:FerroxLabs/waylandmcp.git followed by a successful fetch, so the deploy key, the secret and the SSH path are confirmed. persist-credentials false is honoured with no key material in the log.
- End-to-end replay of the exact workflow steps in a CI-shaped tree (clone to ./waylandmcp, then the literal run block, then the bundle): all four connectors resolved from the sibling position and all four .mjs artifacts were produced.
- Containment A/B as described above, executed both ways.
- Fatal-vs-bypass: with no candidate reachable, the bundle throws and names the opt-out; with WAYLAND_ALLOW_MISSING_MCP=1 it warns and returns.
- Local dev fallback: with no in-repo tree, all four resolve from the ~/dev sibling.
- buildMcpServersGate.test.ts: 8 passed. The new case pins the containment invariant, that no candidate resolves inside the repo root, and was mutation-checked: reintroducing an in-workspace candidate turns it red.
- bunx tsc --noEmit clean. bun run lint (oxlint) exits 0 with zero findings in the touched files. oxfmt over exactly the touched files produced no changes.

## Not verified without a real CI run

- The ssh-key checkout on the Windows and macOS legs. Proved on ubuntu by this PR's run only.
- build-matrix.yml will NOT close that gap on merge as things stand: it is currently red on main. In run 31993152345 four legs went green, macos-x64 and both preview Linux legs failed at Run packaged build, and all four Windows legs were cancelled.
- The cheapest way to retire both items is a single build-manual.yml dispatch across all six platforms. It goes through _build-reusable with secrets inherit, which is the exact release path, and it neither tags nor publishes.

## Split out

#1013: apple-mcp ships without its Swift EventKit bridge. waylandmcp gitignores dist/, so the binary is never in a CI checkout, and the copy in build-mcp-servers.js was a silent no-op. The connector still ships and still advertises 5 Calendar and 4 Reminders tools that route through that binary and cannot work; Notes, Mail, Maps and Photos use AppleScript and are fine. Before this PR no Apple connector shipped at all, so this is newly reachable rather than pre-existing. This PR does the minimum, turning the silent no-op into a loud build warning naming the connector and the absent binary. Actually building the Swift bridge on the macOS legs is #1013.

### DIFF
--- a/.github/workflows/_build-reusable.yml
+++ b/.github/workflows/_build-reusable.yml
@@ -45,13 +45,6 @@ env:
   # PINNED_BUN_VERSION in scripts/prepareBundledBun.js and the
   # corresponding entry in scripts/bundled-bun-shasums.json.
   WAYLAND_BUN_VERSION: '1.3.14'
-  # #940: four @wayland MCP connectors (apple, imap, news, cal.com) build from
-  # the sibling waylandmcp repo, which has no GitHub remote yet - CI cannot
-  # check it out. scripts/build-mcp-servers.js now FAILS the build when a
-  # connector cannot be bundled, so CI must opt out explicitly until that repo
-  # lands. DELETE this line the moment waylandmcp is checkout-able in CI; that
-  # is the whole point of the gate.
-  WAYLAND_ALLOW_MISSING_MCP: '1'
 
 jobs:
   code-quality:
@@ -243,6 +236,30 @@ jobs:
           max_attempts: 3
           retry_wait_seconds: 30
           command: bun install --frozen-lockfile
+
+      # #940: the four bundled @wayland MCP connectors (apple, imap, news,
+      # cal.com) build from FerroxLabs/waylandmcp. actions/checkout refuses any
+      # `path` outside github.workspace, so the sources land at ./waylandmcp
+      # (gitignored) - the first candidate scripts/build-mcp-servers.js resolves.
+      # Must run AFTER the repo checkout above: that checkout's `git clean -ffdx`
+      # would delete this tree. There is deliberately NO bypass here - if this
+      # fails, the release build goes red rather than shipping a Library that
+      # offers four connectors it does not contain.
+      - name: Checkout @wayland MCP connector sources
+        uses: actions/checkout@v6
+        with:
+          repository: FerroxLabs/waylandmcp
+          ssh-key: ${{ secrets.WAYLANDMCP_DEPLOY_KEY }}
+          path: waylandmcp
+          persist-credentials: false
+
+      # esbuild resolves connector deps from the PER-PACKAGE node_modules
+      # (nodePaths in scripts/build-mcp-servers.js). bun's isolated linker does
+      # create packages/*/node_modules symlinks into the root store, so a plain
+      # workspace install at the root is enough.
+      - name: Install @wayland MCP connector dependencies
+        working-directory: waylandmcp
+        run: bun install --frozen-lockfile
 
       - name: Run postinstall
         run: bun run postinstall || true
--- a/.github/workflows/_build-reusable.yml
+++ b/.github/workflows/_build-reusable.yml
@@ -303,7 +303,20 @@ jobs:
           [[ "$mcp_sha" == "$WAYLANDMCP_REF" ]] || {
             echo "::error::waylandmcp HEAD $mcp_sha != pinned $WAYLANDMCP_REF" >&2; exit 1;
           }
-          bun install --frozen-lockfile
+          # Retried because this is an unattended network op on the release and
+          # publish paths, four lines below an app-side `bun install
+          # --frozen-lockfile` that IS wrapped in nick-fields/retry@v3. The
+          # checkout above needs no wrapper: its only network call (git fetch)
+          # already runs inside actions/checkout's own 3-attempt retry helper.
+          # Loop shape matches pr-checks.yml - a genuine failure still exits
+          # non-zero after the third attempt, so a retry cannot mask one.
+          for attempt in 1 2 3; do
+            if bun install --frozen-lockfile; then exit 0; fi
+            echo "waylandmcp bun install attempt $attempt failed; retrying" >&2
+            sleep 30
+          done
+          echo "waylandmcp: bun install --frozen-lockfile failed after 3 attempts" >&2
+          exit 1
 
       - name: Run postinstall
         run: bun run postinstall || true
--- a/.github/workflows/_build-reusable.yml
+++ b/.github/workflows/_build-reusable.yml
@@ -240,11 +240,11 @@ jobs:
       # #940: the four bundled @wayland MCP connectors (apple, imap, news,
       # cal.com) build from FerroxLabs/waylandmcp. actions/checkout refuses any
       # `path` outside github.workspace, so the sources land at ./waylandmcp
-      # (gitignored) - the first candidate scripts/build-mcp-servers.js resolves.
-      # Must run AFTER the repo checkout above: that checkout's `git clean -ffdx`
-      # would delete this tree. There is deliberately NO bypass here - if this
-      # fails, the release build goes red rather than shipping a Library that
-      # offers four connectors it does not contain.
+      # (gitignored), and the next step moves it to the sibling position that
+      # build-mcp-servers.js already resolves. Must run AFTER the repo checkout
+      # above: that checkout's `git clean -ffdx` would delete this tree. There is
+      # deliberately NO bypass here - if this fails, the release build goes red
+      # rather than shipping a Library that offers four connectors it lacks.
       - name: Checkout @wayland MCP connector sources
         uses: actions/checkout@v6
         with:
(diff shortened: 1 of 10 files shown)


wayland:
STOP TAGGING ME!

FerroxLabs:

Run on 2026-08-17; the verdict was never posted here. Recording it so the PR carries its own evidence.

**Round 1 verdict: FIX-FIRST.**

What is genuinely fixed, verified by execution: the issue #940 containment check is correct on 6 of 6 shapes. The old predicate was wrong on 3 of those 6, and mutation M4 reds as expected.

What was open: the connector checkout is unpinned and unattested on all four publish paths, so a shipped installer can contain code from a tree whose identity appears in no receipt. The /waylandmcp/ path is gitignored, so the dirty-tree gate cannot see it either.

**Round 2 (delta audit of the fix) verdict: GO.** Head 1e83193e2 is cleared on audit grounds.

One recommendation raised during the audit was refused on executed grounds, and correctly so: adding a hand-rolled git clone retry buys nothing, because actions/checkout already retries fetch three times in retry-helper.ts, and input-helper.ts:92-94 maps a 40-hex ref straight to result.commit so getDefaultBranch is never reached.

The hardcoded drift-guard list noted in review is tracked separately and is not a merge blocker.

FerroxLabs:

Issue #1008, opened 2026-08-17
Bundled first-party MCP servers expose 0 tools at runtime in shipped builds


FerroxLabs:
## What happens

Two bundled first-party MCP servers report **Enabled but exposes 0 tools** at runtime in the shipped app.

## Evidence: two independent reporters, identical result

Both from in-app bug-report diagnostics, both on 0.11.18 / macOS arm64, from two different users who filed for unrelated reasons (#947 and #931). Neither was reporting MCP problems, so this is incidental capture rather than someone chasing a known symptom:

    MCP servers (mcp.config): 5
    - wayland-search-skills — Enabled but exposes 0 tools
    - wayland-concierge-diag — Enabled but exposes 0 tools
    - com.microsoft-playwright-mcp — Enabled but exposes 0 tools

The third-party playwright entry may be a separate cause and is noted only for completeness. The two @wayland entries are the defect.

## Why this is not #940

These are **not** the optional sibling-repo extras. They are first-party core builtins bundled from src/process/resources/builtinMcp/, emitted by scripts/build-mcp-servers.js as builtin-mcp-search-skills.js and builtin-mcp-concierge-diag.js. Per that script's own comment the first-party core servers **hard-fail the build** if they do not bundle, and the release built green.

So the artifacts are present in the shipped app and the servers still register nothing. That makes this a runtime connect/registration failure, not a packaging gap, and it is a different defect from the four missing sibling connectors in #940.

## What is not yet established

No root cause. Specifically unknown:

- whether the server process spawns at all, fails to spawn, or spawns and registers an empty tool list
- whether it is path resolution to the bundled .js inside the packaged asar, a runtime dependency missing from the bundle, or a handshake or timing failure
- whether it reproduces on a clean profile or needs prior state
- whether Windows and Linux are affected, since both data points are macOS arm64

A prior note in this project records that a fresh dev worktree can make search-skills crash-loop with -32000 unless the MCP build script has been run. That is a dev-tree condition and should **not** be assumed to be the same cause here, since these users are running a packaged build where the bundle step demonstrably succeeded. Worth checking whether the two failure modes share a mechanism.

## Suggested first steps

1. Launch a packaged build against a scratch profile and read the main-process log for the spawn of each builtin.
2. Confirm the bundled .js files are actually present inside the shipped app bundle and resolvable at the path the spawn uses.
3. Establish whether the process starts and returns an empty tool list, or never starts.

## Impact

[...]

PR #1015, opened 2026-08-17
fix(mcp): run first-party builtin MCP servers on a real JS runtime

FerroxLabs:
Refs #1008. Fixes the reported symptom and one real spawn defect; does NOT close the issue — see "What remains" below.

## Root cause (established by execution, not by reading)

The three first-party bundled MCP servers are seeded into mcp.config as
{ command: 'node', args: [<absolute path to out/main/builtin-mcp-*.js>] }
and were spawned with that bare command.

macOS ships no /usr/bin/node. On any Mac without a developer Node install the
probe dies with ENOENT, so the server's tool inventory is never written back to
mcp.config, and the in-app diagnostics report "Enabled but exposes 0 tools"
forever. That is exactly the line both reporters produced.

A carve-out that routes our own builtins through a resolved JS runtime already
existed in the probe path (bundled Bun when packaged, Electron-as-Node in dev),
added for precisely this failure mode. It only matched the four sibling @wayland
servers, which store a BARE FILENAME in args[0]. The core builtins store an
absolute path, so the filename match never saw them.

## Evidence

Run against the bundled artifacts in out/main, driving a real MCP handshake
(initialize, notifications/initialized, tools/list):

- builtin-mcp-search-skills.js under node: 2 tools (wayland_search_skills, wayland_read_skill)
- builtin-mcp-concierge-diag.js under node: 1 tool (wayland_concierge_diag)
- all three under bun 1.3.11: same tool counts
- spawn of bare "node" with the PATH a Finder-launched app inherits
  (/usr/bin:/bin:/usr/sbin:/sbin): spawn node ENOENT

So the artifacts are fine and the runtime selection is the defect.

The app ships a bundled Bun and no Node (electron-builder.yml extraResources),
which is why Bun is the correct runtime here.

## concierge-diag and better-sqlite3

Verified: under Bun, require('better-sqlite3') resolves but opening a database
throws "not yet supported in Bun". openReadonlyDb already catches that and
degrades, so the sqlite-backed sections report "db unavailable" while the
config- and log-backed sections still work, and a tool call returns a report
rather than throwing. This is not a regression: the packaged binding is built
for Electron's ABI, so a system node cannot load it either (reproduced locally
as NODE_MODULE_VERSION 145 vs 127).

## Diagnostic message

The old flag told every zero-tool server to "check its command, args, or
credentials". A bundled builtin has none of those that the user owns, so that
advice sent two reporters hunting for a mistake they could not have made.
Bundled builtins now say so and ask for a report. The string lives in the
standalone stdio server and the bug-report collector, not in the i18n system,
so no locale keys change; localeKeyParity was run and is green.

## Fail-before proof

Source files reverted to HEAD via git show (no stash), new tests kept:

- mcpProtocol.test.ts, 3 new tests FAIL: "expected 'node' to be
  '/res/bundled-bun/darwin-arm64/bun'", "expected 'node' not to be 'node'",
  "expected 'node' to be '<execPath>'"
- conciergeDiagServer.test.ts, 1 new test FAIL: flag did not contain
  "bundled with Wayland"
- builtinMcpConstants.test.ts, new tests FAIL: isBuiltinCoreMcpArg is not a function

After restoring the fix: 70 pass, 0 fail across the three files.
Wider sweep tests/unit/process/services/mcpServices + tests/unit/process/resources/builtinMcp:
188 pass, 0 fail. Typecheck clean.

## What remains (do not close on this PR)

This fixes the PROBE path, which is what writes the tool inventory the
diagnostic reads. The LIVE SESSION injection paths still forward command 'node'
verbatim, so on a Mac with no Node the builtins remain uncallable in an actual
chat. Twelve such paths were mapped, including McpConfig.projectStorageConfig,
buildAcpSessionMcpServers, buildGeminiStdioMcpConfig, buildCodexMcpServerTable
and the six per-CLI config writers. None of them call resolveJsRuntime, and none
carry the runtime env it returns (which is load-bearing in dev).

I did not change them because three existing tests explicitly assert the current
behaviour, and the repo rule is to report rather than relax a test:

- tests/unit/acpBuiltinMcp.test.ts:52 and :174 assert command 'node'
- tests/unit/claudeMcpAgent.test.ts:46 asserts command 'node'

That needs its own decision, because the session serializers pass no env and
would need the ELECTRON_RUN_AS_NODE half of resolveJsRuntime threaded through
for dev to keep working.

Also unresolved and worth a follow-up: formatDiagnostics renders
server.flag ?? server.lastError, so the flag masks the real error. Had it shown
lastError too, this issue would have been root-caused from the two reports
directly. And the third-party playwright entry in the same reports is a separate
cause, untouched here.

## Not reproduced end-to-end on a packaged build

Stated plainly: I could not reproduce the user-visible failure on this machine,
because this machine HAS Node on the login shell PATH, which getEnhancedEnv
hydrates. A packaged build here would spawn successfully and show tools. The
causal chain was instead verified by its parts, as above. What would settle it
from the reporters: whether "node --version" works in their Terminal, and the
lastError value for those two servers in the MCP Library.

### DIFF
--- a/src/process/acp/session/McpConfig.ts
+++ b/src/process/acp/session/McpConfig.ts
@@ -143,7 +143,9 @@ export class McpConfig {
           // Same runtime tuple as the Library probe: `npx`→bundled Bun (#827)
           // AND Wayland's own bundled MCP servers→resolved JS runtime (#1008),
           // carrying the runtime env the dev runtime needs to BE a Node runtime.
-          const spawn = resolveSessionMcpStdioSpawn(server.transport.command, server.transport.args ?? []);
+          const spawn = resolveSessionMcpStdioSpawn(server.transport.command, server.transport.args ?? [], {
+            libraryEntryId: server.libraryEntryId,
+          });
           runtimeServer = {
             name: server.name,
             command: spawn.command,
--- a/src/process/acp/session/McpConfig.ts
+++ b/src/process/acp/session/McpConfig.ts
@@ -2,7 +2,7 @@
 import type { IMcpServer } from '@/common/config/storage';
 import type { AcpMcpCapabilities } from '@/common/types/acpTypes';
 import type { McpServer } from '@agentclientprotocol/sdk';
-import { resolveMcpStdioSpawn } from '@process/services/mcpServices/mcpStdioSpawn';
+import { mergeMcpSpawnEnv, resolveSessionMcpStdioSpawn } from '@process/services/mcpServices/builtinMcpRuntime';
 import { isServerActiveForSession } from '@process/agent/acp/mcpSessionConfig';
 import {
   createMcpSessionState,
@@ -140,12 +140,15 @@ export class McpConfig {
             reason = 'ACP runtime did not advertise stdio MCP transport support';
             break;
           }
-          const spawn = resolveMcpStdioSpawn(server.transport.command, server.transport.args ?? []);
+          // Same runtime tuple as the Library probe: `npx`→bundled Bun (#827)
+          // AND Wayland's own bundled MCP servers→resolved JS runtime (#1008),
+          // carrying the runtime env the dev runtime needs to BE a Node runtime.
+          const spawn = resolveSessionMcpStdioSpawn(server.transport.command, server.transport.args ?? []);
           runtimeServer = {
             name: server.name,
             command: spawn.command,
             args: spawn.args,
-            env: toNameValueArray(server.transport.env),
+            env: toNameValueArray(mergeMcpSpawnEnv(server.transport.env, spawn.env)),
           };
           break;
         }
--- a/src/process/agent/acp/mcpSessionConfig.ts
+++ b/src/process/agent/acp/mcpSessionConfig.ts
@@ -10,6 +10,7 @@ import type { IMcpServer } from '@/common/config/storage';
 import type { AcpMcpCapabilities } from '@/common/types/acpTypes';
 import { BUILTIN_CONCIERGE_DIAG_ID } from '@process/resources/builtinMcp/constants';
 import { resolveMcpStdioSpawn } from '@process/services/mcpServices/mcpStdioSpawn';
+import { mergeMcpSpawnEnv, resolveSessionMcpStdioSpawn } from '@process/services/mcpServices/builtinMcpRuntime';
 import { sanitizeMcpServerName } from '@process/services/mcpServices/validateMcpServer';
 
 export interface AcpSessionMcpNameValue {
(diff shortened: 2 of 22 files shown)


wayland:
STOP TAGGING ME!

FerroxLabs:

Four independent legs on head 15693c412 (Codex 5.6 Sol, Gemini 3.1 Pro, Kimi K3, plus an
internal reviewer). Two legs returned FIX-FIRST, one returned GO. Fail-before was confirmed
real: 10 tests fail against d82e7a448 with the new test files applied. No existing assertion
was relaxed, skipped or deleted; no locale key changed; the bridge deny-list is byte-identical
at 144 entries. All three builtins were verified live under bun 1.3.11 with a real
initialize / tools/list / tools/call handshake, so the core mechanism works.

Two findings must be fixed before merge.

**F1 -- the probe goes green while chat stays broken.** The probe now spawns under the resolved
runtime and persists its tool count, which clears the health flag, but the live-session
serializers still emit command "node" (mcpSessionConfig.ts:117, McpConfig.ts:143,
GeminiAgentManager.ts:93). On the reporters' machines that yields a green MCP Library, a green
concierge-diag, builtins still uncallable in an actual chat, and the new "bundled with Wayland"
message unreachable in exactly the case it was written for. That is worse than today's loud,
correct failure. The comment at McpProtocol.ts:337-341 -- eight lines below the new code --
already forbids this split.

The three assertions pinning command "node" (acpBuiltinMcp.test.ts:51 and :174,
claudeMcpAgent.test.ts:45) are being updated, not relaxed. They hard-code the defect rather
than an invariant. The no-relaxing rule protects live security checks from being weakened to
satisfy a fixture; it does not freeze a bug in place.

**F2 -- basename-only matching hijacks a user-configured server.** A user's own server that
merely shares our basename is silently re-pointed onto our runtime. Same file and args, so not
privilege escalation, but a silent substitution on a file the user owns. The negative test at
builtinMcpConstants.test.ts:124 uses a .ts extension, so it cannot fail; with .js it fails
today. Fix is an exact match against our own resolved script path.

## Deferred, tracked

- #1017 -- win32 without a bundled Bun falls back to node.exe and can ENOENT. Pre-existing
  since #438 / #706, not introduced here.
- #1018 -- concierge-diag loses three sqlite-backed sections under Bun. Verified live, not a
  regression: the Electron-ABI binding will not load under a system Node either.
- #1019 -- the better-sqlite3 asarUnpack entry is now load-bearing for builtin MCP under Bun
  and no test asserts it. Removing that line would arm an uncatchable panic that surfaces as
  "0 tools", pointing the next investigation at the wrong layer exactly as #1008 did.

Gemini's claim that an eager better-sqlite3 import kills the server at boot was refuted by
execution: the require resolves, the open throws catchably, and openReadonlyDb degrades.

FerroxLabs:

PR #1022, opened 2026-08-17
Unblock the pre-tag packaged-build gate on darwin-x64 and Linux preview

FerroxLabs:
## Why

The pre-tag packaged-build gate was RED on main, blocking the next release tag. Build Matrix run 32006041817 on main @ d82e7a448 exposed two independent defects, both in the post-packaging release gate rather than in the packaging itself.

## Defect 1 - bundled-constitution-fs failed verification on darwin-x64

verify-packaged-resources reported FAIL bundled-constitution-fs (CRITICAL) on both stable and preview macos-x64 legs, against a directory that contained a real 776,888-byte helper plus a matching manifest.json and package-authority.json.

Root cause, established by execution: verifyConstitutionFsBundle ran codesign --verify --strict on every darwin build unconditionally. A build with no Developer ID identity deliberately stages the helper UNSIGNED (see signDarwinStagedBinary, the sign-at-stage change), and ld64 only linker-signs arm64 - never x86_64. So darwin-arm64 passed on the linker's ad-hoc signature and darwin-x64 could never pass.

Executed evidence:

- Real Mach-O built by the same linker: x86_64 gives "code object is not signed at all", exit 1; arm64 gives "adhoc,linker-signed", exit 0.
- Real bundle fixtures through the real verifier: darwin-x64 unsigned threw at verify-packaged-resources.js:172 inside the codesign call; the SAME x86_64 bytes ad-hoc signed returned true; darwin-arm64 returned true in both states. The signature call was the only discriminator - digest, size, manifest shape, package-sealed authority and arch identity all already passed on exactly what CI produced.

The function also declared five parameters while its caller passed six, silently dropping requireDarwinSignature. Wiring that through matches the policy verifyWCoreBundle and verifyWNanoBundle already implement, and makes the SIGNED release path stricter than before: the old check accepted any signature including ad-hoc, and the new one demands a real Ferrox Labs Developer ID signature - ad-hoc is precisely what Apple refuses to notarize. Byte identity is untouched; the digest, size, package-sealed authority and the runtime re-hash against app.asar still pin these exact bytes.

I fixed the verifier, not the packaging, because the artifacts were provably complete and correct and the build's own signing contract already declares when a signature is required.

## Defect 2 - both Linux preview legs produced no packaged app

Both preview Linux legs failed with "expected exactly one current linux-x64 packaged app under .../out-preview; found 0. Inventory: <empty>" while electron-builder had in fact packaged out-preview/linux-unpacked and built the deb and rpm. Both stable Linux legs passed.

Root cause, established by execution: findPackagedCandidates matched posix executables against /^wayland(?:-preview)?$/, but electron-builder writes executableName verbatim, so the Preview overlay ships an executable literally named "Wayland Preview" - with a SPACE. The CI log confirms it: "Applying Electron fuses to .../out-preview/linux-unpacked/Wayland Preview". Stable passed only because "Wayland" matches the base alternative. The win32 branch of the same regex already allowed the space; only the posix branch carried the wrong assumption, and "wayland-preview" appears nowhere in the repo except as the URL protocol scheme.

Reproduced against the real layout with a known positive as control: stable / "Wayland" resolved, preview / "Wayland Preview" produced the exact CI error. After the fix both resolve.

The same wrong assumption sat in expectedReleaseIdentity in scripts/platform-package-smoke.mjs and scripts/release-acceptance/verifyPlatformPackageSmokes.js, which compare that name against the installed payload - so fixing only the gate would have moved the failure one step later into "Install and smoke real package payload". Proven by executing resolveInstalledCandidate: preview found 0 before, 1 after. Both now derive the Linux executable name from the product name, exactly as electron-builder does. Track separation is unaffected - captureCandidateState still filters candidates by that exact name, so a stable payload is invisible to a preview capture and vice versa.

## Proof the gates can still FAIL

Every case below was executed against the FIXED code and each one still blocked the build.

Constitution helper:

- helper bytes appended (authority stale) - blocked
- helper binary deleted - blocked
- package-authority.json deleted - blocked
- manifest protocolVersion downgraded to 1 - blocked
- embedded authority sha disagreeing with the package-sealed authority - blocked
- foreign-arch binary staged under the runtime directory - blocked
- an extra foreign runtime directory alongside the real one - blocked
- unsigned helper with requireDarwinSignature set - blocked
- ad-hoc signed helper with requireDarwinSignature set - blocked (this one used to PASS)

Packaged app discovery:

- no packaged app at all - blocked
- resources directory removed - blocked
- wrong arch - blocked
- a stray second matching executable - blocked
- a foreign executable name - blocked

## Tests

New regression tests, both driven by the real production names:

- verifyPackagedResources: a parameterised test per arch asserting the signature check is skipped without an identity, invoked with one, and that the real Developer ID check refuses unsigned fixture bytes.
- platformPackageSmoke: the Linux executable name is bound to the product name on both tracks, and a Linux Preview packaged app resolves by its real name while a foreign name still yields "found 0".

The existing fixture that renamed a preview payload to "wayland-preview" now uses "Wayland Preview". No assertion was weakened - the fixture encoded a name production never emits.

## Scope notes

No signing pipeline changes, no security-shell changes, bridgeAllowlist.ts untouched, no migrations touched, no test relaxed or skipped.

### DIFF
--- a/scripts/platform-package-smoke.mjs
+++ b/scripts/platform-package-smoke.mjs
@@ -64,14 +64,11 @@ export function expectedReleaseIdentity(releaseTrack, targetPlatform, targetArch
   if (!VALID_RELEASE_TRACKS.has(releaseTrack)) throw new Error(`${TAG} invalid release track: ${releaseTrack}`);
   const preview = releaseTrack === 'preview';
   const productName = preview ? 'Wayland Preview' : 'Wayland';
-  const executableName =
-    targetPlatform === 'win32'
-      ? `${productName}.exe`
-      : targetPlatform === 'linux'
-        ? preview
-          ? 'wayland-preview'
-          : 'wayland'
-        : productName;
+  // Both electron-builder configs set `executableName` to the product name and
+  // electron-builder writes it verbatim, so Linux ships `Wayland` /
+  // `Wayland Preview` - it does not sanitize either into a lowercase hyphenated
+  // name. Comparisons against this value are case-insensitive.
+  const executableName = targetPlatform === 'win32' ? `${productName}.exe` : productName;
   const baseChannel = preview ? 'preview' : 'latest';
   const updateChannel =
     targetPlatform === 'win32' && targetArch === 'arm64'
--- a/scripts/release-acceptance/verifyPlatformPackageSmokes.js
+++ b/scripts/release-acceptance/verifyPlatformPackageSmokes.js
@@ -77,14 +77,9 @@ function expectedReleaseIdentity(releaseTrack, platform, arch) {
   }
   const preview = releaseTrack === 'preview';
   const productName = preview ? 'Wayland Preview' : 'Wayland';
-  const executableName =
-    platform === 'win32'
-      ? `${productName}.exe`
-      : platform === 'linux'
-        ? preview
-          ? 'wayland-preview'
-          : 'wayland'
-        : productName;
+  // Mirrors scripts/platform-package-smoke.mjs: electron-builder writes
+  // `executableName` verbatim, so Linux ships `Wayland` / `Wayland Preview`.
+  const executableName = platform === 'win32' ? `${productName}.exe` : productName;
   const baseChannel = preview ? 'preview' : 'latest';
   const updateChannel =
     platform === 'win32' && arch === 'arm64'
--- a/scripts/verify-packaged-resources.js
+++ b/scripts/verify-packaged-resources.js
@@ -218,8 +218,10 @@ function verifyConstitutionFsBundle(
   // helper unsigned - see signDarwinStagedBinary - and ld64 only linker-signs
   // arm64, never x86_64, so demanding *any* signature here failed every
   // darwin-x64 build that had no signing secrets. Byte identity is unaffected:
-  // the digest, size and package-sealed authority above still pin these exact
-  // bytes, and the app re-hashes the helper against app.asar at runtime.
+  // the digest and size checks above still pin these exact bytes, and the app
+  // re-hashes the helper at RUNTIME against the digest embedded in app.asar -
+  // which is what actually authenticates them, since package-authority.json
+  // ships inside the package alongside the manifest.
   if (targetPlatform === 'darwin' && requireDarwinSignature) {
     verifyDarwinSignature(binaryPath);
   }
--- a/scripts/verify-packaged-resources.js
+++ b/scripts/verify-packaged-resources.js
@@ -28,7 +28,11 @@ const fs = require('fs');
 const path = require('path');
 const crypto = require('crypto');
 const { execFileSync } = require('child_process');
-const { isDarwinDeveloperIdSigned, darwinSigningIdentifier } = require('./signDarwinStagedBinary');
+const {
+  assertDarwinDeveloperIdSigned,
+  isDarwinDeveloperIdSigned,
+  darwinSigningIdentifier,
+} = require('./signDarwinStagedBinary');
 const prepareWaylandCore = require('./prepareWaylandCore');
 const prepareWaylandNano = require('./prepareWaylandNano');
 const prepareOfficeCli = require('./prepareOfficeCli');
@@ -168,8 +172,8 @@ function verifyConstitutionFsBundle(
   targetPlatform,
   targetArch,
   authority,
-  verifyDarwinSignature = (binaryPath) =>
-    execFileSync('/usr/bin/codesign', ['--verify', '--strict', binaryPath], { stdio: 'pipe' })
+  verifyDarwinSignature = (binaryPath) => assertDarwinDeveloperIdSigned(binaryPath),
+  requireDarwinSignature = false
 ) {
   if (targetPlatform === 'win32') return !fs.existsSync(bundleRoot);
   const runtime = `${targetPlatform}-${targetArch}`;
@@ -207,7 +211,16 @@ function verifyConstitutionFsBundle(
     return false;
   const identity = inspectExecutable(binaryPath);
   if (identity?.platform !== targetPlatform || identity?.arch !== targetArch) return false;
-  if (targetPlatform === 'darwin') {
+  // Signature policy matches wcore/wnano: only a build that HAD a Developer ID
+  // identity must ship a signed helper, and then it must be a real Ferrox Labs
+  // Developer ID signature (ad-hoc / linker-signed is exactly what Apple
+  // rejects at notarization). A build with no identity deliberately stages the
+  // helper unsigned - see signDarwinStagedBinary - and ld64 only linker-signs
+  // arm64, never x86_64, so demanding *any* signature here failed every
+  // darwin-x64 build that had no signing secrets. Byte identity is unaffected:
+  // the digest, size and package-sealed authority above still pin these exact
+  // bytes, and the app re-hashes the helper against app.asar at runtime.
+  if (targetPlatform === 'darwin' && requireDarwinSignature) {
     verifyDarwinSignature(binaryPath);
   }
   return true;
@@ -266,9 +279,12 @@ function findPackagedCandidates(outDir) {
       .filter(({ executablePath, identity }) => {
         if (!identity) return false;
         const name = path.basename(executablePath).toLowerCase();
+        // electron-builder writes `executableName` verbatim on every platform,
+        // so the Preview overlay's `Wayland Preview` keeps its SPACE on Linux
+        // too - it is not sanitized to `wayland-preview`.
         return identity.platform === 'win32'
           ? /^wayland(?: preview)?\.exe$/.test(name)
-          : /^wayland(?:-preview)?$/.test(name);
+          : /^wayland(?: preview)?$/.test(name);
       });
     for (const { executablePath, identity } of executables) {
       candidates.push({ appDir: dir, resourceDir, executablePath, ...identity });
--- a/src/common/releaseTrack.ts
+++ b/src/common/releaseTrack.ts
@@ -36,6 +36,25 @@ export function getReleaseProtocolScheme(track: WaylandReleaseTrack): 'wayland' 
   return track === 'preview' ? 'wayland-preview' : 'wayland';
 }
 
+/**
+ * The basename of the packaged launcher for a track on a platform.
+ *
+ * electron-builder writes `executableName` verbatim, and both configs set it to
+ * the product name, so Linux ships `Wayland` / `Wayland Preview` - it is never
+ * sanitized into a lowercase hyphenated name. Assuming otherwise is what made
+ * both Linux Preview legs of Build Matrix report an empty packaged inventory.
+ *
+ * This is the single source the packaged runtime derives its boot-start
+ * `releaseIdentity.executableName` from. That event is compared byte-for-byte
+ * against `expectedReleaseIdentity` in scripts/platform-package-smoke.mjs, so
+ * the two derivations must never drift; tests/unit/releaseTrack.test.ts binds
+ * them together.
+ */
+export function getPackagedExecutableName(track: WaylandReleaseTrack, platform: NodeJS.Platform): string {
+  const { appName } = getPackagedReleaseIdentity(track);
+  return platform === 'win32' ? `${appName}.exe` : appName;
+}
+
 export type ReleaseRuntime = {
   platform: NodeJS.Platform;
   arch: string;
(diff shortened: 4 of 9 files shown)


FerroxLabs:
## Verification completed after the PR was opened

Real packaging entry point, not a shortcut. Ran WAYLAND_LOCAL_VERIFICATION=1 node scripts/build-with-builder.js auto --mac --dir end to end (the full prepackage chain: cargo build of the helper, wayland-core and wayland-nano staging, skill pack, electron-builder pack, fuses, afterPack, Developer ID bundle signing, then the gate). Result:

    [verify-packaged-resources]   OK   bundled-constitution-fs
    [verify-packaged-resources] PASS - all critical bundled resources present (2 optional warning(s)).
    Build completed, exit 0

That real packaged helper is Format=Mach-O thin (arm64), flags=0x20002(adhoc,linker-signed) - no Developer ID. codesign --verify --strict returns 0 on it, which is exactly why arm64 always passed the old gate and x86_64 never could. Running the fixed check against that same real artifact with --require-darwin-signature now BLOCKS it ("code failed to satisfy specified code requirement(s)"), where the old check would have accepted it - the release path got stricter, not looser.

Deliberate corruption of the REAL packaged bundle, each one still blocked:

- helper bytes tampered - false
- manifest protocolVersion downgraded to 1 - false
- package-authority.json deleted - blocked
- an extra foreign runtime directory added - false
- restored to original - true again

Tampering the packaged helper also trips the app-bundle signature check first (a sealed resource is missing or invalid), so there are two independent gates on that path.

Full unit suite on the final merged state: 18036 passed, 155 skipped, 3 failed. The 3 failures are all in tests/unit/process/task/acpAgentManagerNpmFallback.test.ts, a known host-environment failure caused by a real wayland-nano binary on PATH at ~/.local/bin/wayland-nano (confirmed present, 10.7M). Unrelated to this change and untouched by it.

oxfmt --check clean on the six changed files. tsc --noEmit clean.

Not verified locally, and honestly out of reach on an arm64 macOS host: the darwin-x64 leg itself (prepareConstitutionFs refuses cross-arch compilation by design, so only an Intel runner can prove it) and the Linux preview legs. Both were established by executing the real code paths against real Mach-O and real ELF layouts plus the CI logs, but CI is the final word on both.

FerroxLabs:
## Cross-audit verdict: FIX-FIRST (one blocker)

Legs: Codex 5.6 Sol FIX-FIRST (reached the same signed-release conclusion independently; missed the
blocker). Gemini 3.1 Pro NO-GO (found the blocker; its deb-sanitisation finding was refuted by
execution). Kimi K3 DID NOT COMPLETE — provider 403 usage limit for the billing cycle, the third leg
lost to that today, so no findings and none invented. Internal compensating leg FIX-FIRST.

## The signed release is NOT blocked, and the tightening is real

This was the highest-risk question and it is answered by execution against the REAL
Developer ID Application: Ferrox Labs, LLC identity on BOTH darwin arches, using the correct
-R '=anchor apple generic and ...' leading-equals form:

    Developer-ID signed arm64   PR: passes    main: passes
    Developer-ID signed x64     PR: passes    main: passes
    ad-hoc signed, both arches  PR: THROWS    main: passes   <- the tightening
    unsigned, both arches       PR: THROWS    main: THROWS

End-to-end through the real staging path, verifyConstitutionFsBundle with requireDarwinSignature
returns true, and appending one byte returns false. wayland-constitution-fs IS inside
mac.signIgnore (electron-builder.yml:322) so electron-builder never re-signs it, and sign-at-stage
gives it a genuine Developer ID identity. --require-darwin-signature is passed iff darwin and
resolveDarwinSigningIdentity(), and _build-reusable.yml sets CSC_NAME, so signed releases take the
strict branch and pass it while build-matrix, which sets no identity, takes the skip branch.
Consistent.

Both root causes are corroborated by CI on this PR's own base commit (build-matrix run 32014210601
at 29b71da20): macos-arm64 SUCCESS on both tracks, macos-x64 FAILURE on both with
"no Developer ID identity ... leaving constitution-fs darwin-x64 unsigned", preview linux FAILURE
with the "Wayland Preview" fuse line then "found 0. Inventory: <empty>", stable linux SUCCESS.

## BLOCKER — a fourth call site was missed, and the PR regresses two GREEN legs

src/index.ts:131-138 — packagedRuntimeReleaseIdentity() still carries the old derivation:

    process.platform === 'linux' ? (releaseTrack === 'preview' ? 'wayland-preview' : 'wayland') : productName

That object is emitted as the boot-start event (src/index.ts:504) and compared EXACTLY,
case-sensitively, as a whole object against the gate's expectedReleaseIdentity at
platform-package-smoke.mjs:1114. It is on the mandatory path of the "Install and smoke real package
payload" step, which the stable Linux legs currently run and pass.

Executed, with a negative control on the base commit:

    base linux stable   emits wayland          expects wayland          PASSED
    base linux preview  emits wayland-preview  expects wayland-preview  PASSED
    PR   linux stable   emits wayland          expects Wayland          THREW
    PR   linux preview  emits wayland-preview  expects Wayland Preview  THREW

So the PR breaks two currently-green required legs AND does not achieve its own goal — preview Linux
still fails, one step later. packagedRuntimeReleaseIdentity is private with zero test coverage,
which is why it was missable. Fix is about six lines plus a test binding the derivation to
expectedReleaseIdentity so the two cannot drift again.

## Two test fixes required in the same pass

**The new gate test's strongest assertion passes for the wrong reason on 2 of 3 shards.**
verifyPackagedResources.test.ts:703 runs the real default, which spawns the absolute path
/usr/bin/codesign, and the it.each block has no platform guard while pr-checks.yml:227 shards the
file across ubuntu, macos and windows. On ubuntu and windows the throw is codesign ENOENT — not a
signature rejection — and a bare toThrow() accepts it. Consequence: NO test asserts the new default
uses -R rather than the old ad-hoc-accepting --verify --strict, so that line could be reverted with
the suite green.

**The exact bug class being fixed is untested at the layer where it happened.** The new test calls
verifyConstitutionFsBundle directly, bypassing the real hand-off through the 23-positional-parameter
isNonEmpty, and the end-to-end harness stubs the signature check out in every case. One reordered
parameter in that 23-arg list — literally what produced this PR — would silently disable constitution
signature enforcement on signed releases, green.

## Verified clean

The 5-param/6-arg claim is true and does not recur: an arity sweep validated on a known positive
reports the mismatch on main and NO mismatches across all 36 declarations on this head; the sibling
verifiers were hand-checked and are all consistent. The Linux name derivation is correct for every
target this repo builds (AppImage, deb, rpm only; electron-builder's sanitiser strips only
/?<>\:*|" and trailing dots and spaces, so an interior space survives — proven by the CI fuse lines
on both tracks). win32 and darwin branches untouched. The full tamper matrix blocks on real signed
bytes, including a self-consistent full forge and a foreign Ferrox-signed binary. No assertion
relaxed, skipped or deleted — three deleted lines total, two fixture literals and one widened type
signature, zero added skip/only/todo, and 6 tests genuinely fail against base. bridgeAllowlist and
the generated constitution file are both 0 lines changed. No other site hardcodes wayland-preview as
an executable name.

## Deferrals, being filed separately

- loadConstitutionFsAuthority() (verify-packaged-resources.js:161) is DEAD CODE — no call sites, and
  the production caller never sets options.constitutionFsAuthority, so the gate compares
  package-authority.json TO ITSELF. Pre-existing; it means the package-sealed-authority claim
  overstates what the gate checks.
- ferrox/release-trust-v1 still pins the old ternary. Dormant only because both protected observers
  are dispatched with release_track=stable hardcoded, where the old value matches case-insensitively.
  Must be refreshed before any preview-track protected observation.
- packaged-cockpit-smoke.mjs:688 silently falls back from out-preview to out, so a Preview cockpit
  smoke can pass while testing the STABLE build.
- The identifier is not bound at :175 where siblings pass darwinSigningIdentifier. Both external legs
  agreed it is not exploitable today, and it is structurally impossible to match because
  prepareConstitutionFs.js:87 computes unsignedSha256 then discards it. Still stronger than main.
- stable/windows-arm64 was RED on the base for an infra reason (Internal CLR error 0x80131506 during
  runner zip extraction), so Build Matrix will not be fully green from this PR alone.

FerroxLabs:
## FIX-FIRST addressed — the blocker was real, and it is fixed

Reproduced the blocker before touching anything, driving the REAL verifyPackageSmokeEventLedger with darwin and win32 as positive controls (my first fixture used the wrong contract string and threw everywhere, which is how I caught that the controls were not yet meaningful):

    before  linux-x64   stable   app "wayland"          gate "Wayland"          THREW packaged runtime release identity mismatch
    before  linux-x64   preview  app "wayland-preview"  gate "Wayland Preview"  THREW packaged runtime release identity mismatch
    before  linux-arm64 stable / preview                                        THREW (both)
    before  darwin-arm64 stable / preview                                       PASS
    before  win32-x64    stable / preview                                       PASS
    after   all four Linux legs, both tracks                                    PASS
    after   darwin and win32, both tracks                                       PASS

So the audit is right on both counts: as it stood this PR regressed the two currently-green stable Linux legs and did not clear preview Linux, it only moved the failure one step later.

### The fix, and why it is shaped this way

The name now has ONE source. getPackagedExecutableName lives in src/common/releaseTrack.ts, src/index.ts calls it, and the private derivation is gone. Simply editing the ternary in place would have left the same two-independent-derivations shape that produced the drift and would still have had zero test coverage, so the derivation moved to the one leaf module that already owns appName, protocolScheme and updateChannel, and is importable from a node-env test.

### Binding test

tests/unit/releaseTrack.test.ts now asserts getPackagedExecutableName equals expectedReleaseIdentity(...).executableName for all 12 track x platform x arch combinations, and that productName agrees too. Reintroducing the old Linux special-case turns exactly the four Linux cases red and nothing else:

    x binds the packaged stable  linux/x64
    x binds the packaged stable  linux/arm64
    x binds the packaged preview linux/x64
    x binds the packaged preview linux/arm64

### Test fix 1 — the wrong-reason pass

Confirmed the mechanism: the old assertion ran the real default, which spawns the absolute path /usr/bin/codesign, and a bare toThrow() accepts the ENOENT the ubuntu and windows shards produce. Two replacements:

- Host-independent: the requirement argv itself is asserted - contains -R, the requirement string starts with =, and includes field.1.2.840.113635.100.6.1.13 plus the pinned team OU.
- Behavioural, and this is the one that actually binds the DEFAULT with nothing injected: Node's own binary is a real thin arm64 Mach-O carrying a valid Developer ID signature from team HX7739G8FX. Executed both ways - codesign --verify --strict returns 0 on it, the Ferrox Labs requirement refuses it. So the fixture discriminates the old default from the new one. Reverting the default to --verify --strict turns that case red, verified.

Worth flagging: my first attempt at this fix ALSO passed for the wrong reason - it injected its own closure, so it asserted the requirement builder rather than the default, and stayed green when I simulated the revert. That is why the behavioural case exists.

### Test fix 2 — the hand-off is now covered

One end-to-end case drives the full sweep (verifyPackagedResources to the 23-positional-parameter isNonEmpty to verifyConstitutionFsBundle) with a recorder, asserting the check fires with the flag and does not without it, and that a refusal fails the whole sweep rather than being swallowed. Transposing the last two hand-off arguments - literally the bug class that produced this PR - turns it red, verified.

### State

Full unit suite: 18050 passed, 155 skipped, 3 failed. The 3 are the pre-existing acpAgentManagerNpmFallback host failure (a real wayland-nano at ~/.local/bin, confirmed present). oxfmt --check clean on the touched files only. tsc --noEmit clean. Rebased onto 9a454acdb. bridgeAllowlist.ts, urlValidation.ts, migrations.ts and constitutionFsAuthority.generated.ts are all absent from the diff against main.

Also corrected an overstatement I had written in a code comment: package-authority.json ships inside the package, so it is the runtime re-hash against the digest in app.asar that authenticates the helper bytes, not that sealed file. See the deferred note below.

### Deferred, not fixed here

- loadConstitutionFsAuthority (verify-packaged-resources.js:161) is dead code - zero call sites - and :1218 reads options.constitutionFsAuthority which no production caller sets, so authority is undefined and the gate compares package-authority.json to itself. Pre-existing.
- The protected branch release-trust-v1 still carries the old ternary in its pinned copies. Dormant while both protected observers are dispatched with release_track hardcoded to stable, but it must be refreshed before any preview-track protected observation. Not touched.
- scripts/packaged-cockpit-smoke.mjs falls back from out-preview to out, so a Preview cockpit smoke can pass while exercising the STABLE build.
- The signing identifier is not bound at the constitution call site the way siblings bind it; structurally unmatchable today because prepareConstitutionFs computes unsignedSha256 and discards it.
- stable/windows-arm64 was red on the base commit for an infra reason (Internal CLR error 0x80131506 during runner zip extraction), so Build Matrix will not go fully green from this PR alone - that leg needs a re-run.

Issue #60, opened 2026-06-11
[Bug]: Claude Code, keeps losing its connection, and stops working every 3-5 mins


mccarthy-os:
### Platform

Windows

### Wayland Version

v0.9.6-rc.2.1

### Bug Description

Claude code stops being active... Claude code is doing nothing... but messages go to queue...

After 3-5 mins error messages start to appers in the message area.

Cannot send in promopting state
process exited unexpectedly (code: unknown, signal: none)
Agent disconnected (connectoion_close, code: null)
Cannot read properties of null (reading 'createSession')

If you press the stop button, because its active, it then gives the same prompting state message.

Killing Wayland and restarting allows you to continue again, but its not ideal.

Sean Donahoe said this "That looks like a JSON streaming problem with Claude Code but do me a favour: drop it in GitHub Issues for Wayland and I'll have the team take a look"

### Steps to Reproduce

As detailed above... it keeps happening, restart, then it hapens again after 3-5 mins of use with Claude Code.

### Expected Behavior

Continous use without errors.

### Actual Behavior

Crashes, and has to restart the software to work again.

FerroxLabs:
Tracked it down. The idle/keepalive clock was not staying alive during an active stream, so a long Claude Code turn looked idle and got torn down mid-response, which is exactly the every-few-minutes drop you were seeing. Patched so the clock keeps ticking while a response is streaming. Fix is in 0.9.7, rolling out now. Grab the update from the app and let me know if you still see drops. Appreciate the detailed report, that is the kind that is easy to act on.

Issue #676, opened 2026-07-04
[Bug] Claude Code backend dies on corrupt bunx adapter install (zod/v4 unresolvable) + respawn-storms the UI


FerroxLabs:
## Repro (hit live on a packaged mac build, 2026-07-04)

Open an existing project, open an existing Claude Code chat inside it. On open, a rapid burst of identical error banners appears back-to-back:
- "Claude Code - Connection error"
- "process exited unexpectedly (code: unknown, signal: none)"
- "process exited unexpectedly (code: 1, signal: none)" x3
- "Agent exited before initialize completed (code: 1) / error: Cannot find module 'zod/v4' from '.../bunx-501-@agentclientprotocol/claude-agent-acp@0.55.0/node_modules/@agentclientprotocol/sdk/dist/schema/zod.gen.js' / Bun v1.3.14 (macOS arm64)"

## Root cause (verified on-machine)

The Claude Code ACP adapter (@agentclientprotocol/claude-agent-acp@0.55.0) is spawned via bunx into an OS temp dir with the app's bundled Bun (1.3.14). In the failing install the zod dependency was left HALF-INSTALLED:
- node_modules/zod/package.json was MISSING (so Bun has no exports map and cannot resolve ANY zod subpath)
- node_modules/zod/v4/index.js was MISSING (the v4/ dir had classic/core/locales/mini subfolders but no entry point)

The SDK code is correct (import * as z from "zod/v4"; peerDep zod "^3.25.0 || ^4.0.0" — all valid). The global bun cache was healthy (zod@3.25.76 / 4.3.6 / 4.4.3 all intact with ./v4). The corruption was isolated to the ephemeral bunx temp install — a partial/interrupted install. Clearing that temp dir fixes it (bunx reinstalls clean from the good cache).

## Two bugs to fix (this incident = one root + one symptom)

BUG A — RESPAWN STORM / no backoff (P1, the ugly part). On the code:1 init failure the desktop ACP connection manager respawned the adapter repeatedly in fast succession, flooding the chat with identical banners and dumping a raw Bun stack trace to the user. It must: cap retries, apply exponential backoff, and surface ONE clear, actionable error ("Claude Code failed to start — adapter install looks corrupt, reinstalling…" or a Retry button), never a storm of raw process-exit lines. This is the same class as the #575 spawn-loop guard, applied to the ACP-adapter init path. Ties to #656 and #629 (no recovery UI on backend init failure).

[...]

PR #1023, opened 2026-08-17
fix(acp): stop reporting a transport drop as a process exit

### LINKED ISSUE #1020: Transport disconnect is reported as "process exited unexpectedly (code: unknown, signal: none)", discarding reason and stderr
Reported by a customer running the Claude Code backend with Sonnet. The chat showed:

    process exited unexpectedly (code: unknown, signal: none)

The user asked what it meant, and the assistant -- given nothing to work with -- answered that
"the underlying Claude Code process died on its own" and listed four plausible causes. That
answer was a confabulation induced by the message. The user then lost an approval turn: their
"approved: verification complete" was never acted on, no approval file was written.

## The message asserts something it has no evidence for

Only one producer renders the literals "unknown" and "none": buildCrashMessage in
src/process/acp/session/AcpSession.ts:80. The other three producers spell it differently
(capital P, or "null"), so this is definitively that line. It emits those literals only when
BOTH info.exitCode and info.signal are null.

Of the four AgentDisconnectReason values (IAcpClient.ts:77), the two that read the code and
signal off a possibly-live child are:

    ProcessAcpClient.ts:391  pipe_close        child.exitCode ?? null, child.signalCode ?? null
    ProcessAcpClient.ts:136  connection_close  child.exitCode ?? null, child.signalCode ?? null

whereas process_exit (:385) and process_close (:388) take them from Node's exit/close events,
where one is always populated.

child.exitCode and child.signalCode are only set after the process exits. Both being null
means the stdio pipe or the ACP connection dropped while the child was, as far as Node could
see, still running. So the banner reports a process exit that did not demonstrably happen.

Caveat, stated honestly: there is a race in which the process is genuinely dying and the
pipe-close handler fires before the exit event, leaving both fields unset. That is not
distinguishable from the current message either -- which is the point.

## The two fields that would identify the cause are discarded

DisconnectInfo (IAcpClient.ts:94) carries reason and stderr. ProcessAcpClient populates both,
including the full stderr buffer at :420 and :437. buildCrashMessage ignores both and renders
only the two fields that are null. The real cause is therefore unknowable by construction.

This is the same failure shape as #1008: a diagnostic that sends the user hunting somewhere
the fault cannot be.

## Suspected trigger: the prompting path does not get the suppression the active path does

AcpSession.onDisconnect (AcpSession.ts:414) treats the states differently:

- state "active" (no prompt in flight) deliberately suppresses the banner. Its comment says the
  agent bridge may shut down after an inactivity timeout, and that emitting a crash signal
  would show the user a scary "process exited unexpectedly" even though the conversation
  completed normally.
- state "prompting" always calls emitCrashSignalIfProcessDied.

So a benign inactivity-shaped transport close that happens to land while a prompt is in flight
produces exactly this banner. The customer was mid-approval, which fits.

ProcessAcpClient already computes the flag that distinguishes these: unexpectedDuringPrompt at
:419, defined as not closing and hasActivePrompt. It is recorded on _lastExit and not consulted
by the message path.

## What to fix

1. Never claim a process exit without evidence of one. Distinguish a transport close from a
   process death in the user-visible text.
2. Include the reason, and a redacted tail of stderr, so the actual cause is recoverable.
   Redaction is mandatory -- route it through the existing secret scrubber, not a new one.
3. Decide whether a transport close during a prompt should be recovered silently (the session
   already resumes via resumeFromDisconnect) rather than surfaced as a crash.
4. Establish the underlying cause of the pipe or connection close on the Claude Code backend.

## Acceptance

A transport-level disconnect on the Claude Code backend produces a message that names the
reason and carries a scrubbed stderr tail, and does not assert a process exit unless an exit
code or signal was actually observed. Proven by driving each of the four disconnect reasons.

FerroxLabs:
Fixes #1020.

A customer on the Claude Code backend was shown

    process exited unexpectedly (code: unknown, signal: none)

asked what it meant, and got a confident, wrong answer about the process having died. They also
lost an approval turn: their "approved: verification complete" was never acted on and no
approval file was written.

## The message asserted something it had no evidence for

buildCrashMessage (AcpSession.ts:80) rendered those literals only when BOTH exitCode and signal
were null. Of the disconnect reasons, pipe_close (ProcessAcpClient.ts:425) and connection_close
(:170) read the code and signal off a child that may still be running, so both come back null.
process_exit (:419) and process_close (:422) take them from Node's exit and close events, where
one is always populated, and the fifth site at :511 only fires when at least one is already set.

So the banner claimed a process exit precisely in the cases where no exit had been observed.

## The two fields that identify the cause were discarded

DisconnectInfo carries reason and the full stderr buffer. Both were populated and both were
thrown away by the message, so the real cause was unknowable by construction. That is what let
the assistant confabulate.

## What changed

- Claim an exit only when a code or a signal is actually present. Otherwise say plainly that
  this is a transport-level disconnect, that no exit was reported, and that the agent process
  may still be running.
- Name the reason in both cases.
- Carry a bounded stderr tail, ANSI-stripped and routed through the shared secret scrubber in
  src/process/utils/secretRedaction.ts. Untrusted subprocess output is on its way to the chat
  transcript and the log file, and an agent bridge prints to stderr exactly when credentials are
  in play, so scrubbing is not optional. No new pattern list was introduced.
- Disclose the lost turn. The prompting path respawns via resumeFromDisconnect and re-flushes
  the queue, but the turn that was in flight is gone: flush() shifts it off the queue before
  execute(), and handlePromptError deliberately refuses to hand it back on a dead stream because
  a tool_call it already ran can be lost with the pipe. Silently recovering would leave the user
  believing their message landed, which is the half of this bug that actually cost the customer
  work.

Diagnostics here are deliberately not routed through i18n, matching the surrounding
AgentDisconnectedError, AgentStartupError and enterError strings. No locale keys change.

## Split after the NO-GO cross-audit: the EXIT_CONFIRM_MS deferral is OUT

The four-leg audit returned NO-GO on an `EXIT_CONFIRM_MS` deferral that had been bolted onto this
branch. It waited up to 200ms for the child's exit event before reporting the disconnect, to
recover an exit code the transport abort had raced. It has been removed entirely, and it is
**deliberately deferred to separate follow-up work** rather than reworked here.

Why it had to come out: `recordAgentExit` rejects the pending SDK requests and then notifies the
disconnect handler. That looks like the wrong order and is not - a rejection's continuation is a
microtask, so `AcpSession.onDisconnect` runs BEFORE `PromptExecutor` sees the rejected prompt,
which is what arms `handlePromptError`'s "someone else owns recovery" guard. Deferring the
notification inverted it: `handlePromptError` ran while the status was still `prompting`, took its
retryable branch, emitted its own raw `Agent disconnected (connection_close, code: null)` banner
and flushed the queued follow-up at the dead client. Net effect, measured: every banner this PR
adds was unreachable on the live-child transport drop that motivated it, and a queued message was
silently lost - #774, reintroduced by the fix for #1020. The floating confirm promise also
outlived `close()`, so our own `gracefulShutdown` SIGTERM could render as an unexpected exit.

Notification is synchronous again, byte-for-byte main's shape plus the new
`unexpectedDuringPrompt` field. Removing the deferral removed the second and third audit findings
as consequences, verified by execution rather than assumed.

### The tradeoff, accepted deliberately

Without a wait, a genuine fast crash whose `exit` event is a millisecond away can be reported as a
transport close rather than a confirmed exit. Measured here: an `exit(7)` child reports
`process_exit / 7` on some runs and `connection_close / null / null` on others. That is the correct
direction. A hedged "no exit was reported, the process may still be running" is honest about what
was observed; asserting "process exited" with no code and no signal is the #1020 bug itself. The
reasoning is in a comment at the call site so nobody re-adds a deferral to chase precision at the
cost of the ordering.

Any rework of `EXIT_CONFIRM_MS` must keep the disconnect notification synchronous - upgrade the
detail some other way, or not at all.

## Verification

All executed on this head, not reasoned about.

**Ordering, restored and measured** on the customer's shape (a fixture that answers `initialize`
and `session/new`, then ends stdout on `session/prompt` and stays alive):

    this head:      disconnect t=48ms  ->  prompt-rejected t=48ms   (gap 0ms, handler FIRST)
    with deferral:  prompt-rejected t=34ms, disconnect never inside the window (gap >200ms)

**The test gap that let the defect through is closed.** Every existing test asserted
`DisconnectInfo`, which is one layer below where the bug lived: the client can hand the session a
perfect payload and the user can still read a completely different string, because
`handlePromptError` also emits a banner and whichever runs first wins the screen. So there is now
an end-to-end test that drives a real child through a real `ProcessAcpClient` into a real
`AcpSession` and asserts the banner **the session emits**. Fail-before, both ways:

    with the deferral:  first banner = "Agent disconnected (connection_close, code: null)"
    against main:       first banner = "process exited unexpectedly (code: unknown, signal: none)"

**Other test changes**

- The `pipe_close` / `process_close` case previously constructed `ProcessAcpClient` without
  calling `start()`, so `this.child` was never assigned and the live-child state the disconnect
  path reads was never exercised. It now drives a really-started client against the fixture, and
  asserts the report lands with no `await` at all.
- The two race tests no longer assert that an exit code is always recovered - that was only true
  because of the deferral. They assert the report is never *wrong* instead: either the real
  code/signal, or a hedged transport drop that claims no exit. Each carries a known-positive
  control read off the child's own `exit` event, so a null can never be a silently broken fixture.
- The three #774 crash-recovery tests are parameterised over both orderings a real client can
  produce. Pinning one ordering is exactly why they stayed green while the deferral reintroduced
  the bug they exist to prevent. Verified load-bearing: all three fail under a deliberately
  deferred ordering.
- Fixture: added the drop-on-prompt mode, removed the never-driven `garbage` mode. All four modes
  are now driven by a test.
- `AcpSession.lifecycle.test.ts` now passes the required `unexpectedDuringPrompt`. Note this was
  never a compile break - `tsconfig.json` includes only `src/**/*` - so it is tidying, not a fix.

`tsc --noEmit` clean. `oxfmt` run on exactly the seven touched files. `oxlint`: 0 errors.
`bridgeAllowlist.ts` is 0 lines changed against `ferrox/main`; no generated or locale files touched.

## What is unchanged from the audited-clean half

The `buildCrashMessage` rewrite, `crashMarkers.ts`, the `DisconnectInfo` additions and both
string-matching consumers (`AcpAgentV2`, `TeammateManager`) are as audited: genuinely fail-before,
nothing weakened, scrubbing sound, the 2048 bound applied to the scrubbed text, and `AcpAgentV2`'s
burst dedup unaffected because it keys on `errorBannerId` rather than message text.

## Tracked separately

- `EXIT_CONFIRM_MS`, per above.
- A scrubber gap where a labelled secret cannot match after an underscore, so `ANTHROPIC_API_KEY`
  survives. Pre-existing, same class as #1004.
- `SessionLifecycle.clearClient` orphaning a live agent process on a transport drop. Pre-existing,
  now proven - the tests here reap the child by hand.

### DIFF
--- a/src/process/acp/compat/AcpAgentV2.ts
+++ b/src/process/acp/compat/AcpAgentV2.ts
@@ -24,6 +24,7 @@ import {
 } from '@process/acp/compat/typeBridge';
 import { AcpError as AcpSessionError } from '@process/acp/errors/AcpError';
 import { AcpSession, type SessionOptions } from '@process/acp/session/AcpSession';
+import { CRASH_MARKER_PROCESS_EXIT, CRASH_MARKER_TRANSPORT_CLOSE } from '@process/acp/session/crashMarkers';
 import { readClaudeModelInfoFromCcSwitch } from '@process/services/ccSwitchModelSource';
 import { buildClaudeSlotModelInfo } from '@process/agent/acp/utils';
 // TODO(ACP Discovery): Re-enable when acp_session persistence is restored.
@@ -622,8 +623,12 @@ export class AcpAgentV2 {
           case 'error': {
             // Detect process crash from error message keywords to emit agentCrash
             // flag that TeammateManager.handleResponseStream relies on (V1 parity).
+            // CRASH_MARKER_TRANSPORT_CLOSE is the #1020 wording for a disconnect
+            // with no observed process exit. Without it here a transport drop
+            // stopped matching, and the renderer's loading state never cleared.
             const isCrash =
-              event.message.includes('process exited unexpectedly') ||
+              event.message.includes(CRASH_MARKER_PROCESS_EXIT) ||
+              event.message.includes(CRASH_MARKER_TRANSPORT_CLOSE) ||
               event.message.includes('PROCESS_CRASHED') ||
               event.message.includes('Process disconnected');
 
--- a/src/process/acp/infra/IAcpClient.ts
+++ b/src/process/acp/infra/IAcpClient.ts
@@ -93,9 +93,24 @@ export type AgentLifecycleSnapshot = {
 
 export type DisconnectInfo = {
   reason: AgentDisconnectReason;
+  /**
+   * Null for BOTH of these only when no process exit was ever observed - i.e. a
+   * transport-level drop while the child was, as far as Node could see, still
+   * running. The disconnect is not reported until the exit event has had a
+   * bounded window to arrive, so null/null is now evidence of absence rather
+   * than a lost race (#1020).
+   */
   exitCode: number | null;
   signal: string | null;
   stderr: string;
+  /**
+   * True if a prompt request was in flight when the transport dropped. That turn
+   * is NOT replayed - `PromptExecutor.handlePromptError` deliberately refuses to
+   * re-queue an in-flight prompt on a dead stream, because a `tool_call` it had
+   * already run can be lost with the pipe - so the user has to be told it did
+   * not land (#1020).
+   */
+  unexpectedDuringPrompt: boolean;
 };
 
 // ─── Client Factory ─────────────────────────────────────────────
--- a/src/process/acp/infra/IAcpClient.ts
+++ b/src/process/acp/infra/IAcpClient.ts
@@ -101,8 +101,10 @@ export type DisconnectInfo = {
    * The report is deliberately NOT delayed to wait for a late 'exit' event (see
    * `ProcessAcpClient.recordAgentExit` - delaying it inverts the ordering the
    * session depends on), so a genuine fast crash can land here too. That is why
-   * `buildCrashMessage` hedges ("the process may still be running") instead of
-   * asserting an exit it has no evidence for (#1020).
+   * `buildCrashMessage` reports only what was observed ("no exit code or signal was
+   * reported") instead of asserting an exit it has no evidence for (#1020), and does
+   * NOT reassure the user that the child is probably alive - measured against each
+   * child's own 'exit' event, that reassurance was wrong for 6 of 20 real crashes.
    */
   exitCode: number | null;
   signal: string | null;
--- a/src/process/acp/infra/IAcpClient.ts
+++ b/src/process/acp/infra/IAcpClient.ts
@@ -94,11 +94,15 @@ export type AgentLifecycleSnapshot = {
 export type DisconnectInfo = {
   reason: AgentDisconnectReason;
   /**
-   * Null for BOTH of these only when no process exit was ever observed - i.e. a
-   * transport-level drop while the child was, as far as Node could see, still
-   * running. The disconnect is not reported until the exit event has had a
-   * bounded window to arrive, so null/null is now evidence of absence rather
-   * than a lost race (#1020).
+   * Null for BOTH of these means no process exit had been observed AT THE MOMENT
+   * the disconnect was reported - typically a transport-level drop while the
+   * child was, as far as Node could see, still running.
+   *
+   * The report is deliberately NOT delayed to wait for a late 'exit' event (see
+   * `ProcessAcpClient.recordAgentExit` - delaying it inverts the ordering the
+   * session depends on), so a genuine fast crash can land here too. That is why
+   * `buildCrashMessage` hedges ("the process may still be running") instead of
+   * asserting an exit it has no evidence for (#1020).
    */
   exitCode: number | null;
   signal: string | null;
--- a/src/process/acp/infra/ProcessAcpClient.ts
+++ b/src/process/acp/infra/ProcessAcpClient.ts
@@ -476,9 +476,10 @@ export class ProcessAcpClient implements AcpClient {
     // The tradeoff this buys is accepted deliberately: because nothing waits for
     // the child's 'exit' event, a genuine fast crash whose exit is still a
     // millisecond away is reported as a transport close rather than a confirmed
-    // exit. That is the correct direction. A hedged "no exit was reported, the
-    // process may still be running" is honest about what was observed; asserting
-    // "process exited" with no code and no signal is the #1020 bug itself.
+    // exit. That is the correct direction. Reporting "no exit code or signal was
+    // reported" is honest about what was observed; asserting "process exited" with
+    // no code and no signal is the #1020 bug itself. The banner stops there and does
+    // not claim the child is probably alive, because for a fast crash it is not.
     const error = new AgentDisconnectedError(reason, exitCode, signal ? String(signal) : null, {
       outputAlreadyEmitted: this.hasActivePrompt,
     });
--- a/src/process/acp/infra/ProcessAcpClient.ts
+++ b/src/process/acp/infra/ProcessAcpClient.ts
@@ -55,25 +55,6 @@ const STARTUP_STDERR_MAX = 8192;
 const MISSED_EXIT_REPLAY_MS = 250;
 
 /**
- * How long a transport-level disconnect that carries NO exit code and NO signal
- * waits for the child's own exit event before it is reported as a transport drop.
- *
- * The four lifecycle signals are first-write-wins, and the SDK's connection abort
- * structurally beats Node's exit event: the abort fires on the readable's 'end',
- * which precedes both `stdout`'s 'close' and 'exit'. Measured on macOS/node 22, a
- * child that called `process.exit(7)` reported `connection_close / exitCode: null
- * / signal: null` on ~3 runs in 5, and a SIGKILL lost its signal every time -
- * while `child.exitCode`/`child.signalCode` were populated 0-1ms later. So the
- * banner asserted "code: unknown, signal: none" for ordinary, fully-diagnosable
- * crashes (#1020).
- *
- * `normalizeInitializeError` already resolves this exact race with the same 200ms
- * budget; that fix only ever covered the initialize path. This extends it to the
- * post-init disconnect path.
- */
-const EXIT_CONFIRM_MS = 200;
-
-/**
  * Resolve `target` to a real filesystem path, symlinks included.
  *
  * bun names the REAL path of the module it could not resolve. On macOS that is
@@ -467,62 +448,51 @@ export class ProcessAcpClient implements AcpClient {
       console.warn(`[ACP ${this.options.backend}] Process exited with code ${exitCode} [reason: ${reason}]`);
     }
 
+    const unexpectedDuringPrompt = !this.closing && this.hasActivePrompt;
+
     this._lastExit = {
       exitCode,
       signal: signal ? String(signal) : null,
       reason,
       stderr: this.stderrBuffer,
-      unexpectedDuringPrompt: !this.closing && this.hasActivePrompt,
+      unexpectedDuringPrompt,
     };
 
-    // Reject all pending SDK requests with our own error type. This stays
-    // synchronous: a request left hanging is worse than a slightly less precise
-    // exit summary, and the in-flight prompt needs to fail now.
+    // Reject the pending SDK requests, then notify the disconnect handler - both
+    // SYNCHRONOUSLY, in that order.
+    //
+    // That order is load-bearing, and it is not what it looks like. A rejection's
+    // continuation is a microtask, so `AcpSession.onDisconnect` still runs BEFORE
+    // `PromptExecutor.handlePromptError` sees the rejected prompt. That is exactly
+    // what arms `handlePromptError`'s "someone else owns recovery" guard
+    // (`status !== 'prompting'`): `onDisconnect` has already moved the session on.
+    //
+    // Put ANY await, timer or deferral between these two statements and the
+    // ordering inverts: `handlePromptError` runs first, takes its retryable
+    // branch, emits its own raw banner, flushes the queued follow-up at the dead
+    // client (#774), and the session-level #1020 banner becomes dead code. Do not
+    // add one.
+    //
+    // The tradeoff that buys is accepted deliberately: because nothing waits for
+    // the child's 'exit' event, a genuine fast crash whose exit is still a
+    // millisecond away is reported as a transport close rather than a confirmed
+    // exit. That is the correct direction. A hedged "no exit was reported, the
+    // process may still be running" is honest about what was observed; asserting
+    // "process exited" with no code and no signal is the #1020 bug itself.
     const error = new AgentDisconnectedError(reason, exitCode, signal ? String(signal) : null, {
       outputAlreadyEmitted: this.hasActivePrompt,
     });
     this.rejectPendingRequests(error);
 
-    // No exit detail and a child still attached: the exit event may simply not
-    // have landed yet. Give it a bounded window before reporting "no process exit
-    // observed", so the banner never asserts the absence of something that was
-    // one millisecond away (#1020). Everything else reports immediately, exactly
-    // as before.
-    if (exitCode === null && !signal && this.child) {
-      void this.confirmExitThenNotify(reason);
-      return;
+    if (this.disconnectHandler) {
+      this.disconnectHandler({
+        reason,
+        exitCode,
+        signal: signal ? String(signal) : null,
+        stderr: this.stderrBuffer,
+        unexpectedDuringPrompt,
+      });
     }
-    this.notifyDisconnect(reason);
-  }
-
-  /**
-   * Wait out {@link EXIT_CONFIRM_MS} for the child's exit, then report whatever is
-   * actually known. `_lastExit` is upgraded in place: `reason` keeps first-write-wins
-   * (the transport close IS what we saw first), but the exit detail is filled in
-   * rather than pinned at null, so `lifecycleSnapshot` agrees with the banner.
-   */
-  private async confirmExitThenNotify(reason: AgentDisconnectReason): Promise<void> {
-    const child = this.child;
-    if (child) {
-      await waitForExit(child, EXIT_CONFIRM_MS);
-      const exitCode = child.exitCode ?? null;
-      const signal = child.signalCode ? String(child.signalCode) : null;
-      if (this._lastExit && (exitCode !== null || signal !== null)) {
-        this._lastExit = { ...this._lastExit, exitCode, signal };
-      }
-    }
-    this.notifyDisconnect(reason);
-  }
-
-  private notifyDisconnect(reason: AgentDisconnectReason): void {
-    if (!this.disconnectHandler) return;
-    this.disconnectHandler({
-      reason,
-      exitCode: this._lastExit?.exitCode ?? null,
-      signal: this._lastExit?.signal ?? null,
-      stderr: this.stderrBuffer,
-      unexpectedDuringPrompt: this._lastExit?.unexpectedDuringPrompt ?? false,
-    });
   }
 
   // ─── Internals: Startup failure watcher ────────────────────
--- a/src/process/acp/infra/ProcessAcpClient.ts
+++ b/src/process/acp/infra/ProcessAcpClient.ts
@@ -41,11 +41,31 @@ import type {
 } from '@process/acp/infra/IAcpClient';
 import { NdjsonTransport } from '@process/acp/infra/NdjsonTransport';
 import { gracefulShutdown, waitForExit, waitForSpawn } from '@process/acp/infra/processUtils';
+import { redactSecrets } from '@process/utils/secretRedaction';
 import type { PromptContent, ProtocolHandlers } from '@process/acp/types';
 import type { ChildProcess } from 'node:child_process';
 import * as fs from 'node:fs';
 
 const STARTUP_STDERR_MAX = 8192;
+
+/**
+ * How much of the stderr ring's trailing non-whitespace run is held back from the
+ * collection-site scrub.
+ *
+ * A credential can straddle two stderr chunks. Scrubbing the buffer the moment it
+ * overflows would mask the half that has arrived, replacing its ANCHOR with
+ * `[redacted]` and orphaning the half that arrives next - which no later scrub can
+ * match, because the thing that identified it is gone. Measured through the real
+ * path: 51 characters of a live-shaped Anthropic key reached the banner that way.
+ *
+ * So the trailing run is left raw for the next chunk to complete. That is safe
+ * because it sits at the ring's TAIL, which the cap keeps rather than cuts, and
+ * because a partial with its PREFIX still intact is redactable
+ * (`redactSecrets('sk-ant-api03-AbCdEfGhIjK')` masks) - it is only the anchorless
+ * REMAINDER that is invisible to the scrubber. A run longer than this is a blob, not
+ * a straddled credential, so it is scrubbed normally.
+ */
+const STDERR_CHUNK_SPLIT_CARRY_MAX = 256;
 
 /**
  * How long to wait for a child that had ALREADY exited when the lifecycle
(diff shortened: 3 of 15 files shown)


FerroxLabs:
## Cross-audit verdict: NO-GO

Four legs on b30777812 (head has since moved to bdb99ddf6, an oxfmt-only reformat of one test
line, so every finding stands). Codex 5.6 Sol: NO-GO. Gemini 3.1 Pro: FIX-FIRST. Kimi K3 ran and
then DIED mid-audit on a provider usage limit, so it emitted no verdict and none is being invented
for it, though its partial reasoning independently reached Finding 1. Internal: NO-GO.

All three legs that got anywhere converged on the same headline defect. Gemini found it from the
source alone, without being able to read the diff.

## F1 (blocking) — the EXIT_CONFIRM_MS deferral disables this entire fix on the path that motivated it

ProcessAcpClient.ts:477-495 keeps rejectPendingRequests synchronous but defers the disconnect
notification via confirmExitThenNotify. Executed against the real client with a fixture that ends
stdout on prompt but stays alive, which is the customer shape:

    branch:  prompt-rejected t=204ms  ->  disconnect t=404ms   (gap 200ms)
    main:    disconnect     t=132ms  ->  prompt-rejected t=132ms  (handler FIRST)

waitForExit on a live child waits the full timer, so this is structural, not load-sensitive.

At session level the consequence is that the rejection microtask reaches
PromptExecutor.handlePromptError while status is still prompting. AgentDisconnectedError is
PROCESS_CRASHED and retryable, so PromptExecutor.ts:350-353 sets status active, emits its own raw
banner, and flushes. 200ms later AcpSession.onDisconnect sees active and silently suspends.

Executed, both orderings:

    ORDER-OLD (main):  banner = process exited unexpectedly (code: unknown, signal: none)
    ORDER-NEW (this):  only signal = Agent disconnected (connection_close, code: null)

So the new banner, the reason, the "may still be running" clause, the scrubbed stderr tail and the
lost-turn disclosure are ALL dead code on the customer's path. The paths that defer are exactly
exitCode === null and no signal and a live child - the #1020 transport drop plus the exit races the
deferral's own docblock cites as its motivation. The fix is disabled precisely where it was aimed.

## F2 (blocking) — F1 reintroduces #774: the queued follow-up is fired at a dead client and lost

Same harness as the existing #774 test:

    ORDER-OLD: events = [start:1, loadSession:1, prompt:1]  - respawn owns recovery, follow-up delivered
    ORDER-NEW: events = [prompt:0]  clients=1  status=suspended  - fired at the DEAD client, message LOST

handlePromptError's retryable branch flushes while lifecycle.client is still the dead client, and
resumeFromDisconnect never runs. The existing #774 regression test does not catch this because it
hardcodes the old ordering, so its guarantee is now false in production.

A PR whose headline is that a customer silently lost an approval turn causes an additional queued
message to be silently lost.

## F3 (fix first) — the floating confirm promise outlives close() and misattributes our own kill

ProcessAcpClient.ts:492 floats the promise; close() sets closing, gracefully shuts the child down
and nulls this.child, but cancels neither the pending confirm nor the handler. Executed:

    AFTER_CLOSE_INFO {reason: connection_close, exitCode: null, signal: SIGTERM}  firedAfterClose=true

That SIGTERM is our own gracefulShutdown kill, and it renders as "process exited unexpectedly
(code: unknown, signal: SIGTERM)" - a brand new confabulation of exactly the class this PR exists
to remove. Fix: after the await, bail if closing or if this.child changed, and never upgrade
lastExit from a kill we issued.

## F4 (fix first) — the deferral branch is effectively untested

acpDisconnectTransport.test.ts:119-135 constructs the client without calling start(), so this.child
is never assigned, the guard is false and confirmExitThenNotify is never entered - while production
DOES defer for a live-child pipe_close or process_close. There is also no test that EXIT_CONFIRM_MS
actually upgrades an exit code. The PR body's claim that each reason is exercised for real is
unsupported for two of four.

## What is sound, verified by execution

The buildCrashMessage rewrite itself is good work. Tests are genuinely fail-before: 15 of 23 fail
on main. Nothing was weakened - the four rewritten assertions all pinned the OLD WRONG message and
coverage went 5 to 23. All five recordAgentExit sites verified, including the guarded fifth. The
active branch of onDisconnect is byte-unchanged and still silent. AcpAgentV2 burst-dedup survives
because it keys on errorBannerId, not message text. Both string-matching consumers were updated and
a repo-wide grep found none left behind. bridgeAllowlist is 0 lines changed; no locale or generated
files; tsc, oxlint and oxfmt clean on the current head.

Secret scrubbing is sound: all four of #1004's trailing-underscore escapes are masked, an
ANSI-CSI-split token is masked so strip-then-scrub is the correct order, the 2048 cap is applied to
the SCRUBBED text so a slice can only cut a redaction marker rather than real bytes, and the 8KB
ring buffer bound is real.

## #60 recurrence

#1020 is the same signature as the closed #60, which pastes verbatim the same null/null banner, the
same "Agent disconnected (connection_close, code: null)", the same backend. #60 was closed without
ever capturing the reason or stderr that would have named the cause. It is NOT #676, which was a
corrupt bunx adapter install storming the UI at start time.

This PR is the right instrument to settle #60 - reason plus liveness plus stderr tail on a real
recurrence - but per F1 it renders none of that on the mid-prompt path, so as written it cannot.

## Disposition

Splitting it, per the audit's recommendation. The buildCrashMessage, crashMarkers and consumer half
lands once F4's test gap is closed. EXIT_CONFIRM_MS comes out and is reworked separately behind an
end-to-end mid-prompt disconnect test that asserts the banner the SESSION emits - the current tests
only ever assert DisconnectInfo, which is how this got through.

Deferrals being tracked separately: a scrubber gap where a labelled secret cannot match after an
underscore so ANTHROPIC_API_KEY survives (pre-existing, same class as #1004), and
SessionLifecycle.clearClient orphaning a live agent process on a transport drop (pre-existing, now
proven - the PR's own test SIGKILLs it by hand).

FerroxLabs:

Legs: Codex 5.6 Sol NO-GO, completed on the correct PR, strongest leg, and its lead finding is the best of this audit. Gemini 3.1 Pro DID NOT COMPLETE, three shell errors, executed nothing, and one of its findings cited a string that exists nowhere in the repo, so it was discarded. Kimi K3 DID NOT RUN, provider 403 usage limit. Internal execution leg FIX-FIRST.

This is the best-tested diff in this swarm. Real child processes, real clients, real sessions, and 10 of 13 mutants died. The core mechanism is correct and stays. Four things to fix.

## Verified correct, do not change

Branching on evidence only at AcpSession.ts:123 with the reason appended as text is right. Codex proposed also treating process_exit and process_close as confirmed exits; measured to be a no-op for real traffic because those reasons never arrive with both fields null. Refuted.

Twenty real children driven through the client: process_exit and connection_close are the only two reasons that ever win the first-write race. A transport drop on a genuinely live child reports connection_close with null code and null signal and a hedged banner, and the process is still alive afterwards. Correct.

The alreadyExited guard holds under a forced real interleaving, 5 of 5 runs, rejecting in 27 to 74ms and never hanging. The 310-line edit to the existing prompt test removed 15 assertions and re-added 15, net zero lost, a strengthening.

## Findings

**1. HIGH. The banner tells the user to do the thing the repo has code to refuse.** AcpSession.ts:131 says the in-flight message was not resent automatically, send it again. PromptExecutor.ts:316-327 is eleven lines that exist to refuse an automatic resend, and says why: on a dead stream we cannot know what the turn already did, the pipe can swallow the very tool_call that would have told us, and the turn may have already run rm. PromptExecutor.ts:31 adds that the pipe can make turnRanTool lie, so the advice cannot even be conditioned on it. The #1020 customer lost an approval mid-turn. Wording fix, no logic change.

**2. MEDIUM-HIGH. The two lines that prevent a permanently stuck spinner have zero test coverage.** AcpAgentV2.ts:631 and TeammateManager.ts:670, both mutants survived all 204 tests. Confirmed by execution: the new banner matches none of the other three crash substrings, so without line 631 isCrash is false, no finish frame is synthesised, and per the PR's own crashMarkers.ts documentation the renderer's loading state never clears. The PR reworded the banner, which breaks both matchers unless updated. They were updated. Nothing locks it in. Verified repo-wide, with a known-positive control, that there is no third matcher.

**3. MEDIUM. "may still be running" is wrong for 30 percent of real crashes.** Ground truth taken from each child's own exit event: 3 of 10 real exit(7) runs and 3 of 10 real SIGKILL runs are described as probably still alive. The obvious fix does not work and was refuted by measurement: in all 6 cases kill(pid, 0) returned true, because the child is an exited-but-unreaped zombie and no synchronous probe distinguishes it. So the PR's reasoning that no deferral is possible holds, and this is a wording bug rather than an architecture bug. Dropping the reassurance leaves a sentence true in 100 percent of cases.

**4. MEDIUM. Agent stderr now reaches the chat transcript through a bypassable redactor.** Redaction is on the path and fires: 8 credential shapes in a real child's stderr, all 8 redacted, and the order is right, redact then slice, so truncation cannot re-expose. An intact key was then leaked into the user-visible banner through a real child by two mechanisms. First, redactSecrets' leading word boundary means a credential glued to preceding word characters is not redacted. Second, cutting even one character off the vendor prefix makes the whole credential survive, and ProcessAcpClient.ts:410 slices the raw ring at an arbitrary character. Both are pre-existing properties of the shared redactor; what this PR does is newly route that buffer into the transcript. Fix at the ring by dropping the partial leading record on truncation. The word-boundary weakness goes to a separate issue against the vendor-prefix rules.

Reported separately: ProcessAcpClient.ts:407 console.errors every raw stderr chunk unredacted, and the diff does not touch it, so the credential is already on disk regardless. That belongs with #1041.

## Suite

Full suite on the pristine head: 18213 tests, 9 failed. Three are the known host-PATH failures tracked as #1030. The other six were phantoms under load average 68, and all 56 tests in those five files pass in isolation, confirmed by re-run. The eight disconnect-relevant test files are 204 of 204 green at baseline. tsc was inconclusive, timing out at 10 minutes under concurrent load, so it is not verified here.

A fix lane is on all four findings, writing finding 2's tests before rewording so the reword is covered.

FerroxLabs:
## Audit fixes applied, all four findings closed

Four commits on top of 7047004a5, each mutation tested. New head eef6caefb.

### F2 first, deliberately (544aed27a)

The audit found the two lines that prevent a permanently stuck spinner had zero coverage: deleting either arm survived the whole suite. F1 required rewording the banner again, which is exactly the change that gap fails to catch, so the tests went in BEFORE the reword and then proved the reword was safe.

Both new assertions derive their input from the real buildCrashMessage rather than a hand-written literal, which is what let this through (the sibling suite pins a literal string that cannot notice a reword).

- tests/unit/process/acp/compat/AcpAgentV2.crashMarker.test.ts, new file, 3 tests
- tests/unit/team-TeammateManager.test.ts, one added test

Mutation: removing the transport arm at AcpAgentV2.ts:631 fails the AcpAgentV2 assertion (agentCrash finish frames 1 to 0). Removing TeammateManager.ts:670 fails the TeammateManager assertion (member status failed to idle). Both restored byte-identical afterwards, verified with diff.

Confirmed still passing after the reword landed: yes, unchanged, 112 of 112 across the five affected files.

### F1 and F3, the reword (2f45a164c)

AcpSession.ts:131 no longer instructs the user to resend a turn that PromptExecutor deliberately refuses to resend. It now warns that part of the message may already have run and to check the result first.

AcpSession.ts:127 no longer says the process may still be running. That was measured false for 6 of 20 real crashes. No probe fixes it, kill(pid,0) is true for the unreaped child, so the reassurance is simply dropped and the sentence is now true in every case.

Three existing assertions on the old reassurance were updated, not deleted, plus one new assertion for the safety property. Comments in IAcpClient.ts, ProcessAcpClient.ts and crashMarkers.ts that described the old claims were corrected with them.

Mutation: restoring the old line 127 fails 7 tests across 3 files. Restoring the old line 131 fails 3 tests across 2 files.

Not i18n. This string sits with AgentDisconnectedError, AgentStartupError and enterError, which are all raw English, so no locale work and no i18n-keys regeneration.

### F4, the ring (4280e02fb)

ProcessAcpClient.ts now truncates on a record boundary, dropping the partial leading line.

Reachability was established by execution, not by reading. The half-token sits at the START of the ring, where the banner's own 2048-character tail normally hides it. It stops hiding it when redaction SHRINKS the ring below 2048, which a credential block does: whole keys collapse to redacted, 8192 characters become about 1300, no tail slice happens, and the one token NOT redacted goes on screen.

Pre-fix, through a real child process: 12 of 14 new assertions fail, at cuts of 1, 3, 6, 10, 13 and 20 characters, ring level and banner level. Post-fix all 14 pass. Known positive in the same run: a whole credential in the same real stderr IS redacted, so a pass is the truncation being fixed rather than the scrubber being absent. A stderr tail with no newline at all keeps its whole 8KB rather than emptying itself, asserted separately.

The first version of this test passed against the unfixed code because the payload arithmetic put the cut in the filler. That arithmetic is now asserted inside the helper.

Mutation: restoring the plain slice fails 12 of 14.

### Filed, not fixed

- #1048, new. The scrubber's leading word boundary means a credential glued to preceding word characters is never redacted. Executed matrix over 5 vendor families and 8 preceding characters, 15 leaks out of 40, with the safe delimiters kept as controls. Left alone here because another lane is editing secretRedaction.ts for #1026 and #1037.
- #1041, commented rather than duplicated. ProcessAcpClient.ts:407 writes the raw chunk to console.error before anything scrubs it, so the credential is on disk regardless of the banner. Verified by execution: the key is verbatim in the console.error output and redacted in the banner built from the same chunk. Also verified this PR's diff does not touch that line.

### Left as verified correct, not changed

The evidence-only branch at AcpSession.ts:123, the alreadyExited guard, the equivalent mutant M10, and the 310-line prompt-test edit. The proposal to also treat process_exit and process_close as confirmed exits was measured to be a no-op for real traffic and is not implemented.

### Gates

tsc --noEmit clean. oxlint over exactly the 10 touched files: 10 files inspected, 0 errors, 12 warnings all pre-existing except one unavoidable empty-stub-class warning on a mock copied from the sibling suite. oxfmt clean on 10 files. No locale files touched.

FerroxLabs:
## Independent adversarial delta audit — head ad31a0be8

**Verdict: FIX-FIRST. Two HIGH findings, and both are regressions this PR introduces** — the same payloads are handled correctly by main and leak on this branch. Both reach the user-facing chat banner, which is persisted to the conversation database and the log file.

### Root cause, one bug in two shapes

ProcessAcpClient.ts:460-466 applies redactSecrets to a PREFIX of a stream it will receive more of later. redactSecrets is not resumable: masking a PARTIAL credential replaces its anchor with the redaction marker, so the continuation arriving in the next chunk has no anchor and is invisible to every later scrub, including the read-time scrub in buildStderrTail that is the only thing between the ring and the transcript.

The STDERR_CHUNK_SPLIT_CARRY_MAX carry mitigates exactly one case, a straddle living entirely inside the last 256 non-whitespace characters. Neither finding below is that case.

### HIGH-1: a straddle inside a non-whitespace run longer than 256 chars

Minified JSON on stderr is one long non-whitespace run, so the carry is disabled. 51 characters of an sk-ant-api03 shaped key reach the banner, with the redaction marker sitting immediately before the plaintext remainder, which is itself the proof that the scrubber ran and created the orphan.

    A ringLen 8190   LEAK_ring true   LEAK_banner true
    control, same payload with the run shortened to 191 chars: LEAK_banner false

Regression proof: replacing those lines with main's single slice statement gives LEAK_banner false. Main keeps the ring raw, so its read-time scrub sees the whole credential and masks it.

### HIGH-2: a PEM private key split across chunks

The carry is irrelevant here, PEM lines are 64 chars. The PEM rule has an end-of-input alternative, so an unterminated block matches from BEGIN to the end of the buffer: the head scrub consumes the anchor while the body is still arriving, and everything after has nothing to anchor to. Leaks the key body and the END line.

    B ringLen 8184   LEAK_ring true   LEAK_banner true
    same payload against main's behaviour: LEAK_banner false

Strictly more reachable than HIGH-1. It needs no unusual line shape, only a multi-KB PEM (a 4096-bit key is about 3.2KB) arriving in more than one write while total stderr exceeds 8KB.

### Fix direction

Stop scrubbing the retained buffer. Hold a ring of COMPLETE records: push each CR or LF delimited record, keep the pending partial separately, cap by dropping whole records off the front. redactSecrets is then only ever handed complete text at read time. The anchor-cut cases that motivated the collection-site scrub disappear too, because a record boundary never lands mid-word.

### Verified clean, no finding

The banner wording is now true for every state that can reach it: handlePromptError re-queues only on AUTH_REQUIRED without a tool run, so a transport drop is genuinely never resent. The prior finding is fixed. CRLF and bare-CR handling is correct, executed with 600 CRLF records. Over-truncation is not a defect: the root-cause record survives. No new win32 path hazard, and both previously Windows-red files now carry explicit skipIf guards for the shape that cannot occur there.

All PR test files are green on macOS.

### Not verified

The win32 shard was not run. Both leak banners were produced by calling buildCrashMessage on a real ring taken from a real child through the real capture path, not through a full live session.

FerroxLabs:
## Both HIGHs FIXED — commit 4e7008285 on the release integration branch

Verified before and after against a real child process, real stderr capture and the real buildCrashMessage.

    case A, straddle in a run longer than 256 chars
      before  LEAK_banner true. Banner read "key":"[redacted]lMnOpQrStUvWxYz..." with 51 plaintext
              characters sitting immediately after the marker that ate their prefix
      after   LEAK_banner false. Banner reads "key":"[redacted]", real diagnostic still present

    case B, PEM split across two writes
      before  LEAK_banner true. 12 body lines plus the END line in the banner
      after   LEAK_banner false. Whole block collapses to one marker, real diagnostic still present

### The design

The ring is now a list of COMPLETE records plus a separate pending fragment, with a private accessor
so all eight existing call sites are untouched. Chunks land in pending, complete CR or LF delimited
records split out, and the budget is enforced by shifting whole records off the front. The join is
lossless, so the ring is always a verbatim substring of the child's stderr.

redactSecrets is no longer imported by that file at all. That is the point: the invariant "collection
never scrubs" is now checkable by grep rather than by argument. Every consumer was checked and all of
them scrub at read time before any slicing.

Two degenerate cases, both handled deliberately rather than by accident:

A single record larger than the whole ring, such as a 21KB minified stack frame, now stops eviction
instead of emptying the ring, so its error still reaches the banner. The old code returned an empty
ring here and the user got no stderr section at all. Notably the fix does NOT scrub-and-truncate that
record even though that would arguably be safe, because keeping the no-scrub rule absolute is what
makes it verifiable.

Pending growth is bounded at 32768, four times the record budget. On overflow the fragment is frozen
as a final record and the stream is discarded until the next real boundary. Freezing keeps the head so
every retained credential keeps its anchor; discarding to a boundary guarantees the resumed stream
cannot begin mid-credential. A bound near 8192 would have thrown away the 21KB record's error text.

All four anchor-cut shapes the original commit existed to fix were re-verified green by execution, and
they now hold for a stronger reason: a record boundary never lands mid-word.

### Verification

32 tests green in the ring-truncation file; the 3 new ones fail on the reverted source, proven by
reverting, re-running and restoring with a checksum check. 63 further tests green across the disconnect,
banner, crash-marker, bridge and prompt suites. tsc exit 0, oxfmt run, oxlint clean, no POSIX path
literals added.

### Three test premises changed, declared rather than quietly rewritten

An assertion that the ring equals exactly the maximum was only achievable by cutting a record's head
off mid-stream, which IS the disclosure path; it now asserts the correct bounds with the substantive
assertions unchanged. Two assertions that the raw ring contains no secret only passed because the ring
used to be pre-scrubbed; they now assert on the scrubbed read, which is what the user is actually handed.

### Declared residual, equal to main, not a regression

Eviction can still drop a BEGIN PRIVATE KEY line while the body stays, orphaning it. That needs a PEM
whose BEGIN line falls more than 8192 characters behind the tail. Main cuts raw characters and has the
identical exposure. Closing it needs either a resumable scrubber or PEM-aware eviction, which is a
separate decision and was deliberately not invented here.

Verified on darwin only. Nothing in the change is platform-sensitive.

FerroxLabs:

Issue #1026, opened 2026-08-17
Secret scrubber misses prefixed labels: ANTHROPIC_API_KEY= survives redaction


FerroxLabs:
Found by the cross-audit of #1023, verified by execution as an actual leak. Pre-existing on main,
NOT introduced by that PR.

## Same root cause as the four escapes fixed in #1004

#1004 fixed patterns whose TRAILING word boundary could not match because the character class
omitted a word character. This is the mirror image: a LEADING boundary that cannot match because
the character before it is also a word character.

LABELLED_SECRET_ASSIGNMENT in src/process/utils/secretRedaction.ts:131 anchors a word boundary
before the label, roughly

    \b api[_-]?key ... =

There is no word boundary between an underscore and a letter, because both are word characters. So
a label with a prefix ending in an underscore never matches:

    ANTHROPIC_API_KEY=<20+ chars>     SURVIVES  (no boundary between _ and API)
    my_api_key=<value>                SURVIVES
    API_KEY=<value>                   masked    (boundary at start of line)

The failing shape is the COMMON one. Environment variables are conventionally prefixed, so the
provider-prefixed form is what actually appears in agent output and in error text.

## Also unmasked, verified

- a 40-character AWS secret access key
- npm_ prefixed tokens
- hf_ prefixed tokens

## Why it matters now

The redaction path is not decorative. It is what stands between untrusted subprocess output and
(a) the chat transcript, (b) the log file, and (c) any report the user copies and sends to support.
It is already reachable on main via AgentStartupError and the log transport.

It is also the defence-in-depth layer that a separate private advisory's fix relies on, so a hole
here weakens two things at once.

## Fix

Do not anchor on a word boundary before a label that can be preceded by an underscore. Match an
optional prefix segment, or anchor on a non-word-non-underscore boundary explicitly, or key off the
assignment shape rather than the label's left edge.

Then add the missing coverage class: a test that asserts the PREFIXED form of every labelled secret
is masked, not just the bare form. The existing suite tests bare labels, which is why this survived.

## Acceptance

ANTHROPIC_API_KEY=, OPENAI_API_KEY=, my_api_key=, a 40-char AWS secret, npm_ and hf_ tokens are all
masked, each proven by a test that FAILS before the fix. Confirm the matcher finds a known positive
before trusting any "no secret found" result — a redaction test that has never seen a real secret
survive proves nothing.

FerroxLabs:
[...]

Issue #1024, opened 2026-08-17
Corrupt engine config.toml bricks launch with "fix the file by hand" and no in-app recovery


FerroxLabs:
Reported by a user on macOS running engine wayland-core 0.13.0. They could not launch, spent
about two hours in a terminal trying to recover, and wrote: "no matter how many times I upgrade,
there are still tons of problems on my side and I am starting to lose hope" and "it seems this is
a software for only pro users that know exactly what they are doing."

## What they saw

    Agent failed to start: Cannot safely update the reserved [profiles.__wayland_desktop_session]
    table in the global Wayland Core config (existing content is not valid TOML: Invalid TOML
    document: invalid value). Fix the file by hand before Desktop can launch against it.

Doctor: 8 pass, 4 warn, 1 FAIL. The failure is Engine config integrity, and it identifies the
exact problem, in their engine config.toml:

    [security]
    enabled = falseegress_allow = [ "mcp.slack.com", "mcp-9827.slack.com", "slack.com" ]
    approval_mode = "yolo"

That is enabled = false and egress_allow = [...] concatenated with no newline between them.

## The refusal is correct. Do not change it.

DesktopProfileSpliceError in src/process/agent/wcore/desktopProfileSplice.ts is right to refuse.
That file holds the user's real providers, credentials, and memory and skills settings, so
discarding it would be catastrophic data loss rather than a benign reset. The code says so
explicitly and it is the correct call. This issue is NOT a request to make the splice more
permissive or to auto-reset the file.

## The defect is that there is no way out

The app tells a non-technical user to hand-edit TOML, and stops. Doctor already knows the exact
fix, names the file, points at the offending line and column, and even suggests removing the file
to regenerate defaults. Nothing in the app will do any of it. The only repair path is a terminal,
which is what cost this user two hours and nearly the relationship.

A repo-wide search for a repair affordance finds the string only inside the Doctor check that
prints advice. There is no implementation behind it.

## What is wanted

An in-app recovery path for a corrupt engine config that respects the constraint above, so it
must be surgical rather than a reset:

1. Show the user the offending line and column. See the security note below before echoing any
   file content.
2. Offer a one-click repair that takes a timestamped backup of config.toml first, then either
   applies the specific fix or, only with explicit confirmation and after the backup, regenerates
   defaults. Never silently discard credentials.
3. Offer to reveal the file in Finder or Explorer as a fallback, so the user is not left hunting
   for a path in Library.

## Security constraint on any fix

[...]

PR #1029, opened 2026-08-17
Sanitise engine config parse errors before they reach the UI

FerroxLabs:
Closes the credential-disclosure defect in private advisory **GHSA-2g2m-r86j-jg6h**. Mechanics are in the advisory, not here.

## The shared helper

`desktopProfileSplice.ts` already stripped the echoed source out of this exact class of parse error and documented why. `doctor/registry.ts` passed `error.message` through untouched. Two components, opposite conclusions about the same data, and the one on the copy-to-support surface was the one that leaked.

That logic is now `src/process/utils/tomlErrorSummary.ts` and **both** call sites use it:

| Call site | Before | After |
| --- | --- | --- |
| `agent/wcore/desktopProfileSplice.ts` | private `summarizeTomlError` | imports the shared helper |
| `doctor/registry.ts` → `doctor/engineConfigProbe.ts` | raw `error.message` | imports the shared helper |

It keeps the first line (the human-readable reason), drops the echoed block, then runs the survivor through `redactSecrets` from `utils/secretRedaction.ts`. No new patterns, no fork of that module.

The strip happens at the **producer** so no consumer can leak it. That producer moved out of `registry.ts` into `doctor/engineConfigProbe.ts` — a security boundary reachable only through Electron singletons is a boundary nobody can test, and this one now has no Electron dependency.

## Still actionable

`smol-toml`'s `TomlError` carries `line` and `column` as numbers, so the position is reported and the text is not:

```
detail:      The engine's config.toml could not be parsed at line 3, column 40: Invalid TOML document: each key-value declaration must be followed by an end-of-line
remediation: Fix the TOML syntax at line 3, column 40 in the engine config.toml, or remove the file to regenerate defaults.
```

## Doctor-surface audit

All 13 registered checks, and what each puts in front of the user:

| Check | Free-form error text? | Credential-bearing source? | Action |
| --- | --- | --- | --- |
| `config.engineConfig` | raw `smol-toml` message | **yes** — the user's real `config.toml` | **the advisory defect.** Strip at producer + scrub |
| `mcp.servers` | probe `result.error` (HTTP body / spawned stderr) | **yes** — the declaration carries `env` API keys, `headers.Authorization`, and the bearer `McpService.attachOAuthToken` adds | scrub |
| `backends.detected` | `AgentRegistry.getLoadErrors()`, raw `error.message` | **yes** — managed-install manifests and remote-agent rows | scrub |
| `engine.contractPin` | raw fs `error.message` | no — reads the engine **binary** | scrubbed for uniformity |
| *(runner catch-all)* | raw thrown `error.message`, for every check | unknown by construction | scrub |
| `providers.connectivity` | no — `ConnectError` is a closed union (`unauthorized` / `no-credit` / `offline` / …) | reads creds, never echoes them | none needed |
| `models.registry` | no — counts and provider ids only | n/a | none needed |
| `engine.reachable` | no | n/a | none needed |
| `engine.routing` | no — counts only | n/a | none needed |
| `workspace.drift` | no — paths and project/chat names | n/a | none needed |
| `workspace.configured` | no — same | n/a | none needed |
| `config.paths` | no — two static resolved dirs | n/a | none needed |
| `config.appArchitecture` | no — `process.platform` / `arch` | n/a | none needed |
| `config.secretStorage` | no — a boolean | n/a | none needed |

The uniform rule is now: no raw error text reaches a Doctor detail. A per-check judgement call is what the next check author gets wrong.

## Verified by execution

`tests/unit/process/doctor/engineConfigParseErrorRedaction.test.ts` writes a real corrupt `config.toml` whose malformed line sits immediately below a realistic `api_key`, then runs the real `readConfig`, the real producer and the real check. **Fail-before against this branch's base:** 2 assertions failed, and the failure output was the credential itself. `tests/unit/process/doctor/doctorSurfaceSecretRedaction.test.ts`: 5 of 6 failed against the base check files.

Both files carry a **known-positive control** asserting the matcher *does* find the key in the unsanitised text — a "no secret found" result is worthless without it. Post-fix: 14/14 pass.

## The scrubber is a backstop, not the fix

`redactSecrets` has a confirmed hole (**#1026**, owned elsewhere, not touched here): its labelled-assignment rule anchors a word boundary before the label, so `ANTHROPIC_API_KEY=<value>` and `my_api_key = "<value>"` survive whenever the value carries no recognisable prefix. Verified by execution.

So the engine-config fix does **not** rest on the scrubber. The assertion that carries it — `drops the echoed source block entirely` — asserts the block is *gone*, not that the credential inside it was masked. Comments and test names say which layer each assertion depends on, and the sweep tests state plainly that they prove those free-form sinks no longer pass *recognisable* secrets through, not that they cannot leak.

## Notes

- **No i18n change.** Only the check *title* is an i18n key (`settings.doctor.checks.engineConfig`, untouched). Doctor `detail` / `remediation` are English literals produced in the main process and rendered verbatim — confirmed by grepping the locale tree. No 12-locale work, no `i18n-keys.d.ts` regeneration.
- `src/common/adapter/bridgeAllowlist.ts` is byte-identical to `main` (`git diff` against `ferrox/main` is empty).
- Full unit suite: **18164 passed, 3 failed**. The 3 are `acpAgentManagerNpmFallback.test.ts`, pre-existing and host-caused (a real `wayland-nano` at `/Users/seandonahoe/.local/bin/wayland-nano` defeats the not-on-PATH fixture). `tsc --noEmit` clean, `oxlint` clean, `oxfmt` on touched files only.
- Dependency tree checked: `bun install` reports no changes, so no stale-`node_modules` artifacts in the suite result.
- Out of scope by design: the in-app recovery UI for #1024, and #1026.

### DIFF
--- a/src/process/agent/wcore/desktopProfileSplice.ts
+++ b/src/process/agent/wcore/desktopProfileSplice.ts
@@ -17,6 +17,7 @@
  * formatting.
  */
 import { parse } from 'smol-toml';
+import { summarizeTomlError } from '@process/utils/tomlErrorSummary';
 import { WCORE_DESKTOP_MCP_PROFILE } from './envBuilder';
 
 /**
@@ -44,27 +45,6 @@ export class DesktopProfileSpliceError extends Error {
 
 /** Any TOML table-header line, e.g. `[providers.anthropic]` or `[[a.b]] # c`. */
 const TABLE_HEADER_LINE_RE = /^\[{1,2}\s*([^\]]+?)\s*\]{1,2}/;
-
-/**
- * A `smol-toml` parse error message echoes the OFFENDING SOURCE LINE verbatim
- * after a blank line, e.g.:
- *
- *   Invalid TOML document: each key-value declaration must be followed by ...
- *
- *   1:  api_key = "sk-ant-REDACTED-EXAMPLE" oops
- *                                           ^
- *
- * This file operates on the user's REAL global `config.toml`, which holds
- * `api_key` values. Embedding the raw message would carry a live credential
- * into an Error that is surfaced, logged, and (via K-02) shown in the UI. Keep
- * only the first line - the human-readable reason - and drop the echoed source
- * block entirely. Found by the K-01 4-leg cross-audit (Gemini 3.1 Pro leg) and
- * confirmed by executing the parser against a key-bearing malformed line.
- */
-function summarizeTomlError(error: unknown): string {
-  const raw = error instanceof Error ? error.message : String(error);
-  return raw.split('\n', 1)[0].trim();
-}
 
 /**
  * Line indices that begin INSIDE a TOML multi-line string (`"""` or `'''`).
--- a/src/process/doctor/checks/backendChecks.ts
+++ b/src/process/doctor/checks/backendChecks.ts
@@ -49,7 +49,9 @@ export async function checkBackends(reader: BackendReader): Promise<DoctorCheckO
       // `AgentRegistry.loadErrors` are raw `error.message` strings from reading the
       // managed-install manifests and the remote-agent rows, both of which carry
       // auth material. Scrub before they reach a Doctor report that gets copied to
-      // support - same class as GHSA-2g2m-r86j-jg6h.
+      // support - same class as GHSA-2g2m-r86j-jg6h. Best-effort: `redactSecrets`
+      // misses the prefixed label form (#1026), and free-form loader text has no
+      // structure to strip instead.
       detail: `${available.length} backend(s) detected (${names}); ${loadErrors.length} loader error(s): ${redactSecrets(loadErrors.join('; '))}.`,
       remediation: 'A configured backend failed to load — check its configuration in Settings → Agents.',
     };
--- a/src/process/doctor/checks/backendChecks.ts
+++ b/src/process/doctor/checks/backendChecks.ts
@@ -14,6 +14,7 @@
  * sub-detector reported a load error.
  */
 
+import { redactSecrets } from '@process/utils/secretRedaction';
 import type { DetectedAgent } from '@/common/types/detectedAgent';
 import type { DoctorCheckOutcome } from '../types';
 
@@ -45,7 +46,11 @@ export async function checkBackends(reader: BackendReader): Promise<DoctorCheckO
   if (loadErrors.length > 0) {
     return {
       status: 'warn',
-      detail: `${available.length} backend(s) detected (${names}); ${loadErrors.length} loader error(s): ${loadErrors.join('; ')}.`,
+      // `AgentRegistry.loadErrors` are raw `error.message` strings from reading the
+      // managed-install manifests and the remote-agent rows, both of which carry
+      // auth material. Scrub before they reach a Doctor report that gets copied to
+      // support - same class as GHSA-2g2m-r86j-jg6h.
+      detail: `${available.length} backend(s) detected (${names}); ${loadErrors.length} loader error(s): ${redactSecrets(loadErrors.join('; '))}.`,
       remediation: 'A configured backend failed to load — check its configuration in Settings → Agents.',
     };
   }
--- a/src/process/doctor/checks/configChecks.ts
+++ b/src/process/doctor/checks/configChecks.ts
@@ -17,6 +17,7 @@
  *     (a fresh install has none).
  */
 
+import { redactSecrets } from '@process/utils/secretRedaction';
 import type { DoctorCheckOutcome } from '../types';
 
 /** Config check dependencies — all injectable so the checks are unit-testable. */
@@ -25,9 +26,17 @@ export type ConfigCheckDeps = {
   isEncryptionAvailable: () => boolean;
   /**
    * Read + parse the engine's user `config.toml`. Resolves `'ok'` (parsed or
-   * absent), or `'corrupt'` with the parse error message. Never throws.
+   * absent), or `'corrupt'` with the parse failure's reason and, when the parser
+   * reported one, its position. Never throws.
+   *
+   * `message` is the reason ONLY. It must never carry the offending source line:
+   * that line is the user's own config, credentials included
+   * (GHSA-2g2m-r86j-jg6h). `line`/`column` are how this check stays actionable
+   * without echoing any of the file. `probeEngineConfig` is the real producer.
    */
-  readEngineConfig: () => Promise<{ status: 'ok'; existed: boolean } | { status: 'corrupt'; message: string }>;
+  readEngineConfig: () => Promise<
+    { status: 'ok'; existed: boolean } | { status: 'corrupt'; message: string; line?: number; column?: number }
+  >;
 };
 
 /**
@@ -48,18 +57,45 @@ export async function checkSecretStorage(isEncryptionAvailable: () => boolean): 
 }
 
 /**
+ * Render `line`/`column` as a position phrase, or `null` when the parser gave
+ * neither. Deliberately NUMBERS only - naming the position is what keeps the
+ * remediation actionable now that the offending line's TEXT is withheld
+ * (GHSA-2g2m-r86j-jg6h).
+ */
+function describePosition(result: { line?: number; column?: number }): string | null {
+  if (typeof result.line !== 'number' || !Number.isFinite(result.line)) return null;
+  if (typeof result.column !== 'number' || !Number.isFinite(result.column)) return `line ${result.line}`;
+  return `line ${result.line}, column ${result.column}`;
+}
+
+/**
  * Engine config integrity — the user `config.toml` parses. FAIL when it exists
  * but is corrupt; PASS when it parses or is absent (a fresh install).
+ *
+ * The corrupt branch names the POSITION of the fault and never its text. The
+ * Doctor panel has a "Copy report" button and exists to be shared with support,
+ * so echoing the offending config line would hand out whatever credential
+ * happened to sit next to it (GHSA-2g2m-r86j-jg6h).
  */
 export async function checkEngineConfigIntegrity(
   readEngineConfig: ConfigCheckDeps['readEngineConfig']
 ): Promise<DoctorCheckOutcome> {
   const result = await readEngineConfig();
   if (result.status === 'corrupt') {
+    // `readEngineConfig` is INJECTED, and the real producer already strips and
+    // scrubs. Scrub again anyway: every future caller of this check inherits the
+    // copy-to-support blast radius, and a defence that depends on the injection
+    // being the right one is not a defence. `redactSecrets` is idempotent.
+    const reason = redactSecrets(result.message);
+    const position = describePosition(result);
     return {
       status: 'fail',
-      detail: `The engine's config.toml could not be parsed: ${result.message}`,
-      remediation: 'Fix the TOML syntax in the engine config.toml, or remove the file to regenerate defaults.',
+      detail: position
+        ? `The engine's config.toml could not be parsed at ${position}: ${reason}`
+        : `The engine's config.toml could not be parsed: ${reason}`,
+      remediation: position
+        ? `Fix the TOML syntax at ${position} in the engine config.toml, or remove the file to regenerate defaults.`
+        : 'Fix the TOML syntax in the engine config.toml, or remove the file to regenerate defaults.',
     };
   }
   return {
--- a/src/process/doctor/checks/configChecks.ts
+++ b/src/process/doctor/checks/configChecks.ts
@@ -35,7 +35,9 @@ export type ConfigCheckDeps = {
    * without echoing any of the file. `probeEngineConfig` is the real producer.
    */
   readEngineConfig: () => Promise<
-    { status: 'ok'; existed: boolean } | { status: 'corrupt'; message: string; line?: number; column?: number }
+    | { status: 'ok'; existed: boolean; path?: string }
+    | { status: 'corrupt'; message: string; path?: string; line?: number; column?: number }
+    | { status: 'unresolved'; message: string }
   >;
 };
 
@@ -81,6 +83,16 @@ export async function checkEngineConfigIntegrity(
   readEngineConfig: ConfigCheckDeps['readEngineConfig']
 ): Promise<DoctorCheckOutcome> {
   const result = await readEngineConfig();
+  // A profile that cannot be resolved is a DIFFERENT fault from a config that
+  // will not parse, and calling it a parse failure is the same misdiagnosis this
+  // check already shipped once by reading the wrong file (see `probeEngineConfig`).
+  if (result.status === 'unresolved') {
+    return {
+      status: 'fail',
+      detail: `The active profile's config directory could not be resolved: ${redactSecrets(result.message)}`,
+      remediation: 'Switch back to the default profile, or repair the active profile directory, then re-run Doctor.',
+    };
+  }
   if (result.status === 'corrupt') {
     // `readEngineConfig` is INJECTED, and the real producer already strips and
     // scrubs. Scrub again anyway: every future caller of this check inherits the
@@ -93,21 +105,26 @@ export async function checkEngineConfigIntegrity(
     // scrub over an unstripped parse error would still leak.
     const reason = redactSecrets(result.message);
     const position = describePosition(result);
+    // The PATH is named in the detail because the check and the recovery panel
+    // that mounts under this row must visibly agree on which file they mean; when
+    // they silently disagreed, Reveal opened a file nobody was complaining about.
+    const where = result.path ? ` (${result.path})` : '';
     return {
       status: 'fail',
       detail: position
-        ? `The engine's config.toml could not be parsed at ${position}: ${reason}`
-        : `The engine's config.toml could not be parsed: ${reason}`,
+        ? `The engine's config.toml${where} could not be parsed at ${position}: ${reason}`
+        : `The engine's config.toml${where} could not be parsed: ${reason}`,
       remediation: position
-        ? `Fix the TOML syntax at ${position} in the engine config.toml, or remove the file to regenerate defaults.`
-        : 'Fix the TOML syntax in the engine config.toml, or remove the file to regenerate defaults.',
+        ? `Fix the TOML syntax at ${position} in ${result.path ?? 'the engine config.toml'}, or remove the file to regenerate defaults.`
+        : `Fix the TOML syntax in ${result.path ?? 'the engine config.toml'}, or remove the file to regenerate defaults.`,
     };
   }
+  const target = result.path ? ` (${result.path})` : '';
   return {
     status: 'pass',
     detail: result.existed
-      ? 'Engine config.toml parses cleanly.'
-      : 'No engine config.toml yet (fresh install) — defaults apply.',
+      ? `Engine config.toml${target} parses cleanly.`
+      : `No engine config.toml yet${target} (fresh install) — defaults apply.`,
   };
 }
 
--- a/src/process/doctor/checks/configChecks.ts
+++ b/src/process/doctor/checks/configChecks.ts
@@ -86,6 +86,11 @@ export async function checkEngineConfigIntegrity(
     // scrubs. Scrub again anyway: every future caller of this check inherits the
     // copy-to-support blast radius, and a defence that depends on the injection
     // being the right one is not a defence. `redactSecrets` is idempotent.
+    //
+    // This is a BACKSTOP, not the fix. The fix is the producer DROPPING the
+    // echoed source block (`probeEngineConfig`): `redactSecrets` misses the
+    // prefixed label spelling `ANTHROPIC_API_KEY=<value>` entirely (#1026), so a
+    // scrub over an unstripped parse error would still leak.
     const reason = redactSecrets(result.message);
     const position = describePosition(result);
     return {
--- a/src/process/doctor/checks/configChecks.ts
+++ b/src/process/doctor/checks/configChecks.ts
@@ -33,10 +33,16 @@ export type ConfigCheckDeps = {
    * that line is the user's own config, credentials included
    * (GHSA-2g2m-r86j-jg6h). `line`/`column` are how this check stays actionable
    * without echoing any of the file. `probeEngineConfig` is the real producer.
+   *
+   * `displayPath` is RENDERED IN PLACE OF `path` whenever the producer sets it.
+   * Under a named profile the path's own segment is the user-authored profile
+   * name, so naming the real path names the name; `path` still exists because the
+   * recovery panel and Reveal-in-Finder need the real file. Same split, same
+   * reason, as `WorkspaceEntry.displayPath`.
    */
   readEngineConfig: () => Promise<
-    | { status: 'ok'; existed: boolean; path?: string }
-    | { status: 'corrupt'; message: string; path?: string; line?: number; column?: number }
+    | { status: 'ok'; existed: boolean; path?: string; displayPath?: string }
+    | { status: 'corrupt'; message: string; path?: string; displayPath?: string; line?: number; column?: number }
     | { status: 'unresolved'; message: string }
   >;
 };
@@ -68,6 +74,18 @@ function describePosition(result: { line?: number; column?: number }): string | 
   if (typeof result.line !== 'number' || !Number.isFinite(result.line)) return null;
   if (typeof result.column !== 'number' || !Number.isFinite(result.column)) return `line ${result.line}`;
   return `line ${result.line}, column ${result.column}`;
+}
+
+/**
+ * The path to render, preferring the producer's withheld `displayPath`. `null`
+ * when the producer named no path at all.
+ *
+ * A helper rather than four inline `??`s because there ARE four render sites - two
+ * fields on the corrupt branch and two spellings on the healthy one - and the leak
+ * this closes existed because a reader assumed there was one.
+ */
+function renderedPath(result: { path?: string; displayPath?: string }): string | null {
+  return result.displayPath ?? result.path ?? null;
 }
 
 /**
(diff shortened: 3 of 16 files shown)


FerroxLabs:

Four legs. Codex 5.6 Sol: NO-GO, and the strongest leg — it independently found F3 and the line-split
gap. Gemini 3.1 Pro: FIX-FIRST, but it got stuck in plan mode and EXECUTED NOTHING, so its only
unique contribution is F4, which the internal leg then confirmed by execution. Kimi K3: DID NOT
COMPLETE, provider 403 usage limit for the billing cycle, so no findings and none invented.
Internal: FIX-FIRST.

## The advisory's own path is structurally closed, and the proof is better than the fix

Established by reading the installed parser rather than trusting the change: all 33 TomlError throw
sites in smol-toml 1.6.1 pass STATIC string literals, and error.js:58 builds the message as
"Invalid TOML document: ${reason}" then a blank line then the codeblock. The codeblock is therefore
always after the first newline, so taking the first line cannot carry file bytes.

Verified across 12 adversarial fixtures with a passing known-positive control on 11 of them: CRLF,
CR-only line endings, no trailing newline, a literal backslash-n inside the offending line, an
unterminated triple-quote swallowing the key, and the worst case of api_key on line 1 with the error
on line 1. tomlErrorPosition reads only the numeric line and column; codeblock is an enumerable own
property but is never read and never spread, because the probe builds a fresh object.

Fail-before was also independently reproduced from main's own files: the detail contained the full
sk-ant key, the base_url, the echoed block and the newlines. The defect was real and the premise is
correct.

## But three of the four fixed checks are scrub-only, so they still leak

Only config.engineConfig is structurally fixed. Executed with a real-shaped prefixed key
(AZURE_OPENAI_API_KEY with a 32-hex value), both controls passing:

    mcp.servers        LEAK
    backends.detected  LEAK
    engine.contractPin LEAK
    runner catch-all   LEAK

That is the #1026 hole: a leading word boundary cannot match after an underscore, so the PREFIXED
form survives while bare API_KEY= is masked. The prefixed form is the common one.

To be fair to the PR: its own comments state this accurately. The second commit's correction is
honest and the final diff does not overclaim.

## Must fix before merge

**F1 (HIGH) — mcpChecks.ts:139. A structural fix exists that the code says does not.** The comment
argues there is no structure to strip in free-form probe output. That is the wrong side of the pipe:
the DECLARATION is in scope. Executed scenario: a stdio MCP server dies with
AZURE_OPENAI_API_KEY=<32-hex> on stderr, McpProtocol.ts:438 appends 300 characters of it as "Server
output: ...", and the whole key lands in the copied report. Fix: before or after redactSecrets,
replace literal occurrences of every value in server.transport.env (stdio) and
server.transport.headers (http/sse/streamable_http) with a redaction marker. Deterministic, about
five lines, and immune to #1026 because it does not pattern-match at all.

**F3 (MEDIUM) — engineChecks.ts:57, declared CLEAN in the audit table and not clean.** detectWCore
returns unvalidated, unbounded execFileSync stdout from "wayland-core --version" and the detail
interpolates it verbatim. Executed: it renders an injected key inside the reachability line. Fix:
match a semver pattern and surface only the capture.

**F4 (LOW/MEDIUM) — workspaceChecks.ts:48 and :119, also declared clean.** The label is
Chat "${conversation.name}" or Project "${name}", and conversation names are auto-generated from the
user's FIRST MESSAGE. So a user who pastes a credential into chat gets it as the chat title and then
as a line in a shareable report. Executed: leaks. Fix: wrap the two map templates in redactSecrets.

Each of the three needs a PREFIXLESS canary added to the sweep test, or the same class walks back in.

## Deferred with sign-off

**F2 — backendChecks.ts:52** is scrub-only with no structural option, so it is genuinely blocked on
#1026. The audit produced the minimal fix for that issue: replace the leading word boundary in
LABELLED_SECRET_ASSIGNMENT (secretRedaction.ts:131) with a negative lookbehind for
[A-Za-z0-9]. Being handled on #1026.

**F5 — tomlErrorSummary.ts:62** splits only on newline, so a CR-only or U+2028-separated message
would survive whole. Not reachable through smol-toml 1.6.1 or Node fs errors, so not a defect today.
Worth widening the split if the file is touched again.

Declined as scope creep: an open-branded ProviderId in providers.connectivity and models.registry,
and config.paths printing the two config directories, which it does by design.

## Verified good

The lane's strongest claim holds: ConnectError is a nine-member literal union and all 16 error
assignments in ConnectionTester are literals or classifyStatus, which returns only literals, so free
text cannot reach providers.connectivity.

No test relaxed, skipped or deleted: 8 modified, 3 added, ZERO deleted, and no pre-existing test file
touched. No desktopProfileSplice regression — all 28 static smol-toml reasons pass through the shared
helper byte-identical despite the added scrub, its canary assertions still hold, wcore suite 96 of 96.
Remediation stays actionable because TomlError always carries both numbers. bridgeAllowlist.ts is
byte-identical by shasum on both sides. No generated constitution file. The i18n claim is correct:
only the title key is i18n and it already exists in all 12 locales; detail and remediation render
verbatim.

FerroxLabs:
## FIX-FIRST addressed — F1, F3, F4 fixed in `6414df01f`

Two of the three are now **structural**, so they no longer depend on the scrubber at all. Every canary uses a prefixed-label secret with a 32-hex value, i.e. the spelling `redactSecrets` provably cannot see.

### F1 (HIGH) — mcpChecks: the audit was right, and the comment was wrong

The comment claimed there was no structure to strip. That was the wrong side of the pipe: the **declaration** is in scope. Every user-supplied value in `server.transport` — stdio `env` and `args`, http/sse/streamable_http `headers` — is now masked in the probe output by **literal string replacement**. No pattern is consulted, so no credential shape can slip past and #1026 is irrelevant to it. `redactSecrets` still runs afterwards, because it is what covers the OAuth bearer `attachOAuthToken` injects at probe time — that token is never in the stored declaration this check reads. The misleading comment is replaced with that reasoning.

Two details worth flagging:

- **An 8-char floor.** Without it, an `env` of `PORT=3000` blanks "3000" out of the probe's own error text. There is a test asserting a short non-secret value survives.
- **A canary caught a gap in my own fix.** A key declared as `--api-key=<value>` was *not* masked when the server echoed the bare value, because only the whole declared token was matched. Values are now matched both whole and unwrapped past a leading `--flag=` / `Bearer ` prefix, longest-first. Found by execution, not review.

### F3 (MEDIUM) — engineChecks: confirmed, my audit table was wrong

`detectWCore` returns unvalidated, unbounded `execFileSync` stdout and reachability interpolated all of it. It now matches `/v?\d+\.\d+\.\d+[\w.+-]*/` and surfaces only the capture — an allowlisted shape, not a scrub. Stdout carrying no version number is treated exactly like no stdout and is **not** echoed to say so.

The optional `v` is load-bearing: the existing `checkEngineReachable` test feeds `v0.10.0` and asserts that spelling survives. My first pattern dropped it and **broke that test**. I fixed the pattern, not the test.

### F4 (LOW/MEDIUM) — workspaceChecks: confirmed, fixed as directed, with one honest correction

Both labels at `:48` and `:119` are now wrapped in `redactSecrets`.

**The prefixless canary cannot pass here, and I am not going to claim it does.** Verified by execution:

```
redactSecrets('Chat "Help me rotate AZURE_OPENAI_API_KEY=a1b2…f90 please"')
  → unchanged
```

A chat title **is** user prose — unlike a TOML parse error or an MCP declaration there is nothing structural to strip — so this sink is scrub-only and blocked on **#1026** in exactly the way F2/`backendChecks` is. Its canaries therefore use a prefixed-**value** secret, which does fail-before and pass-after and does prove the scrub is wired. Recommend F4 be tracked with F2 under #1026; it needs no further change in this file once #1026 lands, because the fix arrives through the same call.

I deliberately did **not** pin the gap with an assertion. A test asserting "we still leak" would red out the #1026 lane's CI without adding protection — the scrub call is already the regression guard. The limitation is recorded at both call sites and on the test's describe block instead.

### Fail-before / pass-after

Six new assertions, run against the three pre-fix files restored via `git show HEAD:<path>` (never `git stash`):

```
FAIL-BEFORE (pre-fix, commit 288dd5add)      PASS-AFTER
6 failed | 7 passed (13)                     13 passed (13)

× PREFIXLESS CANARY: masks a declared env value the server echoed on stderr
× PREFIXLESS CANARY: masks a declared header value and a declared arg
× PREFIXLESS CANARY: surfaces only the version number, never the rest of the banner
× warns without echoing stdout when it carries no version number
× scrubs the label in the drift check
× scrubs the label in the configured check
```

### Deferred, as instructed

`secretRedaction.ts` untouched — **F2** and **F5** left alone, and the scope-creep items declined. F5 remains worth doing whenever that file is next opened.

### Verification

Full unit suite **18171 passed, 3 failed** — the same three pre-existing `acpAgentManagerNpmFallback` host failures (a real `wayland-nano` on PATH), zero new. `tsc --noEmit` clean. `oxlint` back to the pre-existing 2 warnings after I cleared the one I introduced. `oxfmt` on the four touched files only. `bridgeAllowlist.ts` byte-identical by `shasum` on both sides (`26a5959f5…`). No test relaxed, skipped or deleted — the one existing test my change broke was fixed at the source.

FerroxLabs:
## FIX-FIRST #2 addressed — FF-1…FF-7 plus the config-target bug, in `22883cccf`

Four of the seven are now closed **structurally**. Where a fix is scrub-only, the code says so rather than claiming immunity.

### FF-1 (BLOCKER) — engine version: the allowlist was not one

Confirmed. `[\w.+-]*` unanchored and unbounded meant the match ran straight through a credential — and the `token=sk-ant-1.2.3-…` case needed no banner at all because the search started *inside* the token. Adopted the suggested pattern plus the cap:

```
/(?<![\w.+-])v?\d+\.\d+\.\d+(?:[-+][0-9A-Za-z][0-9A-Za-z.]{0,10})?(?![\w.+-])/  +  .slice(0, 32)
```

All six adversarial inputs are now permanent test cases asserting the 40-char token, the JWT, the 32-hex key and the 200,000-char tail are all absent and `detail.length < 200`. The three legitimate spellings (`v0.10.0`, `wayland-core 0.13.0 (env …)`, `banner v0.10.0 extra`) plus `0.13.0-rc.1` are asserted to still pass. The comment no longer claims general immunity — only immunity to what cannot fit the shape.

### FF-2 / FF-3 / FF-4 — mcpChecks

- **FF-2** `transport.url` added. Canary covers all three leaking shapes (path-embedded, `?t=`, `?azure_openai_api_key=`) with a `getaddrinfo` echo.
- **FF-3** the 3-line catch in `probeWithTimeout`. This was the one that mattered most — it was a *bypass*, not a gap. Canaries cover both the check directly and end-to-end through `runDoctor`, plus one asserting a malformed server no longer destroys the other servers' detail.
- **FF-4** `trim()` plus **iterative** unwrap. Both `mcp-remote --header` spellings covered against both echoes (bare token and `Bearer <token>`), and the trailing-space env value.

### FF-5 — residual stated, not asserted away

You're right on both halves: the token is genuinely absent from the stored declaration, and `redactSecrets` only catches it with the literal prefix. The comment now records that `"invalid token: Bearer <opaque>"` masks and `"invalid token: <opaque>"` does not, and that closing it means threading the probe-time authed server back through the `testConnection` contract. **I did not take that option** — it changes a contract outside this check, and I'd rather flag it than half-do it. Say the word if you want it.

### FF-6 — floor wrong in both directions, and the acceptance criterion

`env`/`headers` now mask from **4** characters; `args` contribute **only** the remainder past a separator, never the whole arg.

The readable-diagnostic proof, executed:

```
A) 1 of 1 enabled MCP server(s) failed to load: filesystem (MCP error -32000: Connection closed.
   Server output: npm ERR! 404 Not Found - GET
   https://registry.npmjs.org/@modelcontextprotocol/server-filesystem - Not found.)
B) 1 of 1 enabled MCP server(s) failed to load: filesystem (ENOENT: no such file or directory,
   scandir '/Users/alice/Documents')
```

Both intact. There is also a test asserting a **1-3 character** env value is *not* masked — `DEBUG=1` would otherwise blank every "1" in the text, which is worse than the leak it prevents. Why 4 and not 0 is documented at the constant.

Accepted limits of literal matching, documented and not chased: case-folded echoes, URL-encoded echoes, and a value glued to a short flag with no separator (`-k<secret>`). That last one is the deliberate cost of FF-6's args rule — closing it would mean masking whole args again.

### FF-7 — you were right, and there is a structural fix

I took the **producer fix**. `doctorEntityLabel` emits the app-generated id, so the name never reaches a label. The label builders moved out of `registry.ts` into `workspaceInventory.ts` for the same reason `engineConfigProbe` exists — a boundary reachable only through Electron singletons can't be tested. The `path` half is scrubbed too.

The known-positive control is now a permanent test: `redactSecrets('Chat "<bare 32-hex>"')` returns it **unchanged**. That is the proof the scrub-only plan could never have closed this, sitting in the suite next to the fix.

Nothing actionable was lost — the workspace **path** is the half users act on and it is still reported. `F2` left deferred as agreed.

### Routed from #1031 — the config target. **The panel was right; my check was wrong.**

Established by execution before changing anything, not inherited from either lane's view:

```
active=work
ACTIVE = /…/profroot-hBByKe/work/config.toml
NATIVE = /Users/…/Library/Application Support/wayland-core/config.toml
same   = false
```

And the trace: `WCoreAgent.resolveWaylandHomeForLaunch` sets `WAYLAND_HOME` from `resolveActiveConfigDir()`, and Core treats `$WAYLAND_HOME` as the **literal** config dir. So the active profile's `config.toml` *is* the file the engine launches against. `readConfig()` already defaulted to it — only my caller was overriding it with the native path.

Three changes: the probe resolves `resolveActiveConfigPath()`; the detail and remediation **name the inspected path** so the row and the panel cannot silently disagree again; and an unresolvable profile returns a new `unresolved` status reported as *"The active profile's config directory could not be resolved"* rather than as a parse failure — calling a broken profile a corrupt config is the same misdiagnosis in a different coat.

Note: `inspectEngineConfig` isn't in my tree yet (unmerged #1031), so the agreement test asserts against `resolveActiveConfigPath()` — the shared contract both sides resolve through — and separately that the target is *not* `resolveUserConfigPath()`. There's a known-positive asserting the two paths genuinely diverge under a named profile, so the assertions can't hold vacuously on a `default`-profile machine.

### Fail-before / pass-after

Sixteen new assertions, all red against `6414df01f`, all green after:

```
FF-1…FF-6 + config target   13 failed | 29 passed   →   128 passed
FF-7 (label producer reverted specifically)  3 failed
```

The fail-before output contained both FF-6 symptoms verbatim — `hunter7` leaking, and `registry.npmjs.org/[redacted]` shredded.

### Verification

Full suite **18192 passed, 3 failed** — the same three pre-existing `acpAgentManagerNpmFallback` host failures, zero new. `tsc --noEmit` clean. `oxlint` back to the pre-existing 2 warnings (I cleared the four import orphans and the one nested function my extraction created). `oxfmt` on the 9 touched files via `xargs`. `bridgeAllowlist.ts` byte-identical by `shasum` (`26a5959f5…`). `secretRedaction.ts` untouched. No generated constitution file. AGENTS.md reverted — thanks for the heads-up, the hook had stamped it.

**No i18n keys changed.** Doctor `detail`/`remediation` are main-process literals rendered verbatim; grepped all five new/changed strings against the locale tree, zero hits, and the diff adds no `titleKey` or `t()` reference. So no 12-locale work and no `i18n-keys.d.ts` regeneration.

FerroxLabs:
## Re-audit verdict on 22883cccf: FIX-FIRST

Legs: Codex 5.6 Sol FIX-FIRST, completed on the correct PR with real executed probes and no provider refusal, strongest leg, and four of its claims that I had not tested were confirmed by my own execution. Gemini 3.1 Pro DID NOT COMPLETE, shell unavailable, executed nothing; of its three checkable claims one verified TRUE and two verified FALSE. Kimi K3 DID NOT RUN, provider 403 usage limit. Internal execution leg FIX-FIRST.

## The three headline fixes work, and there are zero regressions

FF-1's replacement regex held against 24 adversarial version banners; I could not get a credential through it. summarizeTomlError was driven against 8 real corrupt TOML files on disk and the key never left probeEngineConfig. Across 14 credential shapes and 11 sinks there is not one cell where main masks and head leaks. Codex reached the same conclusion independently.

The control that makes those zeros mean something: my first harness reported "mask" for engine.reachable and mcp.servers on main as well as head, which is impossible. Both were inert, because WCoreDetection uses available rather than found and checkMcpServers filters on server.enabled. After fixing the harness, main leaks all 14 shapes at both sinks and head masks all 14.

Two sinks mask all 14 including the prefixless shapes, because they use structure rather than patterns: an allowlist, and literal declaration matching.

## Findings

**F1 HIGH. mcpChecks.ts:259, 267, 271, 296 render the user-authored server.name raw in all four branches.** A server named as a bare 32-hex string surfaces verbatim in the errored, needsAuth, toolless and timedOut details, executed. This is exactly the conversation.name defect the previous pass established and this PR fixed by switching to an app id, left unfixed in a file this PR rewrote 173 lines of. IMcpServer.id already exists at src/common/config/storage.ts:932.

**F2 HIGH. The id-label fix does not remove project.name, because it re-enters through the path.** collectConfiguredWorkspaces emits a label plus a path and the drift detail renders both. The default project workspace is Documents/Wayland/<project-name>, so for projects the name IS the path. workspaceChecks.ts:55 claims emitting the id makes the leak "UNREACHABLE rather than unlikely"; that is false for projects, executed.

**F3 MEDIUM. FF-1's fix has no discriminating oracle, and the partial-leak floor is 25 characters.** Reverting the anchors alone leaves the suite green. Removing the 32-char slice alone leaves it green. Only reverting both goes red, because the oracles are whole-literal checks and the slice truncates every payload just enough to satisfy them. With anchors reverted and slice kept, 25 characters of a credential surface with the suite green.

**F4 MEDIUM. summarizeTomlError fails open on the one non-TOML error that reaches it.** probeEngineConfig's unresolved branch passes a ProfileIsolationError whose message interpolates the profile name on the first line, so layer 1 is inert and layer 2 misses a bare value, executed. Also latent: bare-CR and the Unicode line and paragraph separators survive the first-line split, and there is no length cap, so a 500KB single-line message returns all of it. For real smol-toml 1.6.1 the first line never carries file content, verified across 8 fixtures, so the primary path is safe.

**F5 MEDIUM. Declaration masking misses the space-separated flag form.** Executed with a prefixless secret: a two-element args pair leaks, a lone separator-free arg leaks because the unwrap discards it, and a token embedded as a URL path segment leaks because the unwrap never splits on a slash. Env whole, header whole, header bare, the equals form and the colon form all mask correctly. This contradicts the file's claim that no credential shape can slip past the sink, and the space-separated flag form is not in its documented accepted-limits list.

**F6 LOW.** An Error subclass whose message getter throws makes runner.ts:58 throw, so runDoctor rejects and the secondary message escapes, violating its never-throws contract. Exotic reachability, hardening only.

**F7 LOW.** Detail length is unbounded: a 2,000,000-character probe error yields a 2,000,052-character detail, and 10MB with 200 env vars costs about 1.08s of synchronous masking. Not a regression, main was also uncapped.

## Explicitly not findings

The version allowlist deliberately surfaces up to 11 attacker-chosen alphanumeric characters in the build tail. Documented, bounded, and the source is the app's own resolved binary. Gemini's claim that this bypasses the 32-char cap is false; the slice always applies.

Pre-existing and not regressions, belonging to other issues: prefixed labels (#1026), suffixed labels (#1037), a bare 32-hex string with no entropy rule, and the npm and Hugging Face prefixes which are absent from SECRET_PATTERNS entirely and are being added on the #1026 branch.

## Mutation and suite

12 of 15 mutants caught. The three survivors are F3's two plus a dropped redactSecrets backstop in the summarizer. Killed mutants include the wrong-config-target revert, the single-pass unwrap revert, and the floor at 8.

Full suite at settled load average 10, after an earlier run at load 77 produced 8 phantom failures: 15,120 tests, 3 failed, all three the known host-PATH failure #1030. Doctor-related 139 of 139. tsc clean. oxlint on the 13 touched files: 0 errors and 1 warning, an asPath closure moved verbatim out of registry.ts and present on main. oxfmt clean. bridgeAllowlist.ts and webserver/adapter.ts are untouched, confirmed by git diff on those paths rather than by counting.

A fix lane is on all of it.

FerroxLabs:
## All seven re-audit findings closed at 7405e1635. Six atomic commits, pushed.

### F1, mcpChecks rendered the user-authored name raw in all four branches

`mcpChecks.ts` gains `doctorServerLabel(server)` returning `server.id`, called at all four branches (now 262, 270, 274, 299).

Executed before: a server named as a bare 32-hex string appeared verbatim in all four details, including "failed to load", "need authentication", "published no tools" and the timeout line. After, each prints an app-generated id and the name is absent. Control established first: a disabled server yields "No MCP servers are enabled." and reaches nothing, and the scrubber leaves the bare value untouched, which is why the label had to change rather than the redaction.

Verified the id is genuinely app-generated rather than user-influenced: `newMcpServerId()` is the only assignment and both add paths take an Omit that excludes id.

Mutation: reverting each of the four branches independently goes red, four for four.

Four pre-existing assertions in `checks.test.ts` that asserted the name were updated to assert a distinct id. That is strictly stronger than before, because a revert to name now reds them too. No assertion was removed or weakened.

### F2, the name re-entered through the path

`workspaceInventory.ts:113` adds `doctorWorkspaceDisplayPath`; both entry types gain `displayPath`; both checks render `displayPath ?? path` while still stat'ing and classifying the real `path`. `registry.ts:137` binds the base dir through the newly exported `defaultWorkspaceBaseDir()` rather than rebuilding the literal.

Executed before: the drift detail printed both the label and a path whose leaf was the project name verbatim. After: "Project proj-1 to /Users/a/Documents/Wayland/(folder name withheld)". Known positive first: `resolveProjectWorkspacePath` returns exactly base plus the bare name, which is what proves the leaf IS the name rather than merely resembling it.

**Two false claims corrected in source.** `workspaceInventory.ts:51` claimed the leak was "UNREACHABLE rather than unlikely"; that is replaced with a correction stating what was measured. The matching claim at `workspaceChecks.ts:57` likewise. A false claim in a load-bearing security comment is the specific failure that produced #1004, #1026 and this very re-audit.

Decided and documented in source: paths outside the app's own base stay verbatim, because they are user-chosen, not derived from an entity name, and they are the half the user acts on. Over-withholding inside the base is accepted. The win32 cross-drive guard is deliberately absent, because for a redaction decision a false "contained" is the safe error, the inverse of an access check, and an unpinned guard covering the one case this host cannot exercise is worse than none.

Mutation: rendering the raw path in either check goes red; inverting the withholding goes red; dropping the case-insensitive arm goes red; replacing the separator-aware walk with a bare startsWith('..') goes red. Plus a test proving the real path is still stat'd, so drift cannot be silently hidden behind the display value.

### F3, no discriminating oracle, and the floor was 25 characters

Reproduced the re-audit exactly on the dispatch head first: anchors reverted alone GREEN, slice removed alone GREEN, both RED. That is the finding.

Three oracles added, plus a `longestSecretRun` helper so partial disclosure counts rather than only whole-literal survival.

On the final head, mutation-tested one at a time: **anchors reverted alone now RED, slice removed alone now RED.** The anchor oracle asserts each adversarial banner is refused rather than trimmed, which is the one thing the cap cannot fake. The cap oracle uses the only input that can reach 32 characters while the anchors hold, so it matches under both patterns and moves only with the cap.

**Floor achieved: 2 characters**, measured across eight banners against five credentials, and both of those two are incidental collisions with the fixed English copy rather than credential material. It was 25.

`engineChecks.ts` is untouched since dispatch, so the version allowlist that was explicitly not a finding was not modified.

### F4, the summarizer failed open on the one non-TOML error that reaches it

`engineConfigProbe.ts:42` adds two constants and `profileUnresolvedReason`, and the unresolved branch no longer routes through the helper at all. `tomlErrorSummary.ts` splits on carriage return, line feed and both Unicode separators, and caps at 200.

Executed before: the detail rendered the profile name. After, the name is absent while the detail still says which fault it is.

The profile name is withheld rather than replaced by an id because there is no separate id: the name pattern admits a 32-hex key and an sk-ant token, and the invalid-marker branch passes whatever the marker file held. The error's detail argument is dropped with it, because on two of its three call sites it is a filesystem message carrying the profile path.

The cap is measured rather than picked: ten real corrupt TOML files gave reason lines of 36 to 87 characters, and the key never reached line one in any of them.

Order is scrub then cap, and that is pinned, because the scrubber floors at 8 characters (measured: a 7-character fragment behind a label comes back untouched, 8 comes back masked), so cap-first would hand out a 7-character fragment.

Mutation: newline-only split red; dropping bare CR alone red; dropping the paragraph separator alone red; dropping the cap alone red; **dropping the scrubber backstop alone red**, which closes the third surviving mutant from the re-audit; flipping scrub and cap order red; routing the unresolved branch back through the helper red.

### F5, the space-separated flag form

`mcpChecks.ts` keeps a separator-free arg whole when the preceding token names a credential, with the transport command counting as the token before the first arg, and contributes URL segments of 8 or more characters after stripping scheme and authority so a hostname is never masked. The minimum declared length stays 4.

All measured with a prefixless value, because a prefixed one would be masked by the scrubber regardless and any pass would be vacuous. Known positive first: an undeclared prefixless value is not masked. Now masked: the two-element flag pair, a credential-naming command plus bare arg, a URL path segment and a query value, plus the five that already worked. Still readable: the hostname, a short route segment, the package name and the directory.

**False claim corrected**: the "no credential SHAPE can slip past it" claim at `mcpChecks.ts:172` is replaced with a correction explaining that shape-blindness is not coverage, and the coverage claim now points at the accepted-limits list.

**Left open by design and pinned as such**: a lone secret-looking arg behind a generic command is not masked, because it is byte-identical in kind to a package name, and masking it whole is exactly the earlier regression this file already carries a note about. There is a test asserting it leaks, so the trade is visible rather than assumed.

### F6 and F7

F6: `runner.ts:74` adds `safeErrorMessage`, so an Error subclass whose message getter throws no longer escapes `runDoctor`. Mutation red.

F7 was done rather than filed, because it landed inside a file F6 already touched and cost one slice. `runner.ts:26-67` caps detail and remediation at 4,000 characters with a truncation marker, applied once in `runOne` so every outcome passes through it. 4,000 is set against the largest legitimate detail, roughly 3,000 for a twenty-server failure, and a 3,000-character real detail is pinned untouched so the cap cannot start eating diagnostics. Three mutations red.

### Gates

tsc clean. oxlint on exactly the 11 touched files: **11 files inspected**, 1 warning, 0 errors, the warning being a closure moved verbatim from another file and present on main. oxfmt: 11 files correct. Full vitest: **18,233 passed, 3 failed, 37 skipped** at load average 7.09, and all three failures are the known host-PATH issue tracked as #1030. Constitution suites green with cargo on PATH.

Untouched and verified by git diff on those paths rather than by counting: `bridgeAllowlist.ts`, `webserver/adapter.ts`, `constitutionFsAuthority.generated.ts`, `migrations.ts`, and `secretRedaction.ts` so there is no collision with the concurrent scrubber lane.

FerroxLabs:

Issue #1030, opened 2026-08-17
acpAgentManagerNpmFallback tests depend on the host PATH and fail wherever wayland-nano is installed


FerroxLabs:
Found repeatedly by lanes today, and independently confirmed by two of them.

## The problem

tests/unit/process/task/acpAgentManagerNpmFallback.test.ts consults the REAL PATH. On any machine
that has a genuine wayland-nano installed, the resolver correctly prefers the user's own binary and
the test's expectation fails.

On this build machine there is a real ~/.local/bin/wayland-nano (about 10.7M), so 3 tests fail on
every full-suite run. Re-running the same suite with a scrubbed PATH gives 5 of 5 green.

## Why it is worth fixing rather than tolerating

The suite has become a permanent 3-failure baseline that every lane has to be told to ignore. That
is corrosive in a specific way: it trains everyone, human and agent, to wave off failures in a
release-adjacent test file. It has already cost time in several lanes this session, each of which
had to independently establish that these three are host artefacts rather than regressions, and it
adds noise to exactly the signal we rely on before tagging.

It also means the test does not test what it claims. It asserts a fallback path while silently
depending on the absence of a real installation, so on a developer machine it verifies the opposite
of what it was written for.

## Fix

Isolate the resolver from the host environment: inject the lookup rather than reading process.env
PATH, or scrub PATH for this suite so resolution is deterministic. Then add the case that is
currently unreachable on a developer machine, which is the one where a real binary IS present and
should be preferred, so both branches are covered on every host.

## Acceptance

The suite passes on a machine WITH a real wayland-nano on PATH and on one WITHOUT, and both
resolution branches are asserted explicitly rather than depending on which machine ran it.

FerroxLabs:
## Sharper diagnosis, from a lane that kept tripping over this

The three failures are not simply "depends on the host PATH". The test asserts a spawn of the pinned npm package, but on a developer box it resolves a REAL locally installed binary at `~/.local/bin/wayland-nano`, 10.7 MB, which exists here. The test's PATH stubbing does not hide it, so the assertion compares against the wrong resolution and fails.

That means the failure is deterministic on any machine that happens to have Nano installed locally, and absent on one that does not, which is why it looks like flakiness and why it survives every "it passes for me" check made on a clean runner.

[...]

PR #1031, opened 2026-08-17
In-app recovery for an invalid engine config.toml (#1024)

### LINKED ISSUE #1024: Corrupt engine config.toml bricks launch with "fix the file by hand" and no in-app recovery

FerroxLabs:
Closes #1024.

A user's engine `config.toml` became invalid TOML and Desktop refused to launch with a sentence telling them to hand-edit TOML. They spent about two hours in a terminal. This gives them a way out that cannot lose their data.

## The refusal is unchanged, and still correct

`spliceDesktopMcpProfile` still declines to touch an unparseable global `config.toml`, because that file holds the user's real providers, credentials and memory/skills settings. Nothing here loosens that guard, auto-resets the file, or makes the splice tolerate invalid TOML. The only change to `desktopProfileSplice.ts` is exporting `multilineStringLineStates` so the repair reuses the same data-vs-structure scanner instead of carrying a second copy of that logic.

## What was built

**`src/process/agent/wcore/engineConfigRecovery.ts`** — the safety-critical half.

- `inspectEngineConfig()` returns the path, the 1-based line and column, a scrubbed one-line reason, and whether an automatic fix exists. It returns **no file content**.
- `repairEngineConfig()` applies a repair that provably **inserts only line breaks** — enforced by comparing both texts with every `\n` removed — and only when it is **unambiguous**: exactly one candidate break position makes the whole document parse. Two working candidates produce two different files, so it refuses to guess. Lines that begin inside a `"""`/`'''` string are refused outright, since a newline there rewrites user data even if the result parses.
- `regenerateEngineConfig()` requires an explicit `confirmed: true` and is re-checked in the bridge handler.
- Every write path calls `createVerifiedBackup()` **first**: a timestamped sibling (no colons — NTFS alternate-data-stream trap), created with `wx` so an existing backup is never overwritten, then read back and proven identical by length **and** sha256. An unproven backup throws and the caller writes nothing at all.

**Surfaced in both places the user hits the wall:**

- **Launch path** — `WCoreManager.emitStartFailure` now stamps `DesktopProfileSpliceError.code` on the terminal error frame, exactly as it already does for a locked Constitution. `WCoreSendBox` routes that code (never the prose) to a new in-thread `WCoreEngineConfigCard`.
- **Doctor** — the `config.engineConfig` row renders the same panel inline when it FAILS. Keyed on the check **id**, not on its detail/remediation text, so the message-sanitisation work happening to that check in parallel cannot unmount the recovery actions. `buildDoctorReportText` is untouched, so the copyable report gains nothing new.

**The shared panel** (`EngineConfigRecoveryPanel`) offers, in descending preference:

1. Fix the identified line — only when main found an unambiguous fix.
2. **Show me the file** — the no-risk escape hatch, present in **every** state including when the app did not understand the failure.
3. Start over with defaults — behind a confirmation that names exactly what is lost (provider list, stored API keys, memory and skills settings) and still runs the verified backup first.

## Security

No `config.toml` content crosses the process boundary or reaches the UI. Only the path, the line/column integers, and the parser's first message line — routed through the shared `redactCommandSecrets` scrubber, on top of the first-line-only rule `desktopProfileSplice.ts` already documents (the echoed source line can be an `api_key`). A test asserts the serialized inspection payload contains none of `sk-ant`, `egress_allow`, `approval_mode` or `enabled` for the reported file.

**Bridge channels: 4 added, all remote-denied.** `engine-config-recovery.{inspect,repair,regenerate,reveal}`. The remote denylist goes 144 → 148. None of them accepts a path from the renderer — main resolves the active profile's `config.toml` itself. That is also why `reveal` is implemented in the new bridge instead of reusing `show-item-in-folder`: that channel runs `confinePath`, and the engine config dir is not one of `pathConfinement`'s authorized roots, so the reveal would be rejected. Revealing the one path main computed itself widens no root set.

## Verification by execution

Backup-before-write ordering, including the failure branches, is proven by the write log of an injected fs seam:

- success: `writes == [backup, config.toml]`, backup holds the original bytes, config parses;
- backup write denied (`EACCES` / `ENOSPC`): `writes == []` and the original is byte-for-byte unchanged;
- backup read-back does not match: only the backup was written, the original untouched;
- regenerate without confirmation: `writes == []`, file still present;
- regenerate confirmed: `writes == [backup, unlink:config.toml]`, credentials still in the verified backup.

Against a **real** scratch config dir (never Sean's profile), using the real `fs` implementations: `spliceDesktopMcpProfile` throws `DesktopProfileSpliceError` on the reported file **before** the repair and succeeds **after** it, with `api_key` and the deliberate `[security]` edits intact — i.e. the app can launch against the repaired file.

Also pinned: the renderer's copy of the error code against `DesktopProfileSpliceError.code`; the four denylist entries, with a mutation guard proving the unqualified spellings would not have protected them.

## i18n

26 new keys under `conversation.engineConfigInvalid.*`, added to `conversation.json` in **all 12 locales**, with `i18n-keys.d.ts` regenerated. `node scripts/check-i18n.js` passes and reports the type definition in sync.

## Suite

Full `vitest run`: **18196 passed, 4 failed, 37 skipped**. All 4 failures reproduce on clean `main` (29b71da20) in a detached baseline worktree: 3 in `acpAgentManagerNpmFallback.test.ts` (a real `wayland-nano` on this host's PATH) and 1 load-sensitive 60s timeout in `constitutionFsService.test.ts` (machine load average was 55). All 292 renderer DOM test files pass. `tsc --noEmit` clean; `oxfmt --check` clean on every touched file.

Screenshots of the before/after failure state are in the lane report.


FerroxLabs:
## Cross-audit verdict: FIX-FIRST — two blockers, both data loss

Legs: Codex 5.6 Sol FIX-FIRST (completed, executed real probes, 5 findings — its two strongest were
independently reproduced). Gemini 3.1 Pro returned GO but **EXECUTED NOTHING** — its transcript shows
"Tool run_shell_command not found" twice — and its claim that the confirmation gate halts on a falsy
`confirmed` is CONTRADICTED by execution, so its GO is unweighted. Kimi K3 DID NOT COMPLETE: provider
403 usage limit for the billing cycle, the fifth audit today lost to that, so no findings and none
invented. Internal: FIX-FIRST.

## Lead answers

**Can the renderer influence which file is written, revealed or deleted?** Not via a path — no channel
accepts one, main resolves it, and extra/prototype-shaped/unexpected fields are ignored by all four
handlers. But this PR DOES grant the renderer a capability it did not have: `remove-entry` and
`write-file` on the engine config dir are rejected today by `confinePath` because that dir is not an
authorized root, and `engine-config-recovery.regenerate` deletes that exact file.

**Can any path lose the user's credentials? YES, reproduced.**

## F1 — BLOCKER. The "verified byte-identical" backup is not byte-exact, and then the original is deleted

`engineConfigRecovery.ts:364` reads with `readFile(path, 'utf-8')`. Node substitutes U+FFFD for invalid
byte sequences DURING that decode. The sha256 at `:298-300` is computed over the resulting STRING, and
the comparison at `:351` therefore compares a mangled string to itself and ALWAYS passes.
`isLineBreakOnlyEdit` (`:176`) is blind for the same reason, and repair was proven to silently rewrite a
0xE9 byte out of the live file.

Executed on a real scratch filesystem: original 92 bytes containing 0xE9, backup 94 bytes with EF BF BD
substituted, byteIdentical=false, and `config.toml` deleted at `:495`. The original bytes are
unrecoverable. A truncated or legacy-encoded config is exactly the corruption class this feature exists
to handle, so this is not a corner case.

Fix: back up by `rename(configPath, backupPath)` — atomic, byte-exact, no decode, and it removes the
TOCTOU window for free. Repair then writes to configPath and renames back on failure; regenerate needs
no unlink at all. Decode for PARSING with TextDecoder utf-8 fatal, and refuse the automatic repair when
the decode is not lossless.

## F2 — BLOCKER. regenerate does not require the file to be invalid

`:470-498`. Executed: with `inspectEngineConfig()` returning status ok on a perfectly valid,
credential-bearing config, `regenerateEngineConfig({confirmed:true})` returned ok:true and deleted it.
The only guard is a boolean the caller supplies. Fix: re-run `findParseProblem(source)` and return
nothing-to-repair when the file parses.

## F3 — MEDIUM. The confirmation is a truthiness check, not identity

`:474`. Executed: `{confirmed:'true'}` and `{confirmed:1}` both delete. Not renderer-reachable today
because the bridge coerces, but the module's own header claims it re-checks. Fix: `!== true`.

## F4 — MEDIUM. Predictable temp name, non-exclusive write, follows a symlink

`:368-372`. Executed: pre-planting the temp path as a symlink to another file caused repair to OVERWRITE
that file with the repaired TOML and still return ok:true. Same-user only, but it is a symlink-following
write outside the config dir — the exact class confinePath exists to stop. Fix: random suffix plus
flag wx, and unlink the temp on failure (it currently orphans a 0600 file containing credentials).

## F5 — MEDIUM. The Doctor row and the panel resolve DIFFERENT files

Doctor's config.engineConfig check reads resolveUserConfigPath (the NATIVE config); the panel mounted
under that row inspects resolveActiveConfigPath (the ACTIVE PROFILE's config). Executed with a named
profile active: Doctor's target corrupt, the panel's target fine, panel reports ok. So the row fails
forever, the panel says there is no problem, and Reveal opens the wrong file. Profile activation is
shipped UI, so this is not latent. The check-side fix belongs with #1029.

## F6 — MEDIUM. Panel handlers have no catch, so a WebUI click is an unhandled rejection

`EngineConfigRecoveryPanel.tsx:88-98,:100-113,:115-123` have try/finally but no catch. On the remote
transport all four invokes correctly reject with remote-forbidden, but chat.response.stream IS
remote-allowed so the card renders in WebUI, inspect fails into unreadable, and Reveal throws instead of
showing revealFailed.

## F7 — LOW. MAX_REPAIR_STEPS bounds steps, not candidates per step

Super-linear in bare `key =` shapes on the failing line, synchronous on the Electron main thread, in a
channel that auto-fires on panel mount. Measured 2.9s at 400 glued assignments, over 4s at 1000, and a
2-minute timeout beyond. The realistic #1024 shape is 2ms.

## F8 — LOW but doctrinal. Four exact keys where the file's own rule says use a prefix

`bridgeAllowlist.ts:609-612`. Executed: `engine-config-recovery.setPath` would be remote-ALLOWED, while
`terminal.anythingNew` is denied — because terminal., workspaceTrust. and waylandTransfer. are PREFIXES,
added specifically so a future provider cannot become remotely reachable by omission. A namespace where
3 of 4 channels write or delete a credential file should follow that rule.

## Verified sound — do not re-litigate

The repair logic is genuinely excellent. Fuzzing deleted each of 25 newlines from a realistic config
(multi-line strings, literal strings, inline table, multi-line array, comments): 9 repaired, 2 refused,
ZERO violations — every repair newline-insertion-only, parsed, and semantically identical to the original
document. Ambiguity refusal works on every crafted multi-parse line. Comments survive; CRLF keeps every
carriage return.

The denylist entries are load-bearing: every unqualified spelling was proven unprotected, and enforcement
is real at the webserver adapter before dispatch plus a client-side fail-fast. The reveal/confinePath
decision is justified and strictly narrower than widening authorizedRoots — both external legs agreed,
and showItemInFolder performs no write.

No config content crosses the boundary, and because the defence is STRIPPING rather than scrubbing the
#1026 hole is not load-bearing: five adversarial configs including a secret on the failing line and on
line 1 produced no canary in the serialized payload. desktopProfileSplice is 6 added / 1 removed (a
comment plus an export keyword) and its refusal was proven to still fire before the repair and clear
after, with api_key and the deliberate [security] edits intact. Doctor files are byte-untouched and the
mount is keyed on the check id. i18n verified in all 12 locales with genuine native translations.
Backups are timestamped, colon-free, win32-legal, never overwrite, and land at 0600; a real EACCES gives
reason backup-failed with zero backups on disk and the original sha unchanged.

Process note: the PR body refers to screenshots but none are attached to the body or comments, so the UI
could not be visually verified; it was judged from the code, the DOM tests and the localized copy.

FerroxLabs:
## Cross-audit fixes — all 8 findings addressed, each reproduced before it was fixed

**Visual + byte-level evidence, including the F1 buffer table and all five screenshots:** https://claude.ai/code/artifact/712de26c-655e-4b3d-b2c0-06349bfaa93b

The repair logic the audit fuzzed and validated is unchanged — the ambiguity refusal, the newline-insertion-only invariant, the multi-line-string guard, CRLF handling and comment survival are all as they were. What changed is the filesystem half, which is where both blockers lived.

### F1 (BLOCKER) — the byte-identical check was blind, then deleted the original

Reproduced first. A config holding a lone `0xE9`:

| | before | after |
|---|---|---|
| original | 96 B · `267b5b8ac9ad…` | 96 B · `267b5b8ac9ad…` |
| backup | **98 B · `6749ab52da4c…`** | 96 B · `267b5b8ac9ad…` |
| `Buffer.equals` | **false** | **true** |
| `EF BF BD` present | **yes** | no |
| `config.toml` | **deleted** | moved to the verified backup |

The old code copied through `readFile(path, 'utf-8')`; Node substitutes U+FFFD during that decode, so the sha256 was taken over the mangled string and compared to itself — `length equal = true, sha equal = true` while the file gained two bytes.

Taking the audit's prescription: **the backup is now a `rename`** — atomic, byte-exact, decode-free. Verification moved to raw `Buffer.equals` against the bytes the repair was planned from, which also closes the read-then-write TOCTOU window (a file that changed underneath us is renamed back and refused). Reserving the backup name is a separate `wx` 0-byte create, so a concurrent recovery cannot land on an older backup.

Separately, parsing now decodes with `new TextDecoder('utf-8', { fatal: true })`. A file that is not losslessly UTF-8 reports `encodingLossy` and is **never offered an automatic repair** — verified byte-exact through all three required fixtures (lone `0xE9`, truncated `E2 82`, UTF-16LE with a non-ASCII char). One honest nuance is documented in a test: UTF-16 of *pure ASCII* does decode under a fatal decoder, because NUL is a valid code point — it re-encodes to the identical bytes, so nothing can be lost, and the rename makes the guarantee independent of the decoder either way.

### F2 (BLOCKER) — regenerate deleted a valid config

Reproduced: `inspect` returned `ok` on a healthy credential-bearing config and `regenerate({confirmed:true})` returned `ok:true` and deleted it. Now re-parses and returns `nothing-to-repair`; the file is untouched. A lossy-encoded file stays eligible, since regenerating is the only in-app way out of a file that is not even text.

### F3 — confirmation was truthiness

Reproduced: `{confirmed:'true'}` and `{confirmed:1}` both deleted. Now `options?.confirmed !== true`. Executed against `'true'`, `1`, `{}`, `[]`, `'yes'` and a missing options object — all `not-confirmed`, file intact; `true` still works (known positive).

### F4 — predictable temp name, non-exclusive write, followed a symlink

Reproduced: a symlink planted at `config.toml.wayland-recovery-<pid>.tmp` was overwritten with the repaired TOML while the call returned `ok:true`. **There is no temp file at all now** — after the rename the path is free, so the repair is a single exclusive `wx` create (which fails on an existing symlink, dangling or not). The whole class is gone by construction rather than patched. A test plants that exact symlink and asserts the victim file is byte-unchanged.

### F5 — Doctor row and panel resolve different files

Confirmed. My half: the resolved path now renders in **every** branch — `ok`, `missing`, `unreadable` and `invalid` — so a user looking at a failing row can see the panel inspected `profiles/work/config.toml` instead (screenshot in the evidence page). **The check-side fix is still needed** and is called out in a comment on `inspectEngineConfig` for the #1029 lane; I did not touch `doctor/registry.ts` or `doctor/checks/configChecks.ts`.

### F6 — handlers had `finally` but no `catch`

All three write/reveal handlers now catch and set an outcome line. Three tests reject the invoke the way the remote transport does and assert `writeFailed` / `revealFailed` renders instead of an unhandled rejection.

### F7 — cost bounded by steps, not candidates

Measured 400 glued assignments at 220 ms here and growing super-linearly, synchronous on the main thread. Now capped by failing-line bytes (4 KB), candidate count (128) and document size (512 KB). After: n=400 / 1000 / 4000 all return in **0 ms** with no repair offered; the realistic 2-candidate shape still repairs.

### F8 — prefix, not exact keys

Reproduced: `engine-config-recovery.setPath` was remotely **allowed** while `terminal.anythingNew` was denied. Now denied by prefix, following the rule `waylandTransfer.` states in that file. The four exact keys are removed as redundant.

**Deny count: 144 → 145.** Breakdown `133 keys + 8 prefixes + 4 config-key prefixes`. Computed by parsing the three collections with comments stripped (a naive line count reads **148**, which is the trap); the same parser reads `ferrox/main` as exactly `133 + 7 + 4 = 144`, which is the known positive that makes the new number trustworthy. `git diff` against `ferrox/main`: **1 prefix added, 0 exact keys added, 0 removed.**

### Verification

The whole executed matrix was re-run after the rework, since the rename changes every write path. The original survives every failure branch of both operations, proven on a **real** filesystem as well as an injected seam — an unwritable directory gives `backup-failed` from both, the config byte-for-byte intact and no half-made backup left behind; a failed repair write renames the original back.

Service tests: **54 passing**. Renderer panel + Doctor wiring: **14 passing**. Allowlist redteam: **8 passing**, including a mutation that the prefix is not so loose it denies a neighbouring namespace.

One extra change beyond the findings: the card's title said *"has a typo"*, which is false for the non-UTF-8 branch — nothing was mistyped, the bytes are not text. Retitled to *"needs a fix"*, true of both faults, retranslated in all 12 locales. Two new keys (`notText`, `repair.notTextUnavailable`) were added across all 12 locales with real translations; `i18n-keys.d.ts` regenerated and `check-i18n` reports the type definition in sync.

FerroxLabs:
## FIX-FIRST addressed. All four findings closed at 3a117cb5a, four atomic commits.

### F1, the cost product, is now bounded by one cumulative budget

`engineConfigRecovery.ts:178-247` adds `MAX_REPAIR_PARSE_BYTES` (2 MiB) and `MAX_REPAIR_MILLIS` (250) behind a `ParseBudget`, charged at every candidate parse and every per-step parse. On exhaustion it bails the whole step rather than returning a partial sweep, because the ambiguity check is only sound once every candidate has been tried. The four old caps stay as cheap pre-filters and the comment no longer claims they are the cost bound.

**A correction to the audit's mechanism, found by execution.** The cost depends on the corrupt line's POSITION, not just document size and candidate count. A parse aborts at the error, so corruption on line 1 is cheap; the expensive regime is corruption near the END of a large document, where every re-parse chews through the whole valid prefix. The first reconstruction put the glued line on line 1, measured 8.4ms, and the finding looked unreproducible until the line moved to the end. Also the loop runs up to `MAX_REPAIR_STEPS`, so the real parse count is steps times candidates, 876 rather than 117.

Before and after, same process, same document, quiet machine at load 9, measured with a counting wrapper around the parser. One parse of a healthy 514 KB config is 72.8ms.

| bytes | candidates | BEFORE parses / MB / ms | AFTER parses / MB / ms |
|---|---|---|---|
| 514,161 | 116 | 876 / 429.5 / **52,937** | 4 / 2.0 / **232** |
| 513,625 | 83 | 588 / 288.0 / **38,166** | 4 / 1.96 / **303** |
| 132,913 | 116 | 876 / 111.0 / **4,652** | 15 / 1.90 / **66** |
| 514,687 | 152 (old cap fires) | 1 / 0.49 / 54 | 1 / 0.49 / 55 |
| 170 (the real #1024 shape) | 24 | 27 / 1ms, repairs | 27 / **0ms, repairs** |

52.9 seconds to 232ms, a 228x improvement, and the first three rows all sat UNDER every individual cap, which was the finding. The realistic #1024 shape still repairs and its api_key survives verbatim, asserted. The residual 232ms is one unavoidable parse, because inspect must parse the document once to know it is broken at all.

Accepted cost, CORRECTED after the delta re-audit. This paragraph previously said a config over 512 KB needing two or more breaks now gets no automatic fix. That was wrong by roughly an order of magnitude and the real boundary is far lower, so the honest numbers are below.

The budget is 2 MiB of parsed bytes and every parse costs the WHOLE document, so the boundary is about 2 MiB divided by the number of parses, and the number of parses is about (breaks x candidates on the glued line) + 2. It therefore depends on the shape of the corrupt line and not on 512 KB at all. Measured by binary search against 1a911e75b as the control, on the glued line the #1024 reporter actually had, which carries 24 candidates:

| breaks needed | HEAD repairs up to | HEAD refuses at |
|---|---|---|
| 1 | 80,652 B | 80,703 B |
| 2 | 41,092 B | 41,142 B |
| 4 | 20,722 B | 20,772 B |

On a simpler 8-candidate glued line the same search gives 209,620 / 210,058, then 87,064 / 87,632, then 31,132 / 31,558. Both sets fit the same formula, which is the cross-check that makes them trustworthy: 2 MiB / (24 + 2) = 80,659 against the 80,652 measured, and 2 MiB / (8 + 2) = 209,715 against the 209,620 measured.

So two breaks fails above roughly 41 KB on the shape actually reported, and four breaks above roughly 21 KB. A 21 KB config needing four breaks is not an absurd artifact, so this is a real capability limit and it deserves stating accurately rather than as a 512 KB figure nothing measures.

The pre-budget control repairs at every one of those sizes, so these are genuinely new boundaries and not pre-existing refusals. The 8-break row is NOT a budget regression: MAX_REPAIR_STEPS = 8 caps achievable breaks at 7, and the control refuses an 8-break document at any size, including 840 bytes. Past the boundary the user still gets reveal and regenerate, which is the module's documented posture.

### F2, EEXIST no longer destroys a file the module did not create

`:773` branches on EEXIST before rollback and returns a new `restore-conflict` reason that always carries the backup path. Nothing is unlinked. Real-filesystem probe with a healthy credential-bearing file created inside the window:

    before  reason=write-failed      externalHealthySurvives=false configIsOldCorrupt=true  backups=0 reportsBackupPath=false
    after   reason=restore-conflict  externalHealthySurvives=true  configIsOldCorrupt=false backups=1 reportsBackupPath=true

Zero unlinks in the operation journal.

### F3, the renderer now names the backup on every failure that has one

`EngineConfigRecoveryPanel.tsx:70-105` reads `backupPath` for all branches, not just the ok branch. Before, `mentionsBackup` was false on both halves. After, both write-failed-with-backup and restore-conflict render the new string plus the filename, and a failure with no backup still uses the plain line. Two new keys in all 12 locales with native translations, `i18n-keys.d.ts` regenerated.

### F4, a symlinked config is refused rather than silently converted

New `isRegularFile` dependency using lstat rather than stat, called from both write paths.

    before  ok=true  configStillSymlink=false backupIsSymlink=true targetStillOriginalBroken=true
    after   reason=not-a-regular-file  configStillSymlink=true, no backup taken

A plain broken file still repairs, as a known-positive control.

Deliberate scope limit: inspect is unchanged, so the panel still offers Fix on a symlinked config and clicking it now refuses cleanly and points at Reveal. Surfacing it in the inspection would need a new inspection status and more renderer surface than a LOW warrants.

### Test integrity, including one of our own tests that was vacuous

Six permanent tests added, all mutation-checked. With the four fixes reverted, five fail and the F1 one PASSED. The first F1 fixture used keys of the form k<letter><index>, and the digits added candidate offsets taking it to 150, over the old 128 cap, so the test was exercising nothing. Fixed to fixed-length letter-only keys, 119 candidates verified, and the bound tightened from 3000ms to 500ms. It now fails at 4766ms with the budget disabled versus 71ms with it. The trap is documented in the test so nobody rebuilds it.

### Gates

tsc clean. oxlint on the 4 touched files: 0 errors, 3 warnings all in pre-existing code. oxfmt clean. i18n: 153 passed across the parity and resolution suites plus 808 in the integration suite.

Full unit suite 18,237 passed, 4 failed, 37 skipped across 1,635 files. Of the four: `WorkflowDetailModal.dom.test.tsx` passes in isolation on a quiet machine and was a load phantom, since the suite ran at load 29 to 125 with around 58 competing vitest processes from other lanes. The other three are `acpAgentManagerNpmFallback.test.ts`, tracked as #1030 and not from this diff, and there is now a sharper diagnosis of it in that issue.

### Not closed

The informational fsync item is untouched, as scoped. On the orphaned-backup question: closing F2 and F3 does NOT make it cheap. Nothing in src knows the `config.toml.backup-*` naming, so noticing an orphan needs a directory scan plus a new inspection status plus a new UI affordance. That is a separate lane, not a rider on this PR.

A focused re-audit of this delta follows before merge.

FerroxLabs:
## Delta re-audit addressed at ec70c9b32. HIGH closed, MEDIUM fixed, one false comment deleted, one capability claim corrected.

### The HIGH: a THIRD state where config.toml is gone and the UI said nothing changed

The re-audit was right, and the comment I wrote in the previous round was false. It asserted that the rollback-failure and restore-conflict branches were the ONLY reported states in which config.toml may not hold the user's original bytes. There is a third. In createVerifiedBackup the rename onto the backup succeeds, so config.toml is gone, and then either the readback throws or the byte-equality check fails. Both restoring renames were catch-and-ignore, so when that rename failed too the error carried no path, both callers returned backup-failed with no backupPath, and the panel rendered "could not create a verified backup, so nothing was changed" over a config that no longer existed.

Reproduced independently on a real filesystem before fixing, both verification branches and both write paths:

    ### readback-EIO / repair
      result            = {"ok":false,"reason":"backup-failed","detail":"...backup could not be read back: EIO..."}
      result.backupPath = ABSENT
      config.toml exists= false
      dir               = ["config.toml.backup-20260817-215959"]
      backup holds the user's ORIGINAL credential bytes = true

    ### readback-EIO / regenerate     identical
    ### toctou-mismatch / repair      identical
    ### toctou-mismatch / regenerate  identical
    ### known-positive / repair       result = {"ok":true,...}   <- harness not inert

The byte-mismatch case is a REAL race in the harness rather than a stub: the config is appended to between the read and the rename, from inside the placeholder create that immediately precedes it. Only the restoring rename is forced to fail.

After the fix, all four carry the path and the known positive still returns ok:true:

    result.backupPath = PRESENT   config.toml exists = false   backup holds the ORIGINAL credential bytes = true

The change, entirely inside the two files already touched:

- engineConfigRecovery.ts:500 — EngineConfigBackupError gains an optional backupPath, set ONLY when the move already happened and could not be undone.
- engineConfigRecovery.ts:577 and :585 — the two catch-and-ignore rollbacks become undoBackupMove (:612), which reports whether the undo actually worked.
- engineConfigRecovery.ts:849 and :953 — both callers spread the path onto their failure via backupPathOf (:743).
- EngineConfigRecoveryPanel.tsx:83 — the backup-failed branch gains && !backupPath. This is load-bearing rather than cosmetic: that branch is tested BEFORE the backupPath catch-all, so reordering alone would not have fixed it.

No new locale string was needed. writeFailedWithBackup already reads correctly for this state, so there is no 12-locale churn in this round at all.

**The false comment is deleted, not edited around.** The phrase "ONLY reported" now appears zero times in src, and the same grep finds it once at 3a117cb5a, which is the known positive that makes the zero mean something. What replaced it states three states explicitly and does not half-assert the old claim.

**Mutation results, six mutations, all red:** dropping the field on the error class; dropping either caller's spread, separately; making undoBackupMove report success unconditionally, which is the old semantics; the OVER-FIX that names a backup even when the restore SUCCEEDED; and dropping the panel's guard. The over-fix reddening is what proves the negative controls are not vacuous.

### The MEDIUM: hardlinks are FIXED, not deferred

Reproduced first:

    D2 before: nlink=2 sameInode=true
    D2 repair = {"ok":true}
    D2 config.toml AFTER = repaired
    D2 dotfile   AFTER   = still the original broken bytes
    D2 inodes now differ = true
    D2 backup nlink      = 2

Fixed at engineConfigRecovery.ts:654. Reasoning, decided by execution rather than preference:

- A hardlink is the ONLY irregular kind besides a symlink that can actually REACH this refusal. Executed on each: a directory inspects as unreadable, a device (/dev/null) inspects as ok, and a FIFO BLOCKS in the read and never returns, so none of them is ever offered an action. A hardlink inspects as invalid with BOTH the Fix and the Start over buttons live. That is what moved this off "cosmetic inconsistency" for me.
- The guard's own doc comment justified the symlink refusal with a rationale that applies to hardlinks verbatim. Shipping that comment while hardlinks walk past it is the same false-load-bearing-comment defect as the HIGH above, in the same function.
- It costs one line, needs no new locale string (a hardlink IS a link to another file), and the refusal path with its message, its no-placeholder ordering and its Reveal affordance already exists and is tested.

Against Gemini's position that breaking the link is correct atomic-save behaviour: that is true of a text editor, but this module already decided the opposite for symlinks and documented why. Consistency with our own stated rationale wins over matching vim.

Written as !(nlink > 1) rather than === 1 deliberately, so only a PROVEN multiple-link count refuses and a platform reporting 0 or an unreliable count falls back to today's behaviour instead of refusing every config. Stated plainly: that specific choice is NOT test-covered, because no local filesystem produces a 0 count, and I did not fabricate a test that would pretend otherwise.

Known positive confirmed: a plain single-link config measures nlink=1 and still repairs, asserted in the same test. Mutation-covered both directions: dropping the clause reds the hardlink case, and widening it to refuse every regular file reds 16 tests.

### The LOW: the wording stays, and this is an evidence-based decision

Every locale's notARegularFile says the file "is a link to another file", and the guard also fires on directories, FIFOs, sockets and devices. Executed reachability decides it: the only two states that can actually reach that string are a symlink and a hardlink, and both of them ARE a link to another file, so the sentence is accurate for every state a user can reach. The kinds it would misdescribe are unreachable, as measured above. The actionable half of the sentence is correct in every case regardless. Not worth touching 12 genuine native translations for a case no user meets, so no change and no issue.

### Follow-up filed

#1052 for the parse residual, with the measured numbers, labelled enhancement and area:desktop. It also records that a FIFO config.toml blocks inspectEngineConfig indefinitely, since that is the same root cause and the same fix.

### Gates, all re-run at ec70c9b32 with a genuinely clean tree

Three probe files from this round, one of them a .ts inside src/, were removed before any of these numbers were taken, and git status --untracked-files=all is empty.

- tsc --noEmit: clean, exit 0. Verified non-vacuous by breaking a parameter type in the touched file and watching it report 4 errors.
- oxlint on the 4 touched files: **4 files inspected**, 0 errors, 3 warnings, all three identical at 3a117cb5a so the delta adds none. Whole repo for reference: 3912 files, 0 errors.
- oxfmt --check: 4 files, all correctly formatted.
- i18n: no locale or i18n file is in this delta at all, and re-running the generator is a byte-for-byte no-op.
- Full vitest run pinned to this worktree with --root: **18,247 passed, 3 failed, 37 skipped across 1,647 files**, load 14.5 at start. The 3 failures are all in acpAgentManagerNpmFallback.test.ts, tracked as #1030, not from this diff. The WorkflowDetailModal.dom.test.tsx load phantom from the previous round did not recur.

One process note worth recording, because it nearly produced a false number. A backgrounded vitest run was rewritten and executed with **another lane's worktree as its root** (audit-1042b), which would have made the suite result meaningless and could have disturbed that lane. It was killed within a minute. Re-run in the foreground with an explicit --root, the root resolves correctly, and every targeted run in this round had already reported the correct worktree.

FerroxLabs:
## Independent adversarial delta audit — round 3, head 197138bd9

**Verdict: FIX-FIRST, resolved to GO for this release.** One finding, and it turned out to be already tracked.

### The round-2 HIGH is CLOSED, proven by execution

restoreOriginal (engineConfigRecovery.ts:717) now classifies the occupant before every rollback. A foreign credential-bearing config.toml was planted inside all four windows on a real filesystem:

    B1 backup-readback-EIO   foreign survived, reason restore-conflict, backup named
    B2 backup byte-mismatch  foreign survived, reason restore-conflict, backup named
    B3 regenerate mismatch   foreign survived, reason restore-conflict, backup named
    B4 write EEXIST          foreign survived, reason restore-conflict, backup named

Both files kept in every case. The panel no longer claims nothing was changed: describe() is now backupPath-driven.

### The ownership rule holds on every restore path

There is no whole-file-snapshot restore anywhere in the feature. restoreOriginal is the only restore and all four call sites route through classifyRestoreTarget, which byte-compares. The honesty invariant held across reserve-EACCES, rename-EXDEV, rollback-rename-EIO and verify-EIO.

### Mutation testing: 12 of 14 killed

90 tests pass clean; every mutant verified landed with diff. An over-fix mutant that classified everything as foreign was also killed, so the guard is not one-directional. The two survivors are untested guards that are correctness-preserving today (an accepted-length invariant and a multiline-string state check); both only ever produce conservative refusals.

### The one finding, and why it is not a blocker

The parse cost bound reaches the repair planner but not inspectEngineConfig (:958) or the regenerate health guard (:1081). Measured 117 ms at 1 MB, 1815 ms at 4 MB, 2956 ms at 5 MB, and the panel re-inspects after every action.

This is already filed as #1052, which states in its own body that inspectEngineConfig is unbounded by construction and explains why it was deferred. The audit adds new measurements and one argument #1052 did not cover, namely that plain growth-by-append is a far more plausible corruption mode than the exotic TOML shape #1052 reasons about. All of that is now recorded on #1052.

Deliberately NOT fixed in this release: the obvious pre-filter would report a large but healthy config as invalid, which is a destructive false alarm inviting the user to regenerate a working file holding their API keys. An honest fix needs a distinct status plus all 12 locales, which is not a same-day change.

### Also confirmed clean, by execution

Windows-red test file fixed (POSIX literals folded at the fakeFs seam, chmod test skipIf win32, 90/90 pass). No secret leakage across 9 hostile configs carrying a canary in keys, values, table names, unterminated strings, bad escapes and duplicate keys. The bridge denies the whole namespace by prefix, not by four exact keys. Path confinement fails closed on ProfileIsolationError. All 59 i18n keys present in all 12 locales.

Not verified: Windows execution this session, real kernel-timing races (all windows were forced deterministically through the injected deps seam), and packaged Electron behaviour.

FerroxLabs:

PR #1032, opened 2026-08-17
Mask prefixed secret labels, npm and Hugging Face tokens (#1026)

### LINKED ISSUE #1026: Secret scrubber misses prefixed labels: ANTHROPIC_API_KEY= survives redaction

FerroxLabs:
Fixes #1026.

## The defect

`LABELLED_SECRET_ASSIGNMENT` anchored a leading word boundary before the label. There is no word boundary between an underscore and a letter, because both are word characters, so the PREFIXED form of every label never matched, and environment variables are conventionally prefixed, so the shape that escaped is the shape that actually appears in agent stderr and error text.

Verified by execution on the parent commit (29b71da20), not by reading:

| input | on main | after |
| --- | --- | --- |
| ANTHROPIC_API_KEY=(no-prefix value) | SURVIVES | masked |
| OPENAI_API_KEY=(no-prefix value) | SURVIVES | masked |
| my_api_key=(value) | SURVIVES | masked |
| AZURE_OPENAI_API_KEY=(32 hex) | SURVIVES | masked |
| AWS_SECRET_ACCESS_KEY=(40 chars) | SURVIVES | masked |
| npm_ token | SURVIVES | masked |
| hf_ token | SURVIVES | masked |
| API_KEY=(value) | masked | masked |

This is the mirror image of the four escapes fixed in #1004: there a TRAILING boundary could not match because the character class omitted a word character; here a LEADING boundary could not match because the preceding character was a word character.

## The pattern change, and why it is safe

The leading anchor is now a negative lookbehind for a single alphanumeric character. It still refuses to match a label buried inside a longer alphanumeric run (`notmyapikey=` is not a label, pinned by a test) but permits the `_`, `-` and `.` that separate a prefix from a label.

It is a strict widening, not a rewrite of the rule. Every position where the old boundary succeeded had a non-word character before it, and a non-word character is also a non-alphanumeric one, so no shape that was masked before can stop being masked. Proved rather than argued: a differential run of main's pattern set against this one over a prefix/label/separator/value grid reports **0 weakenings**, re-confirmed independently at 4992 cases with 3768 newly masked and 1224 unchanged.

No backtracking risk: the lookbehind is single-character, fixed-width, and sits outside the quantified section.

## Second change: masking is one pass over the pristine text

Sequential per-pattern `replace` made the result order-dependent in a way that leaked. A narrow rule firing first injects the `[` and `]` of the marker, those characters sit outside the wider rules' character classes, so the wider match aborts and the REST of the credential survives. This is a whole class, not one bug: ten pattern pairs had it. `maskPatternMatches` now collects spans against the pristine text, merges touching spans, and replaces once.

Measured on 506 container/token combinations carrying 682 sensitive segments: sequential masking leaked 142, this leaks 10, and all 10 are the generator's own bookkeeping that sequential masking also left.

**Weakenings, stated exactly.** An earlier version of this PR and of the doc comment claimed zero weakenings. That was FALSE and is corrected in this branch.

- For any input where two credentials are SEPARATED by anything at all (space, comma, quote, ampersand, newline, pipe, semicolon) there are **zero weakenings**: 447216 segment checks over every ordered triple of the pattern set under seven separators, against both sequential masking and `main`, found none.
- Two credentials concatenated with **no separator** can lose coverage that sequential masking had by ACCIDENT. Injecting `[redacted]` put a non-word character into the text, and a later boundary-anchored rule then matched a boundary that does not exist in the pristine string. Only the two fixed-length rules (`AKIA`/`ASIA`, `AIza`) can stop mid-run and supply it; every other class is greedy over `[A-Za-z0-9]`. The rules that lose out are the later boundary-anchored ones: `AIza`, the JWT, `Authorization:` and the Slack webhook.
- Scale of it: 9 ordered pairs and 295 ordered triples of the pattern set, against 1442 triples this masks and sequential masking leaked. Even the no-separator shape is a large net improvement (triple-corpus leaks go from 17993 to 16846).
- `AKIAIOSFODNN7EXAMPLE` immediately followed by a JWT is now **pinned as a test**, with a space-separated control alongside it, so the boundary is recorded rather than assumed. If anyone closes it, the test fails and the doc comment has to come with it.

Merging is provably neutral for over-redaction: a merged span is the union of matched spans and touching means `end == start`, so every character in the union is inside some match. 50000 innocuous lines and 3.4M characters changed nothing, 20 Unicode probes were byte-identical to main, and output is idempotent and deterministic across 300 random permutations of the pattern array.

Span merging closes the ordering question for `SECRET_PATTERNS` **only**. The two rules that run after it, over its output, are still sequential and still lose an anchor: `1//<refresh-token>postgres://user:pw@host` leaks `pw`, because the greedy `1//` span swallows the `postgres` scheme that `URL_USERINFO_PASSWORD` needs. Sequential masking leaked it too, so it is not a regression. Pinned as a test, and left to #1037, which owns the anchor rules.

A three-line fixed point (loop `maskPatternMatches` until stable) was built and measured, and is deliberately **not** included. It closes every no-separator weakening and cuts triple-corpus leaks to 15369 with zero new over-redaction, but it reintroduces exactly the marker dependence this change exists to remove, it does not close the pass-2 anchor loss, and every shape it recovers requires two distinct high-entropy credentials glued together with no delimiter, which is not a shape subprocess stderr produces. Cost was free on clean lines and 2.0x on a line that carries a secret.

## Also folded in

- **AWS secret access key**, via a `secret[_-]?access[_-]?key` label. The 40-character value is caught by the LABEL and deliberately NOT by its shape, because 40 characters of the base64 alphabet is also every git SHA and sha1 digest in a build log, and masking those would make the logs this module exists to keep readable useless. A 40-char SHA line is pinned in the clean corpus to hold that line.
- **npm and Hugging Face prefixes**, with the same `{20,}` floor over a class that omits `_` as the sibling vendor prefixes. That floor is what keeps them off `npm_config_cache`, `npm_lifecycle_script` and `hf_hub_download`, none of which have a 20-character unbroken alphanumeric run.

Noted and deliberately NOT done, to keep the diff to the reported defect: a bare `SECRET_KEY=` label (Django, MinIO, S3-compatible providers) is still unmatched. Say the word and it is one line.

## The missing coverage class

The old suite exercised the bare form of two labels, which is the one form the broken anchor could still match, and that is why this shipped. The label list is now an exported array rather than an inline alternation so the suite can ENUMERATE it: 8 labels, expanded into every spelling the pattern accepts, across 6 prefixes and both casings, is **312 prefixed assignments** asserted masked. A label added later is held to the prefixed form without anyone remembering to come back. A label written in regex syntax the expander cannot read fails the test rather than being silently skipped.

## Known positives, not vacuous negatives

Every negative assertion is paired with a control. The first draft of the ANTHROPIC_API_KEY case used an `sk-` value and PASSED on unfixed main: the prefix rule caught it and the label rule never ran, so it proved nothing about the defect. The committed values carry no recognizable prefix, so only the label can catch them. The sweep additionally asserts the input really contains the value, that `[redacted]` really appears in the output, and that an unlabelled variable of the same shape (`ANTHROPIC_HOSTNAME=`) still passes through untouched. If that ever starts being masked, the sweep has stopped testing labels.

## The window oracle's floor is now measured, not picked

`not.toContain(secret)` plus `toContain('[redacted]')` is satisfied by a partial replacement, and a partial replacement is exactly how a credential leaks, so the suite also asserts that no short window of any corpus secret survives. That window length is the oracle's floor, and it was undocumented at 8.

Mutation-tested: with a tail-leak injected into the masker, 24 tests fail when the surviving tail is at least WINDOW characters, and only the 3 whole-string `toBe` composition tests fail when it is shorter. At WINDOW = 8 a 7-character survival of a corpus secret passed green, and the two shortest corpus secrets are 10 and 11 characters. WINDOW is now **4**, which catches tail leaks of 4, 5 and 7 characters (24 tests each) and leaves the suite green. 3 is too low: `cre` and `npm` collide with ordinary words in the surrounding log text.

## Over-matching

Realistic innocuous log lines (npm env vars, `hf_hub_download`, `passwordless`, `reset_password_token_expired`, a 40-char commit SHA, a sha256 digest, a DSN with no password, prose mentioning `ANTHROPIC_API_KEY` with no assignment): **0 changed**. Six are pinned in `CLEAN_CORPUS`, which asserts byte-identical passthrough.

One accepted readability cost the earlier corpus could not see, because it had no Authorization-header-carrying-a-vendor-token line:

| input | on main | after |
| --- | --- | --- |
| `Authorization: ghp_<token> failed for repo owner/name` | `Authorization: [redacted] failed for repo owner/name` | `[redacted] failed for repo owner/name` |

The header name is consumed. This is a readability cost, not a leak. The `Authorization:` rule already eats its own header name whenever no narrower rule pre-empts it (true on main for a header carrying a value with no vendor prefix), and merging makes it fire in one more case. The `Bearer` form, which is what most real headers look like, keeps the header name. Both are pinned as tests.

## Verification

- `tests/unit/secretRedaction.test.ts` and `tests/unit/secretRedaction.superset.test.ts`: 100 passed, **no existing assertion relaxed, skipped or deleted**.
- Fail-before on the new pins: reverting the module to sequential masking fails 5 tests, including both new `toBe` pins. Applying the rejected fixed point fails exactly 1, the AWS-key-plus-JWT boundary pin, and nothing else.
- Fail-before on the original defect: 8 tests fail on the parent commit; with only the leading anchor reverted, the 312-case sweep fails and lists exactly the underscore-prefixed spellings.
- `tsc --noEmit`: clean.
- `oxlint` and `oxfmt --check` clean on the touched files.

`bridgeAllowlist.ts` is byte-identical to main (empty `git diff`).

Scope note: this touches only `src/process/utils/secretRedaction.ts` plus its tests and fixtures. No Doctor check files are touched, so PR #1029's finding F2 (`backendChecks.ts:52` is scrub-only) is unblocked by this landing.

### DIFF
--- a/src/process/utils/secretRedaction.ts
+++ b/src/process/utils/secretRedaction.ts
@@ -89,6 +89,13 @@ const SECRET_PATTERNS: RegExp[] = [
   /\bgithub_pat_[A-Za-z0-9_]{20,}/g, // GitHub fine-grained PAT
   /\bglpat-[A-Za-z0-9_-]{8,}/g, // GitLab PAT
   /\bgsk_[A-Za-z0-9]{20,}/g, // Groq
+  // npm automation/publish token and Hugging Face user access token. Both carry
+  // an underscore INSIDE the prefix, and the `{20,}` floor over a class that
+  // omits `_` is what keeps them off ordinary npm env vars: `npm_config_cache`
+  // and `npm_lifecycle_script` have no 20-character unbroken alphanumeric run,
+  // so they cannot reach the floor. No trailing `\b`, per the rule above.
+  /\bnpm_[A-Za-z0-9]{20,}/g, // npm access token
+  /\bhf_[A-Za-z0-9]{20,}/g, // Hugging Face access token
   /\br8_[A-Za-z0-9]{20,}/g, // Replicate
   /\bdop_v1_[A-Za-z0-9]{20,}/g, // DigitalOcean
   /\bya29\.[A-Za-z0-9_.-]{8,}/g, // Google OAuth access token
@@ -126,9 +133,56 @@ const SECRET_PATTERNS: RegExp[] = [
  * {@link SECRET_PATTERNS} rather than indexed inside it: an index-based special
  * case silently mis-applies itself the moment somebody inserts a pattern above
  * it, which it did on the first attempt here.
+ *
+ * The labels are a list rather than an inline alternation, and exported, so the
+ * suite can ENUMERATE them instead of spot-checking a few. #1026 survived
+ * because the tests covered the bare form of two labels; a new label added here
+ * is now automatically held to the prefixed form too.
  */
-const LABELLED_SECRET_ASSIGNMENT =
-  /\b(api[_-]?key|auth[_-]?token|access[_-]?token|refresh[_-]?token|client[_-]?secret|password|passwd)(\s*[:=]\s*)["']?[^\s"',}]{8,}["']?/gi;
+export const LABELLED_SECRET_LABELS: readonly string[] = [
+  'api[_-]?key',
+  'auth[_-]?token',
+  'access[_-]?token',
+  'refresh[_-]?token',
+  // `AWS_SECRET_ACCESS_KEY` (#1026). Its 40-character value has no prefix and no
+  // shape worth keying on - 40 characters of the base64 alphabet is also every
+  // git SHA and every sha1 digest in a build log - so the label is the ONLY
+  // high-confidence signal, and until #1026 the label could not be reached.
+  // Listed before `access[_-]?token` would matter only for a shared prefix; it
+  // does not share one, and alternation backtracks through every branch anyway.
+  'secret[_-]?access[_-]?key',
+  'client[_-]?secret',
+  'password',
+  'passwd',
+];
+
+/**
+ * #1026: the leading anchor is a NEGATIVE LOOKBEHIND, not `\b`.
+ *
+ * `\b` here was unreachable for the common case and that was a live leak. There
+ * is no word boundary between `_` and a letter because both are word characters,
+ * so `ANTHROPIC_API_KEY=...`, `AZURE_OPENAI_API_KEY=...` and `my_api_key=...` all
+ * failed to match while the bare `API_KEY=...` matched. Environment variables are
+ * conventionally PREFIXED, so the shape that escaped was the shape that actually
+ * appears in agent stderr and error text.
+ *
+ * This is the mirror image of the four escapes fixed in #1004: there a TRAILING
+ * boundary could not match because the character class omitted a word character;
+ * here a LEADING boundary could not match because the preceding character was a
+ * word character.
+ *
+ * `(?<![A-Za-z0-9])` still refuses to match inside a longer alphanumeric run, so
+ * `notmyapikey=` is not a label, while permitting the `_`, `-` and `.` that
+ * actually separate a prefix from a label. It is a strict WIDENING: every
+ * position where `\b` succeeded had a non-word character before it, and a
+ * non-word character is also a non-alphanumeric one, so nothing that was masked
+ * before can stop being masked. Single-character, fixed-width and outside the
+ * quantified section, so it adds no backtracking.
+ */
+const LABELLED_SECRET_ASSIGNMENT = new RegExp(
+  `(?<![A-Za-z0-9])(${LABELLED_SECRET_LABELS.join('|')})(\\s*[:=]\\s*)["']?[^\\s"',}]{8,}["']?`,
+  'gi'
+);
 
 /**
  * `scheme://user:PASSWORD@host` - the password segment of a URL or DSN. No
--- a/src/process/utils/secretRedaction.ts
+++ b/src/process/utils/secretRedaction.ts
@@ -154,6 +154,12 @@ export const LABELLED_SECRET_LABELS: readonly string[] = [
   'client[_-]?secret',
   'password',
   'passwd',
+  // #1042 routed this here rather than editing its own branch, because that PR
+  // newly puts a backup passphrase on the IPC payload and this array is the only
+  // backstop for it. `password` and `passwd` already matched; `passphrase` did
+  // not, so `passphrase=`, `pass_phrase:` and a JSON `"passphrase"` all survived
+  // verbatim. Verified by execution with `password=` as the known positive.
+  'pass[_-]?phrase',
 ];
 
 /**
--- a/src/process/utils/secretRedaction.ts
+++ b/src/process/utils/secretRedaction.ts
@@ -171,6 +171,10 @@ export const LABELLED_SECRET_LABELS: readonly string[] = [
  * here a LEADING boundary could not match because the preceding character was a
  * word character.
  *
+ * Not a novel construct here: `@common/utils/redactCommandSecrets` already used
+ * `(?<![A-Za-z0-9])` for this exact reason, with a comment saying so. The sibling
+ * bank had the rule right and the canonical one did not.
+ *
  * `(?<![A-Za-z0-9])` still refuses to match inside a longer alphanumeric run, so
  * `notmyapikey=` is not a label, while permitting the `_`, `-` and `.` that
  * actually separate a prefix from a label. It is a strict WIDENING: every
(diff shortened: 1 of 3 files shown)


FerroxLabs:
## FIX-FIRST addressed in 740b7a007

### FIX 1 — reproduced, then fixed wider than prescribed

Reproduced the interaction leak myself before touching anything. My own measurement over **506 container/token combinations carrying 682 sensitive segments**:

| variant | leaked segments | weaker than main | better than main |
| --- | --- | --- | --- |
| main | 142 | 0 | 0 |
| head as pushed (b269f3ffd) | 130 | **12** | 24 |
| audit's fix: move the 2 entries to the end | 118 | 0 | 24 |
| **shipped: one-pass span merge** | **10** | **0** | **132** |

The audit's reorder works — 16 to 0 in its count, 12 to 0 in mine. I did not ship it, because it fixes 2 of 12 instances of the same class and the next appended pattern re-opens it.

Shipped instead: `maskPatternMatches` collects every pattern's match spans against the **pristine** text, merges overlapping or touching spans, and replaces once. Order stops mattering, so the class is closed for all twelve rules rather than two.

All 10 remaining leaks in that corpus are the generator's own bookkeeping — a public JWT header, and `ya29.`/`1//` values that cannot form a real webhook path — and **main leaks the same 10**. Corroborating checks: 0 weakenings across the 4200-case labelled-assignment corpus; 0 of 28 innocuous log lines changed, identical to main; 0 cases where sequential masking removed something the one-pass version does not (so no `[redacted]` marker acts as glue). Performance 664KB in 9.1ms (main 3.5ms), and a 200,000-character unterminated `Bearer` run in 1.8ms — still linear.

**The ten pre-existing rules were not moved, deliberately.** Under a span merge their position is irrelevant, so reordering them would be a no-op diff that also separates each rule from its vendor siblings. One correction to the audit's framing worth recording: I expected `Bearer npm_<token>` to REGRESS under the reorder, since the Bearer rule runs at index 1 and would truncate at the underscore. It does not — `[A-Za-z0-9._\-+/=]` contains `_` (the `._` is dot-then-underscore). Measured, not reasoned.

### FIX 2 — the oracle

Correct diagnosis: `not.toContain(secret)` plus `toContain('[redacted]')` is satisfied by a partial replacement, which is why the suite was green while the leak was live. Two additions, no existing assertion touched:

- **Three composition cases** — Slack webhook, JWT, and `Authorization:` header, each carrying a synthetic `npm_`-prefixed inner segment and each a credential *in whole* — asserted with `toBe('[redacted]')`. Fail-before, with only the sequential loop restored:
  ```
  FAIL  masks the whole Slack incoming-webhook URL whose middle path segment is a vendor token
  FAIL  masks the whole JWT whose payload segment is a vendor token
  FAIL  masks the whole Authorization header carrying a vendor token with no scheme
        Tests  3 failed | 93 passed (96)
  ```
  Plus a control that each container really contains the inner token and is materially longer than it, so `toBe` cannot pass vacuously.
- **A windowed oracle over the whole corpus**, which is the general form of your suggestion: no 8-character window of any corpus secret may survive. It sees a partial mask anywhere in the corpus, not only where the whole input is the secret. Guarded by a test asserting every corpus secret is at least 8 characters, so the window cannot silently skip everything.

Both scrubber suites: **96 passed**, up from 61, with the only deletion in `tests/` still the one extended import line.

### Correction to my own report

You are right and I was wrong on the compatibility premise. Four lookbehinds already existed in `src/` — my grep had failed with a zsh glob error and I read "no matches" as a zero instead of as a broken method. `@common/utils/redactCommandSecrets.ts:51` already used `(?<![A-Za-z0-9])` for *exactly* the #1026 reason, with a comment saying so: the sibling bank had the rule right and the canonical one did not. That is the real justification and it is now recorded in the module. `tsconfig target: ES6` was irrelevant twice over, as you say.

### Not pulled in

The mirror half — the trailing boundary, `ANTHROPIC_API_KEY_BACKUP=`, `SECRET_KEY=`, `{"api_key":"..."}` and the rest — stays out of this PR for its own issue. Withdrawing my "one line" characterisation: `redactCommandSecrets.ts:67` shows the shape that actually covers it (`(?:[_-][a-z0-9]+)*` plus a trailing negative lookahead), and that is a rule change with its own over-matching surface, not a label addition.

### Verification

`tsc --noEmit` clean. Full suite **18099 passed, 3 failed**, all 3 the machine-local `acpAgentManagerNpmFallback.test.ts` PATH failures (#1030). An earlier run of the same tree reported a 4th failure that did not reproduce and left no FAIL line — load average was 23-35 at the time versus 9.7 for the clean run, so it was a phantom. `oxlint` clean; `oxfmt` on the two touched files only. `bridgeAllowlist.ts` diff versus the base is still empty. Base note: `main` has since advanced to 9a454acd; this branch is unrebased at merge-base 29b71da20.

FerroxLabs:
## Independent adversarial audit — #1032 plus the stacked scrubber branch (745d291e2)

**Verdict: FIX-FIRST.** One HIGH, three MEDIUM, and one pre-existing issue filed separately. The rest of the work verified clean by execution.

### Verified clean, executed rather than read

    #1026 prefixed labels, npm and HF tokens   all masked
    #1051 JSON quoted-key form                 all 12 labels masked, plus spaced and single-quote forms
    #1037 suffixed labels                      SECRET_KEY=, GITHUB_TOKEN=, API_KEY_PROD=, bare token= all masked
    ordering, stacked fix vs #1032             differential over 45,000 labelled-assignment cases:
                                               weakened 0, widened 26,400. SECRET_PATTERNS: 120 cases, 0 differing
    catastrophic backtracking                  14 adversarial families at n=1000 to 16000, all LINEAR.
                                               worst case 4.5 ms at 78 KB
    windows shard hazard                       clean, the only path comparison is separator-normalised

Mutation testing: 13 designed independently, 13 killed, including 5 over-fix mutants. Every mutation verified landed with diff and restored byte-identical. Baseline 134 passing before and after.

### HIGH — a multi-word passphrase is not masked

The value class excludes whitespace, so:

    {"passphrase":"correct horse battery staple"}      UNCHANGED, fully unmasked
    passphrase=correct horse battery staple            UNCHANGED, fully unmasked
    {"passphrase":"watermelon sugar high tonight"}     -> "[redacted] sugar high tonight"
    BACKUP_PASSPHRASE="watermelon sugar high tonight"  -> "[redacted] sugar high tonight"

The partial case is worse than the plain leak: it looks redacted, so a human reviewing a feedback bundle sees nothing wrong while most of the secret ships.

This is reachable, not hypothetical. The file itself records that #1042 routed the passphrase label into this array because this array is its only backstop. The product's own sync passphrase input has a 16-character minimum and NO character restriction, so a four-word passphrase passes validation and is the expected shape. The path to disk is the console log transport, and the feedback bundle re-scrubs the same text.

### MEDIUM — the KNOWN GAPS entry is worse than documented, and its pinned test covers only the benign half

The comment says a JSON value containing a space or a closing brace falls under the 8-character floor. That holds only when the FIRST run is under 8 characters, which is exactly what the pinned test uses. With a longer first run:

    {"api_key":"SUPERSECRETPART1 SUPERSECRETPART2"} -> {"api_key":"[redacted] SUPERSECRETPART2"}

So the real gap is a partial mask presenting as complete, not an absent match.

### MEDIUM — masked JSON does not stay valid for unquoted values

The property holds for quoted values only. Epoch timestamps are not covered by the count-magnitude argument, since epoch seconds is 10 digits and epoch millis is 13:

    {"token_expires_at":1755424800} -> {"token_expires_at":[redacted]}   invalid JSON

### MEDIUM — over-redaction destroys two fields this app reads by name

Neither is in the accepted-false-positive inventory:

    token_endpoint   a PUBLIC OAuth discovery URL, read at xaiOAuthCore.ts:109. On a discovery
                     failure the one field you need is masked while jwks_uri survives, so the
                     report looks intact
    api_key_env      an environment-variable NAME, never a value, validated at profileStore.ts:118.
                     Masking it makes "which variable did I misspell" unanswerable

Confirmed safe by execution and needing no change: token_family, token_blacklist, total_token_usage, max_tokens, input_tokens, total_tokens.

### Pre-existing, filed as #1065

Per-line scrubbing on the engine stderr path destroys the PEM anchor, so the whole-file pass that builds the feedback bundle can no longer match and the private key body ships. The PEM rule arrived in #1004 and no caller changed in this stack, so it is out of scope here.

### Not verified

The larger corpus figures quoted in earlier comments were not reproduced; this differential covers 45,120 cases. Windows was static review only. The auditor designed its own 13 mutants rather than reproducing the previously claimed set.

FerroxLabs:

Issue #1017, opened 2026-08-17
Windows without a bundled Bun falls back to node.exe and can ENOENT


FerroxLabs:
Deferred from the cross-audit of #1015. Pre-existing, NOT introduced by that PR.

## What happens

On a win32-x64 machine without AVX2, prepareBundledBun.js:189-196 excludes win32 from
needsBaselineVariant, so no "win32-x64-baseline" runtime is ever staged. getBundledBunDir()
then returns null and resolveJsRuntime falls back to "node.exe".

If that machine has no Node on PATH, the spawn fails with ENOENT and the server reports
"Enabled but exposes 0 tools" -- the same symptom as #1008, on a different platform.

## Scope

This is not specific to the three core builtins. It has behaved this way for the four
sibling @wayland builtins and for every npx-based MCP server since #438 / #706.

## Why deferred

#1015 fixes the macOS case (no /usr/bin/node) and does not make this worse. Fixing it means
either staging a baseline Bun variant for win32 or defining an explicit, declared fallback --
that is its own change with its own packaging verification, and it needs a real non-AVX2
Windows box to prove.

## Acceptance

A win32-x64 machine without AVX2 and without Node on PATH can call a first-party builtin MCP
server and get a tool result, proven by execution on real hardware rather than by reading.

wayland:
STOP TAGGING ME!

FerroxLabs:
Fixed and shipped in v0.12.14. tests/unit/windowsBaselineBun.test.ts pins that a win32-x64 machine without AVX2 gets a bundled Bun, across the three coupled staging edits.

PR #1042, opened 2026-08-17
Stop the legacy restore reporting success when it applied nothing

### LINKED ISSUE #1021: Bug report: restore fails to bring in anything
### What happened
Export files from windows. include keys.
Import on the previous version of Wayland == nothing happens.
Import on current version:   message appears indicating import successful.  No prompt for keys password.   Restart of Wayland shows no imported keys and no chats and no projects.

### Environment
- App: 0.12.0
- Engine: wayland-core 0.13.0
- OS: linux x64 (7.0.0-28-generic)

### Diagnostics
**MCP servers** (mcp.config): 5
**Providers** (model_registry_providers): 1
**Workspace** (projects + conversations): 2
- `Locating Wayland Configuration Restore Logs` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
- `Locating Wayland Backup Restore Logs` — This chat is using a temporary workspace, so files it writes (e.g. "save to the local workspace") land in a throwaway directory you may not be able to find. Open it inside a project that has a workspace folder, or set one.
**Config paths**
- app: ~/.config/Wayland/config
- engine: ~/.config/wayland-core
**Recent errors** (~/.config/Wayland/logs)
```
2026-08-17.log: [2026-08-17 20:32:43.434] [error] (node:20223) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
2026-08-17.log: (Use `Wayland --trace-deprecation ...` to show where the warning was created)
2026-08-17.log: [2026-08-17 20:32:49.016] [warn]  [ijfw] safeSpawn(npm view) failed {
2026-08-17.log: err: 'Error: Could not resolve trusted npm (platform=linux). Tried:\n' +
2026-08-17.log: [2026-08-17 20:32:49.030] [warn]  [ijfw] safeSpawn(npm view) failed {
2026-08-17.log: err: 'Error: Could not resolve trusted npm (platform=linux). Tried:\n' +
2026-08-17.log: [2026-08-17 20:32:49.031] [error] [ijfw] safeSpawn(npx install) failed {
2026-08-17.log: err: 'Error: Could not resolve trusted npm (platform=linux). Tried:\n' +
```

---

Image

FerroxLabs:
Fixes #1021.

## What the reporter hit

Exported from Windows with "include API keys" checked, imported on Linux, was told the import succeeded, was never asked for the keys passphrase, and after a restart had no keys, no chats and no projects.

## Root cause, per family

All three families fail for the SAME underlying reason, and it is not cross-platform anything. Chats (conversations, messages), projects and provider credentials (model_registry_providers.creds_encrypted) all live in userData/wayland/wayland.db. The legacy file export only ever walks userData/conversations, userData/attachments, userData/config and the optional userData/keys.json, and its own manifest declares desktop.database, projects, providers and the rest as excludedAuthorities. So an archive taken from a modern install carries none of what the user wanted.

On top of that, three reporting defects turned a documented scope limit into silent data loss:

1. The desktop restore never collected a passphrase. handleRestoreClick invoked storage.importBackup with an empty payload, and backupImport skips keys.json.enc outright when opts.passphrase is undefined. The browser path has always had a passphrase field. The desktop path had no dialog at all, so there was nowhere to type one. That is exactly why no prompt appeared.

2. backupImport returned void, so storageIpc returned a flat ok:true whether it installed four directories or zero files. "Did not throw" was reported to the user as "your data moved".

3. backupExport silently recorded includesKeys:false when the caller asked for keys and userData/keys.json did not exist. Nothing in the app writes that legacy file any more, so ticking the box and typing a passphrase produced a keyless archive with no hint.

## Version asymmetry

Same cause. At v0.11.18 the desktop click was fire-and-forget with no toast on either outcome, which is why the previous version "did nothing". #925 added the success toast over an import path that could still apply nothing.

## Refuted by execution

- Cross-platform credential backend (DPAPI vs libsecret) is NOT involved. A Windows exporter emits the same forward-slash entry names as a POSIX one, and a Windows-origin archive imports cleanly on POSIX. There is a test for it now.
- transferImportPlanner.ts is NOT on this path. planTransferImport has zero callers in src; familyDecision and decideConflict never run during a legacy restore.

## The fix

backupImport now returns what it applied, what fell outside the legacy restore scope, and whether encrypted keys were skipped for want of a passphrase. backupExport reports whether keys were requested but absent. Both travel through the existing IPC channels to the renderer, which now reports the applied list, warns plainly when nothing was applied and explains why, and warns when keys were left behind. Desktop restore opens the same confirmation dialog the browser already uses, so the scope warning is read before anything is overwritten and the passphrase can be entered.

No change to what the legacy export covers. No signature, authority or containment check was relaxed. bridgeAllowlist.ts is byte-identical to main.

## Left for separate work

Actually moving chats, projects and provider credentials between installs is the transfer subsystem's job, not this one. This PR makes the legacy path stop lying about its scope; it does not widen that scope.

## Verification

- tests/unit/storage.test.ts: 23 passed, 5 new (nothing-applied on a database-only install, applied-entry reporting, keys-skipped reporting, out-of-scope reporting, Windows-origin archive applies on this platform). No existing assertion weakened.
- tests/unit/StorageBackupExportSuccess.dom.test.tsx: 8 passed. The two desktop restore tests were updated to drive the new confirmation dialog and keep their original assertions; 4 added, including "never reports success when a desktop restore applied nothing".
- Locale key parity: 83 passed, 5 new keys added to all 12 locales.
- tsc --noEmit clean. Targeted slice: 243 passed across 8 files.
- Reproduced and re-verified against the real backupExport/backupImport with a scratch userData, before and after.

## Also found, not fixed here

The same Doctor output shows the Linux variant of the "Could not resolve trusted npm" family. safeSpawn.ts lines 74-78 give POSIX exactly three hardcoded candidates, two of which cannot exist on Linux, and the POSIX branch never reads the PATH argument the win32 branch sweeps. That is untouched by #1008, #1015 and #1017 and needs its own issue. The two builtin MCP servers showing 0 tools are the #1008 mechanism, which open PR #1015 does cover on Linux.


FerroxLabs:
## Cross-audit verdict: NO-GO

Legs: Codex 5.6 Sol NO-GO, completed on the correct PR with no drift and no provider refusal, strongest leg. Gemini 3.1 Pro NO-GO, completed with real file reads but executed nothing, and its line numbers were unreliable so every substantive lead was re-verified independently. Kimi K3 DID NOT RUN, provider 403 usage limit for the billing cycle, the same leg lost on every audit today, so no findings and none invented. Internal execution leg NO-GO.

The reporting logic this PR adds is good and it is genuinely tested. Eight of nine mutations were killed. The direction is right. It cannot ship over an error path that says nothing.

## What is right, so nobody re-litigates it

Desktop cannot report success while applying nothing. The applied list is appended only after a successful rename into the live profile, and every early continue leaves it untouched. Reverting the original #1021 bug, returning the full target list instead of what was applied, fails 4 tests.

A wrong passphrase is not destructive. It aborts during staging before replaceFromStaging, executed: pre-existing config intact, no conversations leaked. And because the PR now sends a passphrase, the pre-restore safety zip finally includes the old encrypted keys file, which it never could before. That is a genuine improvement.

No passphrase leak, established by execution with a credential-shaped value and a byte-level sweep of every file under the restored profile. All 5 locale keys are present in all 12 locales with genuine native prose, and the i18n type generator reports up to date with no diff.

## Blockers

**F1. A mistyped passphrase hangs the panel forever with no message.** backupImport.ts:52 throws on a wrong passphrase, storageIpc.ts:72 has no try/catch, and the bridge cannot carry a rejection. Executed with a known positive first:

    known-positive probe:ok        -> {"settled":"resolved","value":{"ok":true,"value":42}}
    UNHANDLED REJECTION IN PROVIDER: bad decrypt
    throwing provider probe:throws -> {"settled":"PENDING FOREVER (timed out)"}

End to end with a real archive: no toast, modal stays open, Confirm stuck on a loading spinner, and because the finally never runs the card's Restore button spins for the rest of the session even after Cancel. The catch at BackupCard.tsx:93 is dead code.

This hazard is already written down in this repo at src/common/types/speech.ts:119-131 and already solved at speechToTextBridge.ts:44-55. The mechanism pre-exists. What this PR does is supply the trigger: on main the desktop sent an empty payload, so decryptBuffer was unreachable.

**F3. A restore that fully succeeded reports failure.** backupImport.ts:143-145 and :228-230 clean up inside a finally, and a finally that throws replaces the successful return, so the applied report this PR is built around is computed and then discarded. Executed on real macOS with no mocks, using a read-only subdirectory in the live config tree:

    CLEANUP outcome: REJECTED: EACCES: permission denied, unlink '.../config/locked/pin.txt'
    CLEANUP live config now: {"theme":"NEW-FROM-ARCHIVE"}

The restore completed and the call rejected anyway. Combined with F1 the user sees nothing, concludes it failed, and may restore from the safety zip, undoing a good restore. On Windows this is routine through EBUSY or EPERM on open handles.

**F4. Told "nothing was restored" when the archive holds their API keys.** BackupCard.tsx:76-84 tests applied.length === 0 before keysSkippedNoPassphrase, so a keys-only archive with no passphrase falls into restoreNothingApplied, whose text says the archive held no legacy files and that API keys live in the primary database. Every clause is wrong for that archive. The keys are in it, one passphrase away. All three working legs flagged this independently. Same class of harm as #1021 itself.

**F5. The titled bug is still fully live on the WebUI surface.** storageRoutes.ts:155 calls backupImport and discards the ImportReport, replying success unconditionally at :159. A WebUI user restoring a modern-install archive still gets a confident success over a no-op. The PR fixed one of two callers.

## Also to fix

F2, the same hang on Export one click away: tick include-keys, leave the passphrase blank, and backupExport.ts:84 throws through the same no-reject bridge. F6, the desktop restore dialog now shows WebUI-only copy ending "Available only from a trusted local network", wrong in all 12 locales for a desktop user. F7, redactSecrets covers password and passwd but not passphrase, verified with a known positive, which is the missing backstop for the secret this PR newly puts on the IPC payload. F8, outOfScope is plumbed renderer-deep and never read.

## Two test problems

A real bug this PR fixes is completely untested. Mutation M9 removed the result.ok guard at BackupCard.tsx:32 and all 31 tests stayed green. That guard fixes a genuine second defect: on main handleExport had no ok check, so cancelling the native save dialog showed "Legacy file export created" over a file that was never written.

Two pre-existing tests are anchored to a shape production cannot emit. StorageBackupExportSuccess.dom.test.tsx:78 and :115 use a rejected mock, but the real bridge never rejects, which is proven above. Those two tests are exactly why F1 shipped invisible: green against a fiction.

## Filed separately, not gating

backupImport.ts:130-142: if the rollback loop's own remove or rename throws, the displaced restore loop is skipped and the finally then deletes the rollback root, destroying the only copy of the displaced originals. Untouched by this diff and the trigger was not reproduced.

## Suite

No trustworthy full-suite number: the machine was at load 90 to 114 with five lanes running vitest concurrently, the condition that fabricates failures here. In isolation, quiet enough to trust: the two PR-touched files 31 passed 0 failed, and a 70-file storage, backup, export, reset, i18n, redteam and locale slice 1421 passed 1 skipped 0 failed. Nothing in the storage area failed under any condition. An authoritative full-suite number will be recorded here separately.

A fix lane is on all of the above.

FerroxLabs:
## Delta audit of the fix, 237a2d4f7 to 60577fbb0: FIX-FIRST. One line and two tests from GO.

Legs: Codex 5.6 Sol RAN, NO-GO, strongest external leg and it independently found the finding below including the entry arithmetic. Gemini 3.1 Pro DID NOT COMPLETE, no shell, worked from grep and read only, and its one Critical finding is refuted by execution. Kimi K3 DID NOT RUN, provider 403 usage limit, the same leg lost on every audit today. Internal execution leg FIX-FIRST.

## All five blockers are genuinely closed, verified by execution and not by reading

The five outcomes were driven through the real exporter into the real importer, and all five now tell the truth: a foreign archive warns and names what it actually held, settings-only succeeds and lists them, keys with no passphrase warns and says to enter it and retry, a wrong passphrase reports a classified failure with the dialog left open, and full success lists everything applied.

**F1 and F2 hold against the REAL bridge, which the lane's own test does not exercise** because it mocks the provider. The mechanism was first read out of the shipped `@office-ai/platform` bundle: subscribe is `.then(...)` with no `.catch`, and invoke is `new Promise(resolve)` with no reject and no timeout. Then executed on that same bridge:

    KNOWN POSITIVE unwrapped provider   -> PENDING FOREVER
    REAL importBackup, wrong passphrase -> resolved {ok:false,failed:true,errorCode:BAD_PASSPHRASE}
    REAL importBackup, right passphrase -> resolved {applied:["keys.json"]}
    REAL exportAll, cancelled           -> resolved {ok:false}

Escape vectors tried and unreachable: a throw before the try block, in getUserData, in the finally, an async rejection after return, a non-Error value, a Symbol. Every throw site in the chain throws an Error.

**F3 holds against a real filesystem fault, not a mock.** An immutable file pulled into the rollback tree so the removal genuinely throws:

    KNOWN POSITIVE: rmSync throws -> EFAULT
    RESULT: backupImport RESOLVED -> {"applied":["conversations"],...}

**F5 is closed on both surfaces through one reporter.** The route forwards the report, the service types it, and both call sites go through the single `reportRestore`, so they cannot drift apart again.

**Nothing is destructive.** A partial apply followed by a real EPERM left every original byte-identical, with nothing lost. A wrong passphrase aborts in staging before the live tree is touched. And the safety-zip claim holds as a two-way invariant: keys are only ever replaced when a passphrase was supplied, which is exactly when the safety export captures them.

**Nothing leaks**, swept with a credential-shaped passphrase over a credential-bearing archive, with a live known positive proving the sweep works. The IPC payload, the classified reply and the log are all clean. The only userData path that crosses is the safety backup path, which is deliberate and is shown to the user on purpose.

**Locales are clean**: 9 new keys in all 12 with zero English fallbacks outside en-US, genuine native prose spot-read in all 11 non-English locales with correct per-locale terminology, the network clause removed from all 12 with no dangling fragment, and the type generator a byte-for-byte no-op.

**21 mutations run, 19 killed.** tsc clean. oxlint 8 files inspected, 0 errors. Full suite 18,062 passed with 3 failures, all of them the known #1030 host-PATH issue confirmed by its message.

## Finding 1, MEDIUM-HIGH. A hostile archive still freezes the panel forever.

`backupImport.ts:205, 241, 257` collect `outOfScope` with no cap on entry count and no cap on name length, and `common/adapter/main.ts:53` drops any reply over 50 MB.

    hostile archive on disk: 234553 KiB for 1000 entries
    report.outOfScope   = 1000 entries
    serialized reply    = 57.2 MB   (adapter limit 50.0 MB)
    OVER LIMIT -> event DROPPED -> renderer await NEVER SETTLES : true
    known-positive (small reply over limit? must be false): false

Roughly 1000 zero-byte entries under 60 KB-long top-level names. All legal zip, and the zip-bomb byte caps never fire because every entry is empty. Import returns normally, the adapter silently skips a 57 MB reply, and because invoke has no reject and no timeout the Restore button spins for the rest of the session with nothing said. The WebUI half instead ships 57 MB of JSON to a browser.

The unbounded field predates this delta. It belongs here anyway, because this delta is the one that claims the hang is gone and it newly forwards the field over HTTP and newly renders it. The existing sanitiser runs too late, after the payload has crossed the boundary, and it maps the whole array before slicing.

Fix, one line at `:241`: cap the set size and slice the name.

## Finding 2, LOW-MEDIUM. "The dialog stays open on BAD_PASSPHRASE" has no killing test.

`BackupCard.tsx:187`. Replacing the guard with a bare close leaves all 18,062 tests green. That is a listed half of the F1 fix, the part that lets someone retry after a typo without re-entering everything.

## Finding 3, LOW-MEDIUM. The guard that stops #1021 recurring has no killing test.

`BackupCard.tsx:141`. Adding a truthiness check on `report.applied` leaves all 18,062 green, **and that mutation restores verbatim #1021**: a success toast over a restore that moved nothing, for any reply lacking `applied`. That input is reachable today, because `StorageService.ts:130` returns `json.data ?? {}`, so a reply without `data` produces an empty report.

The behaviour on this branch is correct. The problem is that the single deliberate decision protecting the originally reported bug is the one thing no test pins.

## Observation, not ranked

`storageRestoreRoute.test.ts:115` is named "does not claim success when a modern-install archive applied nothing" and asserts `success: true`. The behaviour is right, since success is the transport signal and the user-facing claim is made in `reportRestore` which warns. But a test whose name contradicts its assertion is the same genre of problem this audit exists to catch. Worth renaming.

## Raised by the panel and declined, with reasons

Gemini's Critical claim about logging decrypted passphrases is **refuted by execution**; the log sweep came back clean with a live known positive, and what the log actually carries is the error code plus a stack of source paths. A non-Error with a throwing toString is real in the abstract but unreachable here. The safety backup path crossing the boundary is deliberate. The rollback-loop throw is #1050, explicitly out of scope with its trigger unreproduced. Other storage providers lacking try/catch is real, pre-existing, and outside this delta.

## Bottom line

This is the most thoroughly built delta in the swarm, and the public statement on #1021 is true for every archive a real Wayland install produces. It is not a GO only because a hostile or malformed archive can still recreate the precise frozen-panel symptom this PR is named for, and the fix is one line at the collection site.

A fix lane is on all three.

PR #1046, opened 2026-08-17
Report the newest log files in diagnostics, not the directory's first

### LINKED ISSUE #1038: Diagnostics collector reports the OLDEST log files, not the newest
Found while triaging #1014. This one has been quietly degrading every macOS bug report we receive.

## The defect

src/process/resources/builtinMcp/conciergeDiagServer.ts:881-883 lists the log directory with readdirSync,
applies NO mtime sort, then takes .slice(0, 6). readdirSync returns entries in directory order, so the six
files it picks are effectively the OLDEST, not the newest.

Every "Recent errors" block we have ever received is therefore likely to be showing the wrong logs.

## Confirmed in the wild, twice

- #1014 was filed today against 0.12.0 and its Recent errors block carries four lines from JULY.
- #940 was filed on 6 August and carries JUNE lines.

Both reporters did nothing wrong. The collector handed us stale evidence and labelled it recent.

## Why this matters more than it looks

The whole point of the diagnostics block is that a reporter does not have to know which log matters. We
built it so a non-developer could give us something useful in one click, and instead it has been giving us
log lines from a different app version than the one they are reporting against.

The cost is not just noise, it is misdirection. On #1014 the July lines are 0.11.x era SKILL.md ENOENT
errors that have nothing to do with the reported problem, and it would be very easy to spend an hour
chasing them. Anyone triaging these reports has been reasoning about the wrong data, and there is no way
to tell from the report itself.

## Fix

Sort by mtime descending before slicing, and state the window in the output so it is self-describing, for
example "6 most recent log files, newest first" with the file dates included. Consider slicing by RECENCY
rather than count, since six files could span months on a light user and one afternoon on a heavy one.

## Acceptance

A report filed today shows today's log lines. Prove it with a directory whose creation order and mtime
order disagree, since a naive test on a freshly written fixture will pass either way. That last point is
the whole reason this survived.

FerroxLabs:
Fixes #1038

## What was wrong

`readdirSync` returns entries in DIRECTORY order. It is not sorted and it is not by time. `readRecentErrors` then took the first `MAX_LOG_FILES` (6) entries, which kept an arbitrary subset that in practice skews oldest, and presented them to the model and to us under the heading "recent errors".

That means every triage that trusted a Recent errors block may have been reading the wrong data. This was confirmed twice against real incoming reports.

There is a second layer of the same defect. `collected.slice(-MAX_LOG_LINES)` is commented "keep only the most recent lines", but `collected` was ordered by the file iteration order, so the tail was really "the last lines of whichever file happened to be processed last".

## The fix

Stat the candidates during selection, sort by mtime newest-first, take the newest `MAX_LOG_FILES`, then reverse so the survivors are walked oldest-first. That makes the trailing slice genuinely the most recent output.

Statting at selection also removes a second `statSync` per file and tolerates a log that rotates away between `readdir` and `stat`, which a rotating log directory does routinely.

## Verified

Both regression tests were mutation-tested. Reverting the selection while keeping the tests:

```
x reports the NEWEST log files, not whichever the directory listed first
    AssertionError: expected 'a1.log: error: marker for file 1\na2...' to contain 'marker for file 7'
x orders collected lines oldest-first so a truncated tail keeps the newest
  Tests  2 failed | 40 passed | 9 skipped (51)
```

With the fix in place: 42 passed, 9 skipped.

The first test uses 8 files with explicitly set mtimes, where the two newest are also last alphabetically AND last created, so any implementation taking the first six of either ordering loses them. The second names the newer file to sort first, so directory order alone would leave the newest line at the front where a truncating tail drops it.

## Note on lint, worth reading

`oxlint` does not inspect either changed file, and it will not inspect them in CI either. `.oxlintrc.json` line 81 lists `resources/` in `ignorePatterns`, which matches anywhere in a path, so `src/process/resources/` is excluded along with the bundled-assets directory the pattern was written for.

Proven with matched controls, identical file content in two locations:

```
src/process/utils/__kpprobe.ts       -> Found 0 warnings and 2 errors.  1 file
src/process/resources/__kpprobe.ts   -> Found 0 warnings and 0 errors.  0 files
```

That blind spot covers 19 tracked source files under `src/process/resources/`, including all six builtin MCP servers, plus 82 under `scripts/`. Filed separately; not fixed here because widening the lint surface will surface pre-existing findings across a hundred files and that does not belong in a bug fix.

`oxfmt --check` does inspect both files and reports correct format.

### DIFF
--- a/src/process/resources/builtinMcp/conciergeDiagServer.ts
+++ b/src/process/resources/builtinMcp/conciergeDiagServer.ts
@@ -875,23 +875,42 @@ export const createConciergeDiagServer = (deps: ConciergeDiagDeps = {}) => {
         lines: [],
       };
     }
-    let files: string[];
+    // #1038: readdirSync returns DIRECTORY order, not sorted and not by time, so
+    // taking the first MAX_LOG_FILES entries kept an arbitrary subset that in
+    // practice skewed oldest. A diagnostics bundle then reported stale lines
+    // under the heading "recent errors", and every triage that trusted it was
+    // reading the wrong data. Select by mtime, newest first, then walk the
+    // survivors oldest-first so the trailing MAX_LOG_LINES slice below really is
+    // the most recent output rather than whichever file happened to be last.
+    let files: Array<{ name: string; stat: fs.Stats }>;
     try {
       files = fs
         .readdirSync(logDir)
         .filter((f) => f.endsWith('.log') || f.endsWith('.txt'))
-        .slice(0, MAX_LOG_FILES);
+        .map((name) => {
+          try {
+            const stat = fs.statSync(path.join(logDir, name));
+            return stat.isFile() ? { name, stat } : null;
+          } catch {
+            // Vanished or turned unreadable between readdir and stat. A rotating
+            // log directory does this routinely, and one missing file must not
+            // cost the whole section.
+            return null;
+          }
+        })
+        .filter((entry): entry is { name: string; stat: fs.Stats } => entry !== null)
+        .sort((left, right) => right.stat.mtimeMs - left.stat.mtimeMs)
+        .slice(0, MAX_LOG_FILES)
+        .reverse();
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);
       return { available: false, source: `log dir read failed: ${scrubHome(message)}`, lines: [] };
     }
 
     const collected: string[] = [];
-    for (const file of files) {
+    for (const { name: file, stat } of files) {
       const full = path.join(logDir, file);
       try {
-        const stat = fs.statSync(full);
-        if (!stat.isFile()) continue;
         const start = Math.max(0, stat.size - MAX_LOG_TAIL_BYTES);
         const fd = fs.openSync(full, 'r');
         try {
--- a/tests/unit/process/resources/builtinMcp/conciergeDiagServer.test.ts
+++ b/tests/unit/process/resources/builtinMcp/conciergeDiagServer.test.ts
@@ -373,6 +373,68 @@ describe('createConciergeDiagServer — recentErrors (logs)', () => {
     expect(serialized).toContain('/Users/<user>');
     expect(serialized).toContain('C:\\\\Users\\\\<user>');
   });
+
+  // #1038: readdirSync returns directory order, so keeping the FIRST
+  // MAX_LOG_FILES entries dropped the newest logs and the section reported
+  // stale lines as "recent". Eight files, mtimes set explicitly, and the two
+  // newest are also last alphabetically AND last created, so any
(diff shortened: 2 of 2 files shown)


FerroxLabs:
## Independent adversarial audit — head 57f21ac19 — FIX-FIRST, now fixed

An independent auditor re-measured the shuffle claim and found a real gap. Both are recorded here because this fixture has now been wrong three times.

### My stated number was optimistic

The auditor materialised the real merge-base implementation and drove both it and the fixed one through a seeded shuffle, asserting order-distinctness explicitly and asserting the readdir mock was actually consulted:

    250 distinct orders    buggy selection 1/250    buggy ordering 0/250    fixed 250/250
    2000 distinct orders   buggy selection 2/2000   buggy ordering 0/2000   fixed 2000/2000

So buggy 0 of 250 holds for the PAIR, but the selection test alone survives about 0.1 percent of orders, consistent with the combinatorial chance of the six newest landing in the first six slots. The ordering test covers it at a true zero. Recording the correction rather than the flattering number.

### The real finding: the fixture could not pin WHICH stat key

utimesSync(file, when, when) set atime equal to mtime, and the write loop created files in ascending age order, so atimeMs and creation order both reproduced the mtime answer exactly. Measured consequence: switching the implementation to atimeMs left the file green at 44 passed.

That is not cosmetic. atime tracks READS, not writes. Keying selection off it means merely opening a stale log promotes it to newest, and the bundle then reports old lines under recent errors, which is #1038 restored silently. Precisely the class this fixture was rewritten twice to remove.

The auditor proved it rather than asserting it, building a decorrelated fixture that killed the mutant, and confirming the known positive that the real implementation passes that fixture unchanged.

### Fixed

atime now runs opposite to mtime, and files are created in reverse age order, in both the selection and ordering fixtures. Verified:

    real implementation           46 passed, 9 skipped
    mutant mtimeMs -> atimeMs     2 FAILED   (previously passed green)

birthtimeMs remains unprovable on macOS whatever the fixture does: APFS clamps birthtime DOWN to mtime when utimesSync backdates a file. That is noted inline in the test so the next reader does not chase it.

### Also verified by the auditor, no action needed

The downstream consumer is coherent: collectBugReport slices the last 8 lines, which combined with the oldest-first walk yields the 8 newest. Stale-stat widening degrades to nothing rather than garbage: a shrink race yields a buffer of NULs that the error filter collects zero lines from. symlinkSync is not a Windows blocker, with existing unguarded precedent that passes the shard. tsc exit 0 and oxfmt clean.

oxlint blind spot re-confirmed with a known positive: oxlint on the two files reports "Finished on 0 files", against 1 file for a src path. tsc and oxfmt are the only real gates for anything under resources.

FerroxLabs:

PR #1107, opened 2026-08-20
DO NOT MERGE: artifacts milestone, opened for CI verification

FerroxLabs:
## DO NOT MERGE — opened to get CI's opinion, nothing else

This is the artifacts milestone: durable task workspaces, a dated artifact series published by
atomic rename, a metadata-only ledger, the publisher wired into the real cron run path, and the
deliverable card with Open / Reveal / Save a copy / run history.

CI has never run on any branch of this work. That is the only reason this PR exists.

## What is verified

Full unit suite on macOS: 19,361 passed, 0 failed.

Live-verified in a real app for the first time (real engine v0.13.3, real provider, isolated
profile, nothing run against a real user profile):

- enabling a bundled routine allocates a Finder-visible workspace under Documents with an
  identity marker, and persists it on the job
- two runs publish a real dated series with a latest pointer and an alias copy
- the card names and launches the actual registered default application, and Reveal opens the
  file manager at the exact run directory
- the run history lists both runs and each is addressable

Eleven defects were found and fixed doing that, each one mutation-proven. Six of them were only
reachable by running the product: a test that was red on macOS and green on Linux, an action bar
whose third button was unclickable at the default panel width, the most-clicked file in the
layout having no actions at all, an app-name cache that could label a deleted file, and two
guards that could be deleted with the suite still green.

## What is NOT verified, and what to watch

- Windows has never executed a line of this. Three containment checks use the
  path.relative(...).startsWith("..") idiom, which is unsound on Windows in both directions
  (case-folding, 8.3 short names): artifactLedger.ts:184, artifactTarget.ts:80, taskRun.ts:310.
  That is the most likely place for this to fail.
- Linux shard likewise unrun in CI.

## Security surface

bridgeAllowlist.ts is touched: 25 insertions, 0 deletions, all of them new denials. The
artifacts and promotion namespaces are denied to remote callers, with a red-team test alongside.
Nothing is removed from the allowlist and no existing denial is relaxed.

## Not included

The Send to connector lane is deliberately out of this PR. It carries two open HIGH findings from
its adversarial review and is branched off an older tip.


FerroxLabs:
Superseded and shipped. Closing.

This branch was opened for CI verification only and was never meant to merge. The artifacts milestone it was verifying is now live in Wayland Desktop v0.12.4, published today: the artifact store, the artifact cards, the Artifacts entry in Cockpit, and the id-based host boundary all landed on main via #1119 and shipped in that release.

Confirmed by comparison rather than assumption: src/process/services/artifacts did not exist at v0.12.0 and carries 12 files on the shipped tree.

PR #1110, opened 2026-08-21
DO NOT MERGE: combined release train, opened for CI verification

FerroxLabs:
DO NOT MERGE. Opened to run CI on the combined release train.

Everything that has been built since v0.12.2 was staged, in one branch, so it can be verified together instead of three ways.

## What is in it

- The artifacts milestone (98 commits, previously PR #1107)
- Workspace folder access, including the security fixes an external audit raised
- The bundled engine bump to wayland-core v0.13.4
- TVControl 2.3.0 and the ten Smart Trader skills

## The one thing held back

The folder-grant card no longer offers its "remember this folder" option, and the folder list in Settings is not shown. Both are gated on a single constant, FOLDER_GRANT_REPLAY_AVAILABLE.

The reason is not caution. A remembered folder is replayed with grant_path, and that command is absent from the desktop contract schema shipped at v0.13.4, so validateOutboundCommand throws before a frame is written. Verified by execution with scope:once, always_prefix and tool_deny as controls in the same run, and filed upstream as FerroxLabs/wayland-core#314. The button's label says the folder is remembered until you remove it. Nothing re-applies it, so that sentence is false, and a false statement on a consent surface is worse than a missing button.

Everything else about the feature ships: the store, the host-side vetting, the transport denial, and the in-band grant. Mutation-proven both directions, so landing the Core fix is one boolean and the suite already covers it.

## What this train fixes that was not visible before

The folder-grant card could never have worked on any engine we had pinned. The always_path approval scope was absent from the pinned contract schema, so Desktop's own outbound validation threw before the frame reached Core. No amount of live testing would have found it, because the failure was on our side of the wire. The corpus re-import is what makes the card reach the engine at all.

The engine bump also turned out to be five coupled edits rather than three. The two that were missing are the contract corpus re-import and the publisher attestation policy entry, and without either one the app fails closed.

## Not yet done

- Live verification. Nothing here has been exercised against a real engine or a real TradingView.
- A second independent audit pass on the fixed state. The first one returned no-go, all five findings are closed, and none of the fixes have been re-reviewed.


FerroxLabs:
Superseded and shipped. Closing.

This was the combined release train, opened for CI verification and explicitly marked DO NOT MERGE. Everything it was carrying reached users in Wayland Desktop v0.12.4, published today, which also bundles Wayland Core v0.13.6.

The work landed on main through #1119 and the follow-ups that unblocked the release: #1120 pinned the producer bun so the protected observers can re-derive the WhatsApp bridge, #1122 cleared two new severe advisories by upgrading rather than waiving and stopped the CLI probes outliving the app, and #1125 fixed the Windows publisher gate and the updater exit budget.

PR #1120, opened 2026-08-24
Pin the build job's bun so the observers can re-derive the bridge

FerroxLabs:
## The blocker

v0.12.2 is built and sitting as a Draft because whatsapp-bridge reads
present-but-invalid on win32-x64 and win32-arm64, on BOTH protected observers.
Four legs, one cause. Publish is skipped, so no user can reach the release.

## Root cause, established by execution

The observers re-run bun install and require the shipped
resources/whatsapp-bridge to be inventory-identical to their own tree
(verifySourceMirror). The build job asked for bun-version: latest while both
observers pin 1.3.14.

From the run logs, not inference:

| date | producer bun | observer bun | win32-x64 leg |
| --- | --- | --- | --- |
| 2026-08-19 (producer 32200374609) | 1.3.14 | 1.3.14 | PASS |
| 2026-08-24 (producer 32751269643) | 1.4.0 | 1.3.14 | FAIL |

git diff over that range is empty for both src/process/channels/whatsapp-bridge
and scripts/verify-packaged-resources.js. Nothing in the repository moved. The
toolchain did, because latest is not a version.

## Why it is invisible until Windows

bun writes node_modules/.bin entries as SYMLINKS on macOS and Linux, and
sourceInventory records a symlink as link:name:target with no bytes. On Windows
there are no symlinks; bun writes real .exe launcher stubs, which it hashes. A
bun version skew therefore cannot fail a POSIX leg and always fails both
Windows legs.

Measured on windows-2022 against this exact lockfile, three installs:

- 1.3.14 vs 1.3.14: IDENTICAL, 9673 entries. bun is deterministic at a pin.
- 1.3.14 vs 1.4.0: 16 files differ, every one a .bin/*.exe stub, 15872 bytes
  against 8192. Zero files added, zero removed.

## The change

Pin the producer to the 1.3.14 that the observers, release-acceptance-trust-root
and the bundled-bun authority already name. Nothing is exempted and no check is
weakened: verifySourceMirror still demands a byte-identical inventory, and the
Windows legs have no reconciliation of any kind. The two sides simply agree on a
toolchain again, which is what an independent re-derivation requires.

The guard fails if the build job floats (latest, or a partial pin like '1.3'
that drifts one patch at a time) or drifts away from either observer. Bumping
bun stays possible; it has to happen on all three together.

Sabotage-verified: latest gives 3 red, '1.3' gives 3 red, an exact-but-mismatched
'1.4.0' gives 2 red and 1 green, and the restored file is byte-identical to its
backup.

## Verification

- Full unit suite 20,260 passed / 0 failed / 164 skipped
- tsc --noEmit clean, oxfmt clean on the touched file

## Follow-up worth knowing

The observers are the only check that inspects the INSTALLED tree; the build
only checks out/. Previously shipped Windows builds may carry the same
non-reproducible bundle without anyone noticing.

### DIFF
--- a/.github/workflows/_build-reusable.yml
+++ b/.github/workflows/_build-reusable.yml
@@ -232,7 +232,29 @@ jobs:
       - name: Setup bun
         uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          # PINNED, and it must stay pinned. The tree that SHIPS as
+          # resources/whatsapp-bridge is built by prepareWhatsAppBridgeResources in
+          # scripts/build-with-builder.js: it deletes the bridge's node_modules and
+          # reinstalls it clean with `bun install --frozen-lockfile --os <platform>
+          # --cpu <arch>`, invoking whatever `bun` this step put on PATH. The
+          # protected observers then re-derive that tree with their OWN `bun install`
+          # and require a byte-identical inventory (verifySourceMirror). That only
+          # holds while both sides run the SAME bun.
+          #
+          # On Windows bun writes node_modules/.bin/* as real .exe launcher stubs,
+          # which sourceInventory hashes; on macOS and Linux they are symlinks,
+          # recorded as `link:<name>:<target>` with no bytes. So a bun version skew
+          # is invisible on POSIX and fails EVERY Windows leg.
+          #
+          # Measured on windows-2022, same tree, same lockfile: 1.3.14 twice is
+          # byte-identical; 1.3.14 vs 1.4.0 differs in exactly 16 files, all of them
+          # .bin/*.exe stubs (15872 -> 8192 bytes), nothing added or removed.
+          #
+          # `latest` is what broke v0.12.2: it resolved to 1.3.14 on 2026-08-19 and
+          # all six legs passed, then to 1.4.0 on 2026-08-24 and both Windows legs
+          # failed on an unchanged tree. The observers pin '1.3.14'; so does
+          # release-acceptance-trust-root.yml and the bundled-bun authority. Match them.
+          bun-version: '1.3.14'
 
       - name: Setup Python
         uses: actions/setup-python@v5
--- a/tests/unit/scripts/producerObserverBunPin.test.ts
+++ b/tests/unit/scripts/producerObserverBunPin.test.ts
@@ -1,0 +1,82 @@
+/**
+ * @license
+ * Copyright 2026 Ferrox Labs
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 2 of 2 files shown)


FerroxLabs:
Independent adversarial audit (Codex 5.6 Sol, read-only, non-author): **GO**.

It was asked to refute the fix and defaulted to "this will not work". It could
not, and it confirmed the two things I could not prove from the x64 box alone:

- **arm64 is covered.** Producer and observers pair windows-11-arm with
  win32-arm64 and windows-2022 with win32-x64, so the pin lands on both. The
  2026-08-19 all-six pass at 1.3.14 is stronger evidence than my x64-only
  reproduction.
- **No credible downgrade break.** package.json declares bun >=1.3.0 <2.0.0 with
  packageManager bun@1.3.11, the bundled runtime is already 1.3.14, and this
  build previously passed on 1.3.14.

It found one real defect, now fixed. My comment claimed electron-builder ships
the root postinstall tree. It does not: prepareWhatsAppBridgeResources deletes
the bridge's node_modules and reinstalls it clean with
bun install --frozen-lockfile --os PLATFORM --cpu ARCH, using whatever bun this
step put on PATH. The pin still lands exactly where it matters, but the comment
was wrong and would have misled the next reader. Corrected in the workflow
comment, the test docblock and the commit message.

Worth recording, not fixed here: the producer installs the bridge with explicit
--os/--cpu while the observers use a plain nested install. On a native runner
those resolve the same, which is why 2026-08-19 passed, but the two sides
deriving the same tree by different commands is a latent fragility. Hardening it
means changing the observers, which lives on the protected branch.

FerroxLabs:
Closed the last unproven link, on the real windows-2022 box.

The audit called the producer/observer command asymmetry theoretical residual
nondeterminism: the producer installs the bridge with
bun install --frozen-lockfile --os win32 --cpu x64 (prepareWhatsAppBridgeResources)
while the observers use a plain nested bun install (postinstall.js). Theoretical
is not proven, so I measured it.

Same box, same tree, same lockfile, both at the pinned bun 1.3.14:

    P (--frozen-lockfile --os win32 --cpu x64)  vs  O (plain install)
    => IDENTICAL, 9673/9673 entries

So the differing commands contribute nothing on a native runner. Combined with
the earlier three-way run, the chain is now fully measured rather than inferred:

- 1.3.14 vs 1.3.14, same command:      IDENTICAL (9673)
- 1.3.14 vs 1.3.14, DIFFERENT commands: IDENTICAL (9673)
- 1.3.14 vs 1.4.0:                      16 files differ, all .bin/*.exe stubs

The bun version was the only variable, which is exactly what this pin removes.

PR #1122, opened 2026-08-25
Clear the new severe advisories, and stop the CLI probes outliving the app

FerroxLabs:
Two blockers that stopped v0.12.3, both found by sweeping ahead rather than by
waiting for CI to hit them.

## 1. The dependency-security gate would have refused the release

`bun audit` picked up two new high advisories that are not on the accepted list:

    fast-uri  GHSA-4c8g-83qw-93j6  host confusion via failed IDN canonicalization
    tar       GHSA-r292-9mhp-454m  uncontrolled recursion, stack-overflow DoS

Run against the real audit output, the gate rejected with
`M8I_SEVERE_DEPENDENCY_FINDINGS:2`.

Both packages are already pinned through overrides/resolutions and both have a
patched release inside the same major line, so this is a version bump rather than
a waiver. That also mattered for a second reason: the gate runs the TRUST
branch's copy of the acceptance list, so waiving would have required a push to
release-trust-v1, while upgrading needs no protected-branch change at all.

The bump clears more than it had to. fast-uri ^3.1.6 also retires the two
host-confusion advisories already on the list, and tar ^7.5.22 retires the
accepted `Decompression/parse DoS` CRITICAL plus a second high. The gate now
reports critical 0 / high 0 outstanding and the accepted list drops from 28
entries to 23, with no criticals left in it.

## 2. The CLI probes outlived the app

`killAllAgentChildren` is the final before-quit step and can only kill what
`trackAgentChild` registered. wcore and AcpConnection register theirs;
safeExec and safeExecFile did not, and they are what AcpDetector uses - a
`where` then a `powershell -Command Get-Command` per agent, eighteen agents
deep, plus a WSL probe for the rest. On quit those probes were invisible to the
reaper.

This is what actually failed v0.12.3, twice:

    [platform-package-smoke] packaged app left descendant processes alive: ...

once on the windows-x64 build and once on the win32-x64 updater observer, both
while the app had already logged a clean `quit (exitCode=0)`. The smoke check
walks the real process tree from the app's root pid, so those were genuine
orphans, not an over-strict check.

Proven red first: before the change the reaper saw zero of them and the new test
failed on `expected +0 to be 1`.

## Scope, stated plainly

This closes a proven hole, not provably every orphan. It reaps everything
spawned through safeExec, which covers all the detector probes in the failing
logs, and the 4-PID updater failure matches it exactly. An earlier 110-PID
failure may have included processes from other spawners.

Not fixed here, on purpose: safeExec's local killChild fires taskkill async and
unawaited, so its timeout path can return before the kill lands, and each
taskkill is itself a short-lived child. An awaited killChild already exists in
@process/agent/acp/utils and the duplicate should fold into it - a refactor, not
this fix, and the reaper now catches the consequences either way.

## Verification

- Full unit suite 20,261 passed / 0 failed / 166 skipped
- tsc --noEmit clean, oxfmt clean on the touched files
- Dependency gate re-run against a fresh `bun audit --json`: critical 0, high 0


FerroxLabs:
The first Windows shard failed on my own test, not on the fix. Corrected.

safeExec runs the command through cmd.exe /c on Windows and sh -c on POSIX, so
the command has to be shell-native rather than an argv the shell re-parses. I had
used a node one-liner, and setTimeout(() => {}, 60000) is all cmd metacharacters:
parens group, the comma delimits. cmd mangled it, the spawn died immediately, the
error handler dropped the child, and the registry correctly read 0.

Reproduced on a real Windows machine with the exact spawn safeExec performs:

    OLD (node one-liner): alive_after_250ms=false  closedCode=1
    NEW (ping)          : alive_after_250ms=true   closedCode=null

So the assertion was right and the fixture was wrong. The test now uses
`ping -n 61 127.0.0.1 >nul` on Windows and `sleep 60` on POSIX, and the reason is
recorded in the test so nobody reaches for a node one-liner here again.

Worth stating plainly: this makes the Windows leg of the test meaningful for the
first time. It had never actually exercised the safeExec path on Windows - it
failed there on its first run.

Issue #1189, opened 2026-08-29
[Desktop] call_announced is dropped on the unknown-event arm, so auto-approved tool calls store no command text


FerroxLabs:
Found while decomposing #1088's blocked criteria. Filing separately because it is its own defect with
its own user-visible symptom, and because it explains a measurement someone already reported.

## The defect

`call_announced` is the frame Core emits for **every auto-approved tool call** — force mode,
allow-listed, command-scoped grant, or a tool the user just granted Always. It is also the **only**
frame that carries the tool `args` for those calls.

It is **absent from Desktop's `WCoreEvent` union and from the dispatch switch**, so it falls through
to the unknown-event drop arm at `wcore/index.ts:1819-1824`.

## Why this is the interesting part

Core's design intends that drop arm as a **safe forward-compatibility fallback for OLDER hosts** —
an unknown frame should be ignorable. That is correct as a design. What nobody noticed is that the
host dropping the frame is not an old host. **It is the current one.**

So a mechanism built to protect against version skew is silently discarding a frame the shipped
Desktop was supposed to consume, and it does it without a warning, because dropping unknown frames
is exactly what it is meant to do.

## The symptom it explains

The reporter who measured *"21 of the 22 tool calls stored no command text"* was measuring this. The
one call that DID store command text was presumably not auto-approved, so its args arrived on a
different frame.

## Acceptance criteria

- [ ] `call_announced` is in the `WCoreEvent` union and handled in the dispatch switch.
- [ ] A tool call that is auto-approved (each of: force mode, allow-listed, command-scoped grant,
      just-granted Always) stores its command text and renders a tool card.
- [ ] The `/rewind` view shows files touched by auto-approved calls (this is the second symptom the
      same drop causes).
- [ ] **A test that a frame Core emits and Desktop claims to support cannot silently reach the
      unknown-event drop arm.** The drop arm must stay for genuinely unknown frames, but a frame in
      the current contract reaching it is a defect, not forward compatibility. This is the criterion
      that stops the class, not just the instance.
- [ ] Red arm quoted verbatim: show the frame landing on the drop arm before the fix.

## Note on scope

Core already emits this frame correctly — that work shipped in v0.13.10 (`call_announced` was added
precisely so auto-approved calls stop dispatching with nothing on the wire). **No Core change is
required.** This is entirely the consuming side.

PR #1325, opened 2026-09-05
fix(wcore): retain auto-approved tool details

FerroxLabs:
## Description

Auto-approved Core tool calls announce their inputs with `call_announced`, which Desktop previously dropped after contract validation. Handle the announcement as an executing tool card, retain its command/file details through completion, and keep results visible alongside those details. No additional approval is requested. The dispatcher now has a compile-time exhaustiveness check for its declared event union.

## Related Issues

- Refs #1189. Issue closure remains a release action.

## Type of Change

- [x] Bug fix

## Testing

- [x] macOS: focused adapter, message-merge and renderer tests; typecheck, lint and local `prek` checks.
- [x] Linux: complete `bun run test` on an isolated Ubuntu 24.04 VM as an unprivileged user: 21,466 Vitest tests passed, 53 skipped; 285 Bun-native tests passed. Typecheck passed.
- [x] Red controls recorded before the fix for missing announcement details and hidden renderer results.
- [x] Tested Git tree `d9f2dc7c3dc94cf162fdca5262b84428e3a6178b` matches the committed tree. Source remained unchanged during the Linux run.
- [x] i18n validation passed with existing warnings; full lint remained at the existing 3,017-warning baseline with zero errors.
- [ ] Windows CI and packaged real-engine verification pending.

## Additional Context

The renderer adjustment prevents retained input details from hiding tool output, including completed edit diffs. The existing path-boundary approval guard remains covered. This draft does not claim native installer, every approval-mode journey, or `/rewind` acceptance.

### DIFF
--- a/src/process/agent/wcore/index.ts
+++ b/src/process/agent/wcore/index.ts
@@ -1614,8 +1614,10 @@ export class WCoreAgent {
         break;
 
       case 'tool_request':
+      case 'call_announced':
         // #520: remember the request-time command so the later running/result
-        // frames (which the wire sends without it) can re-surface it.
+        // frames (which the wire sends without it) can re-surface it. #1189:
+        // auto-approved calls announce the same ToolInfo without a request.
         this.toolDescriptionByCallId.set(event.call_id, event.tool.description);
         // #746: the agent is now waiting — first on the human (this frame renders an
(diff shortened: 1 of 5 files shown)


FerroxLabs:

Issue #1328, opened 2026-09-05
Core contract: expose effective tool approvals and preserve per-tool default inheritance (#1188)


FerroxLabs:
Desktop #1188 needs producer-owned tool approval and registration truth to remove its remaining mode/registry mirrors without inventing defaults.

At the audited Desktop baseline 8354ea7e4 (Core v0.13.12, Desktop contract 1.23), `workspace_policy` contains trust, backend, readable/writable roots and executable capabilities. It does not contain the registered tool inventory or effective per-tool approval decisions. `wcoreConfig.getSection` reads the raw TOML and preserves absent fields. The exposed CLI has init/config-path actions, while effective TOML rendering is currently used internally by the TUI.

Source correction to #1188: Core v0.13.12 `crates/wcore-config/src/config.rs:1388` currently has the same 11-tool default list as Desktop. Omission of Write/Edit/Bash means they still require approval, not that they are unregistered or disabled. The confirmed defect is duplicated authority and accidental materialization of inherited configuration.

Desktop work in progress forwards and retains the workspace receipt, shows live session filesystem access, and removes the default-list seed: absent remains inherited and per-tool status is unknown. Explicit custom lists can still be edited. This prevents silent default replacement, but cannot honestly expose Core's effective defaults or dynamic registration from the current receipt.

Requested producer contract (Core-owned implementation):

- Publish a typed, session-scoped effective tool-policy/inventory snapshot with canonical tool IDs, registration state, effective approval disposition, and source/provenance. Re-emit or revise it after mode/config/registry changes.
- Include engine-derived mode effects so Desktop can remove its Write/Edit AutoEdit mirror and stop treating a curated static catalog as runtime availability.
- Provide an inheritance-preserving per-tool override mechanism, or an equivalent bounded mutation contract: changing one tool while the base allow-list is absent must preserve the effective decisions for untouched tools and preserve future inherited defaults. A whole-list replacement from a Desktop seed is insufficient.
- Pin producer fixtures and negative tests for absent versus explicit-empty lists, default/AutoEdit/Force modes, unregistered tools, and stale session revisions. Older producers must remain explicitly unknown in Desktop, never fall back to a hardcoded guessed list.

Acceptance: replaying the snapshot lets Desktop render effective decisions without a default or mode-specific tool-name literal; toggling a single tool from inherited configuration leaves every untouched tool's effective disposition and inheritance intact. Keep filesystem access separate from tool approval and registration.

[...]

Issue #1188, opened 2026-08-29
[Desktop] workspace_policy receipt is dropped, and ToolsPane ships a stale copy of core's default allow-list (#1088)


FerroxLabs:
**Handoff from the core lane.** Core's half of #1088 is complete; this is the Desktop half, split out so #1088 stops reading as a partial. Investigating it turned up a **live second defect** that is worse than the report, so please read past the first section.

## Contract

Desktop must consume the `workspace_policy` receipt core already emits, and must stop shipping a hardcoded copy of core's default tool allow-list. Between them these are why a user is told Read / Glob / Write / Edit are restricted when they are not — and, in the second case, why they can be *made* restricted for real by a single click.

## Why core cannot do this

Core emits the typed event. Desktop decodes it and then throws it away before it reaches any UI.

**Core's side is present and typed.** `WCoreWorkspacePolicy` (`readable_roots`, `writable_roots`, `capabilities`, `backend`) is already modelled in Desktop at `src/process/agent/wcore/protocol.ts:232-256`, with a union arm at `:269`. The main process decodes it and caches it: `src/process/agent/wcore/index.ts:1457-1461` — `this.lastWorkspacePolicy = event.policy; … onStreamEvent({ type: 'workspace_policy', data: event.policy, msg_id: '' })`.

**Then it is dropped twice.**

1. `src/process/task/WCoreManager.ts:1911-1923` holds an explicit allowlist of empty-`msg_id` session events that are forwarded to the renderer: `execution_policy`, `workflow_started`, `workflow_node_event`, `workflow_finished`, `anvil_receipt`, `anvil_receipt_invalidated`, `anvil_trust_changed`, `mcp_ready`, `mcp_failed`. **`workspace_policy` is not in it.** Falling past that, it hits the guard at `:2111` — `if (!data.msg_id) return;` — and is discarded.
2. Defence in depth against ever fixing it by accident: even if it reached the renderer, `transformMessage` has no `workspace_policy` case and it is not in the inert skip-list, so it lands on the `default:` warn arm at `src/common/chat/chatLib.ts:1220-1226`.

Corroboration: `rg 'workspace_policy|readable_roots|writable_roots'` over `src/renderer/` and `src/common/` returns **zero** hits.

[...]

PR #1329, opened 2026-09-05
fix(wcore): preserve workspace receipts and inherited approvals

FerroxLabs:
Core workspace-policy receipts currently disappear at the manager’s empty-message guard, leaving the chat without filesystem-access evidence. This change preserves accepted receipts and exposes live readable/writable roots and executable capabilities beside the permission selector; dead or disposing sessions report unknown access. Raw stream payloads only trigger a fresh main-process read.

The Tools pane now preserves an absent allow-list as inherited instead of writing a hardcoded default list when a setting changes. Inherited per-tool states remain unknown, and independent registration gates remain editable. Existing explicit custom lists still round-trip.

Related to #1188. This is an intermediate repair: producer-backed tool inventory, mode effects and inheritance-preserving per-tool overrides remain tracked in #1328. The workspace receipt describes filesystem access and does not prove a tool’s registration or approval state. No issue closure is requested.

Validation: full Linux `bun run test` passed 21,470 Vitest tests (53 skipped) and 285 Bun-native tests; typecheck passed against the exact committed tree. Focused macOS receipt/settings/access tests passed 142 cases, plus all 24 voice integration cases. Both receipt-drop and default-materialization regression tests failed before the fixes. Repository lint has the existing 3,017 warnings and zero errors; i18n and the complete local prek checks passed. The initial full run exposed a missing mock for the new access control in the voice harness; its evidence was retained and the corrected full run passed with no source drift. Native packaged user-flow acceptance remains pending.


FerroxLabs:

Issue #1223, opened 2026-08-29
Desktop: chat interface still shows Read/Glob/Write/Edit as restricted after a refused set_mode (wayland#1088 c2)


FerroxLabs:
Split out of FerroxLabs/wayland#1088 so its Desktop remainder is a ticket rather than a sentence in a ledger. Core's half of #1088 shipped; this is the half that is graded on the Desktop surface.

## Originating criterion, verbatim

> **c2** — "The user-visible half — the chat interface no longer reports Read/Glob/Write/Edit as restricted"
>
> `state: blocked`, `owner: desktop`
> note: "the surface that renders the restriction is Desktop's; core emits the typed event it needs. Ticket carries needs:desktop"

Parent: FerroxLabs/wayland#1088. Ledger entry: `.planning/ledger/wayland-1088.md` in wayland-core.

## What core already shipped, so the input exists

- `set_mode_refused` is a typed event on the wire — `crates/wcore-protocol/src/events.rs:1319`. It is the typed nack for a wire `set_mode` that asked for an auto-approving mode the local launcher did not opt into.
- The published desktop contract went 67 -> 68 events and contract version 21 -> 22 to carry it (`crates/wcore-protocol/tests/desktop_contract_corpus.rs:368`, `crates/wcore-protocol/src/contract/generate.rs:116`).
- The corpus guard was moved INSIDE `generated_artifacts()` so a corpus carrying the wrong row cannot be emitted at all, rather than being caught afterwards.

So the refusal is no longer prose Desktop has to guess at; it is a typed frame with a schema branch and a fixture.

## Acceptance criteria

- **c1** — Against a wayland-core built from contract 1.22, a session in which a `set_mode` request is refused renders that refusal from the `set_mode_refused` frame, and Read, Glob, Write and Edit are NOT presented to the user as restricted.
- **c2** — A Desktop-side test drives a `set_mode_refused` frame and asserts the four tool names are shown as available. If Desktop has no harness that can inject a frame, a manual run is acceptable ONLY with the raw JSON-stream frame log and the rendered screenshot both attached to this issue.
- **c3** — The negative arm is graded too: a session where `set_mode` is ACCEPTED also shows the four tools available. Without this, "always show available" passes c1 and c2 while fixing nothing.
[...]

PR #1330, opened 2026-09-05
fix(wcore): wait for confirmed approval mode changes

FerroxLabs:
Changing Core from Default to an automatic approval mode currently updates Desktop’s local approval authority and reports success before Core accepts the command. A later typed refusal is dropped, leaving the permission selector and host-side approvals out of sync with Core.

This change keeps the previous mode until an accepted policy/config receipt arrives, consumes `set_mode_refused`, and returns the effective mode on rejection. Overlapping requests, disconnected transports and unconfirmed requests fail without widening approval authority. Temporary scheduled-run mode changes remain unpersisted. The renderer explains the refusal, refreshes main-process truth, ignores delayed results from another conversation and prevents a capability-label refresh from reinstating an old stored preference.

Related to #1223. A mode refusal does not itself disable tools; the UI says that explicitly rather than inventing a tool inventory. Producer-backed per-tool availability remains the separate #1328 contract dependency. No issue closure is requested.

Validation: exact committed tree `5e3cdf68c18faa88215a9bba3002e0ad2723f3bf` passed full Hetzner verification: 21,472 Vitest tests (53 skipped), 285 Bun tests and typecheck, with no source drift. Local focused manager/decoder/renderer suites pass 76 tests. Regression tests reproduce premature local elevation and stale stored-mode restoration before their fixes; accepted and refused paths are both covered. Local typecheck, i18n validation, lint (existing 3,017 warnings, zero errors) and full prek checks pass. Remote setup errors and the initial missing nested dependency test failure were preserved before the corrected full run. Native packaged user-flow acceptance remains pending.


FerroxLabs:

Issue #1249, opened 2026-08-30
[Desktop] Workflow and cron sessions must select an approval mode instead of inheriting yoloMode (wayland#1195 c7)


FerroxLabs:
Carrier for the **Desktop half** of FerroxLabs/wayland#1195, `c7` in `.planning/ledger/wayland-1195.md` (wayland-core).

Core's half is done: the engine's gating behaviour under each of the three session modes has been **measured on the real `--json-stream` wire** (not inferred from a predicate unit test) and pinned in `crates/wcore-cli/tests/permission_mode_matrix.rs`. This issue carries the product fix, which is Desktop's.

## What core measured

One real `wayland-core --json-stream` process per cell, mode applied exactly as Desktop applies it (`WAYLAND_ALLOW_WIRE_FORCE=1` + a wire `set_mode`), verdict read from real stdout — `approval_required` = gated, `call_announced` = the engine ran it without asking.

| tool (class) | `default` | `auto_edit` | `force` (= `yolo`) |
|---|---|---|---|
| `Read` (info, on the shipped allow_list) | auto | auto | auto |
| `todo` (info, **not** allow-listed) | GATE `info` | GATE `info` | auto |
| `Write` (edit) | GATE `edit` | **auto** | auto |
| `Bash` (exec) | GATE `exec` | GATE `exec` | auto |
| `WebFetch` (mcp, allow-listed) | auto | auto | auto |
| `Read` outside the workspace | GATE + `escalation.kind = path_boundary` | GATE + `path_boundary` | auto, then refused by the workspace policy |
| `AskUserQuestion` | GATE `info` | GATE `info` | GATE `info` |

So `auto_edit` does what its name says: **it moves exactly the built-in `Write`/`Edit` pair from gated to auto and widens nothing else.**

### 1. Workflow and cron sessions must SELECT a mode, not inherit `yoloMode: true`

- `src/process/services/workflow/WorkflowSessionService.ts` — `effectiveSessionMode` currently falls back to `getFullAutoMode(backend)`, which is `'yolo'` for `wcore`.
- `src/process/services/cron/WorkerTaskManagerJobExecutor.ts` — every scheduled run is built with `{ yoloMode: true }`.

Recommended: **`auto_edit` for both**, declared per workflow / per task rather than inherited, with `yolo` available as an explicit opt-in for a task whose author accepts it.

### 2. The unattended posture must drop the `'info'` arm of `tryAutoApprove`

[...]

PR #1331, opened 2026-09-05
fix(automation): declare modes and bound unattended approvals

FerroxLabs:
Scheduled tasks currently enable blanket auto-approval regardless of their stored mode, and workflow launches inherit global full-auto preferences. Core also treats broad Info/Edit categories as safe during automation, while unanswered Core prompts have no bounded hold.

This change declares guarded per-run defaults, exposes approval choices in task/workflow launch UI, preserves supported saved task choices, and enables blanket approval only for a supported explicit bypass mode. New Core routines use Auto-Edit; unsupported Core mode values use ask-first. Scheduled runs recreate idle runtimes to clear sticky flags and abort safely if settings still fail after retry, while preserving the user's conversation settings and protecting successor tasks.

Automated Core runs no longer auto-approve broad Info or non-file Edit categories. Scheduled confirmations and token-based approvals expire by denial, including filesystem-boundary cards. Duplicate prompt updates cannot renew the hold; a foreign boundary reply cannot disarm it; transport death cannot turn expiry into an uncaught host exception. The deadline is the scheduler's existing duration, not a Unix timestamp. README documents the posture.

Related to #1249. Integrated native acceptance also depends on the confirmed-mode repair in #1330. No issue closure is requested.

Validation: full Linux `bun run test` and typecheck pass on exact tree `a49aba478fb265be9f2961f13537556f09db551d`: 21,479 Vitest tests (53 skipped), plus 285 Bun tests, with no source drift. Targeted tests cover defaults, explicit modes, legacy unsupported values, task/workflow UI, failed retries, borrowed-session restoration, successor protection, duration units, duplicate prompts, boundary denial and transport races. The initial full run's eight older cron-fixture failures were preserved before the updated lifecycle checks and successful rerun. Lint adds no diagnostics over the existing baseline; i18n validation passes. Packaged user-flow and integrated release acceptance remain pending.


FerroxLabs:

PR #1401, opened 2026-09-19
fix(npm): ship the managed OfficeCLI shims in the headless payload

### LINKED ISSUE #1316: [Bug]: Headless npm install (getwayland): every chat fails — "Wayland managed OfficeCLI fallback guard is unavailable" — payload ships without resources/managed-cli-shims/officecli
## Platform

Linux, any (verified on Debian 13 headless, KVM VPS, getwayland installed via npm).

## Wayland Version

getwayland 0.12.13 (npm). Server payload `payload/dist-server/server.mjs`.

## Bug Description

On a fresh headless npm install, **every WebUI chat fails** with an error, and the
journal shows the same throw at boot:

```
error: Wayland managed OfficeCLI fallback guard is unavailable
41675 |   if (!shimDir) throw new Error("Wayland managed OfficeCLI fallback guard is unavailable");
```

Provider setup is fine (env key import logs `Imported provider key from
environment: openrouter`), the WebUI works, login works — only chats die.

## Root cause

`getBundledNpmBinDirs()` unconditionally calls `getManagedOfficeCliShimDir()`,
which throws when `resolveManagedOfficeCliShimDir()` doesn't find
`<resources>/managed-cli-shims/officecli` (a sha256-pinned 139-byte stub, exec
bit required, symlink forbidden). That file ships in the desktop build
(repo path `resources/managed-cli-shims/officecli`, commit b3cd0511) but is
**absent from the npm package payload** — `files` includes `payload/`, whose
contents are only `dist-server/`, `out/`, `src/` (no `resources/` beyond what
postinstall creates for wayland-core). The throw escapes during agent/session
spawn, so every chat attempt errors out.

## Steps to Reproduce

1. Fresh Debian 13 box: `npm i -g getwayland && wayland setup` (systemd service: y).
2. Add any provider key via env (e.g. `OPENROUTER_API_KEY` in `wayland.env`), restart.
3. Open the WebUI, log in, start any chat → error.
4. `journalctl -u wayland` shows the throw at boot and at each chat attempt.

## Expected Behavior

Either the npm payload includes the pinned shim (or postinstall fetches it),
or the office toolchain degrades gracefully when absent — office support is
optional on a headless server; its absence must not break chat.

## Workaround (verified working)

Install the repo's pinned shim into the payload and restart:

```bash
curl -s https://raw.githubusercontent.com/FerroxLabs/wayland/main/resources/managed-cli-shims/officecli \
  -o /tmp/officecli-shim
sudo install -m 755 /tmp/officecli-shim \
  /usr/local/lib/node_modules/getwayland/payload/resources/managed-cli-shims/officecli
sudo systemctl restart wayland
```

(Content sha256 matches the code pin `b77af087b9ce061bd26957e387b6dc56491ef894409e52317d11692a63ac327f`.)
Note: `npm i -g getwayland` wipes the payload — the shim must be reinstalled
after any package update/reinstall, until the packaging includes it.

## Related context

Two other headless defects reported from the same deployment: silent
non-start on masked-CPU VPSes (bun 1.4.0 init spin), and the WebUI
connect-provider CSRF failure (single-outstanding token rotated per HTTP
response). Headless smoke coverage seems worth strengthening across the board.

FerroxLabs:
Fixes #1316.

Every chat fails on a fresh headless npm install, and has since 0.12.13: the server throws "Wayland managed OfficeCLI fallback guard is unavailable" at boot.

## Cause

getManagedOfficeCliShimDir (src/process/utils/shellEnv.ts:180-188) is fail-closed, and the npm launcher runs the server with cwd = payload (installer/bin/wayland.mjs:302), so the guard resolves payload/resources/managed-cli-shims.

installer/scripts/build-payload.mjs copies exactly four resource trees (skills, assistant, skills-library, bundled-workflows). The shims are not one of them: the desktop receives them through electron-builder extraResources (electron-builder.yml:154), which the payload build never runs. The directory has therefore never existed in any published tarball.

Confirmed against what we published yesterday:

    npm pack getwayland@0.13.1 && tar -tzf getwayland-0.13.1.tgz | grep -c managed-cli-shims/
    0

## Change

Stage the shims into the payload, and fail the build rather than emit a payload that boots into the guard error. The executable bit is checked too, because the guard rejects a non-executable POSIX shim, and a checkout that lost the bit would otherwise copy cleanly and fail at runtime.

Staging is its own module (installer/scripts/stageManagedCliShims.mjs) so it can be tested without running a full renderer and server build.

## Evidence

- 5 new tests in tests/unit/scripts/stageManagedCliShims.test.ts: both shims present, POSIX shim stays executable, missing source refuses the build, partial source names the offender, non-executable source refuses the build
- mutation-checked twice. Replacing the copy with an empty directory turns 2 tests red. Removing the executable-bit guard initially turned NOTHING red, because the copy preserves modes anyway - so the fifth test was added to cover the guard itself, and it now goes red under that mutation. Worth stating plainly: the first version of this test file would have passed over a broken guard.
- full suite: 1762 passed, 25 skipped, 1 unrelated failure in tests/integration/i18n-performance.test.ts (a wall-clock budget test that went red while other work ran on the same machine; passes 8/8 in isolation)
- typecheck clean, oxfmt and oxlint scoped to the touched files

🤖 Generated with [Claude Code](https://claude.com/claude-code)


FerroxLabs:
Pushed a follow-up: the executable-bit guard I added failed every Windows unit shard, because NTFS has no POSIX mode bits - writeFileSync with mode 0o755 does not create one, so the guard threw on a host that cannot express the condition. The guard and the two mode-sensitive assertions are now gated on a POSIX host, which is where the payload that ships to POSIX users is built. Local suites could not have caught this; CI did.

